PR #17, opened 2019-01-22
v1.2.2 - Fix latest datapoint format

### LINKED ISSUE #16: `Datapoints.retrieveLatest` returns undefined
It seems from the test that expected response is
```
        data: {
          items: [
            {
              name: 'Cognite',
              datapoints,
            },
          ],
        },
```
but the real one from `https://api.cognitedata.com/api/0.5/projects/akerbp/timeseries/latest/IA_21PT1019.PV` is
```
"data": {
    "items": [
      {
        "timestamp": 1548140884056,
        "value": 0.054688
      }
    ]
  }
```

ghost:
Fixes #16

### DIFF
--- a/src/Datapoints.ts
+++ b/src/Datapoints.ts
@@ -8,6 +8,12 @@ export interface Datapoint {
   value: number | string;
 }
 
+interface DataPointResponse {
+  data: {
+    items: Datapoint[];
+  };
+}
+
 export interface DataDatapoints {
   name: string;
   datapoints: Datapoint[];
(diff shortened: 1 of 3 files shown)


src/Datapoints.ts
```diff
@@ -154,7 +154,7 @@ export class Datapoints {
     )}`;
     const params = { before };
     const response = (await rawGet(url, { params })) as AxiosResponse<any>;
-    return response.data.data.items[0].datapoints;
+    return response.data.data.items[0];
```
valgussev, line 165: What if we have empty array in response?

ghost: ~~Good catch 🚀 . Will push a fix~~

ghost: No problem. In JS `[][0]` will return `undefined` and expected behaviour.
I have committed a new test for that.

valgussev (approved):

PR #23, opened 2019-02-04
Store X-Request-Id and use exponential backoff retry

ghost:
- If a request fails (1xx, 4xx, 5xx) and the `X-Request-Id` is present in the headers then keep it when generating the `Error` class instance and print it out by default on unhandled exceptions
- Retry requests if the http method is `GET`, `DELETE`, `PUT`, `HEAD`, `OPTIONS` and response status code is 1xx, 5xx or 429. The retry policy is exponential backoff.
- Added option to specify request timeout in the `configure` method

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
   "dependencies": {
     "axios": "^0.18.0",
     "jwt-decode": "^2.2.0",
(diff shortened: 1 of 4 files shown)


ghost (approved):
LGTM

PR #24, opened 2019-02-04
Added capabilities to Group

ghost:


ngerstle-cognite (commented):
On a quick skim it looks reasonable to me, but I'd much rather have reuse the source proto files, or this will get out of data/be incomplete.
I did note that the generic.* capability is missing, which is presently used for the resource types without their own acls (ML stuff, I think? maybe extraction resources?) and the pipeline acls.
I do wonder if we should expose all the projects actions, as create/list aren't/shouldn't be exposed to any external users at this point in time, but I guess this sdk will be used for internal code as well.

ghost:
@ngerstle-cognite I couldn't see `generic.*` capability in the API reference. Could you point me to the right place? I couldn't see ML or pipeline acls in the docs.

Yes, it will also be used internally.

I don't see any good solutions with using the proto-files. I need to deploy a new version for every change anyway, and don't know how to convert them to typescript.

ngerstle-cognite:
They're explicitly hidden from the api docs, as the the endpoints requiring them aren't public- hence why I ask if they should be present or not.
Eg, the `getPipelines` endpoint requires the `Pipeline.Read` capability-> the `getPipelines` operation is hidden from the generated docs, so the capabilities will as well, except via the `createGroup` spec.

ngerstle-cognite:
Actually, that's on me. 
I retract my earlier statements- feel free to release (though I should redact the projects.{create,list} capabilities), and be aware that not all capabilities respected by CDP are covered.

ghost:
@ngerstle-cognite Ok, thanks. ✅ ?

src/Groups.ts
```diff
@@ -10,12 +10,160 @@ export interface Permissions {
...
+    };
+  };
+}
+
+export interface UsersAcl {
+  usersAcl: {
+    actions?: (LIST | CREATE | DELETE)[];
+    scope: {
+      all?: {};
+      currentuserscope?: {};
```
mattmurph9-zz, line 64: What is the reason the name is currentuserscope and not currentUserScope?

mattmurph9-zz (approved):
LGTM!

src/Groups.ts
ghost: This is the name used in the REST api unfortunately.

@ngerstle-cognite Any reason for why this is not camel case?

ngerstle-cognite: not that I recall off the top of my head. 
fixing it is an annoying breaking change, but could be done. It's just not worth the effort any time soon.

PR #27, opened 2019-02-06
v1.4.3 - Made metadata optional in ThreeD.listNodes, removed docs step from README

anderscognite:
The `/3d/nodes` API endpoint has an optional `metadata` key (see https://doc.cognitedata.com/api/0.6/#operation/getNodes), but the SDK required them. This is fixed in this PR.

The `yarn docs` is no longer needed as Travis does it for us, so I removed that from the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,12 +122,11 @@ We use `jest` to run tests, see [their documentation](https://github.com/faceboo
 How to release a new version:
 
 1. Create a new branch
-2. Run `yarn docs` to update the documentation
-3. Commit changes (if any)
-4. Run
+2. Commit changes (if any)
+3. Run
     ```bash
     $ npm version [patch/minor/major]
     ```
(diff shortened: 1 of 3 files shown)


src/__tests__/ThreeD.test.ts
```diff
@@ -109,7 +109,6 @@ const nodes = [
     subtreeSize: 13,
     boundingBox: { min: [12, 3, -19], max: [18, 6, 0] },
     sectorId: 1235,
-    metadata: {},
```
ghost: This line should not be removed? All nodes will have an empty metadata object.

anderscognite: I agree :) `git commit --amend && git push --force`.

ghost (approved):
LGTM

PR #32, opened 2019-02-08
Expose loggedIn status from authorize

valgussev:
The only prop which is not exposed from `authorize` `loginStatus` is `loggedIn` which can bring a user information about the logging status


ghost:
@valerii-cognite But that value will always be `true`?

valgussev:
Right, but how application should know that it is authenticated? Just by verifying if `authResult` is not `null`? Can we assume that?

PR #36, opened 2019-02-14
Added access to response headers from search

ghost:
```js
const result = await sdk.Assets.search({ ... });
const { headers, status } = sdk.getMetadata(result);
```

Functionality added to be able to access the header `X-Cognite-Fingerprint` in `*/search`.
Problem discussed here: https://github.com/cognitedata/operational-intelligence/issues/2050

Summary: The problem we are facing is that the SDK didn't have any support to read response headers. This was important for the CDP `/search` endpoints. This PR adds support to read the headers with the interface in the code snippet above.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,15 @@ See [examples](./examples).
 - [SDK documentation](https://js-sdk-docs.cogniteapp.com/)
 - [API documentation](https://doc.cognitedata.com)
 
+## Response headers
+
+On `Assets.search`, `Events.search`, `Files.search` and `TimeSeries.search` it is possible to read the response headers from the server with the following code:
+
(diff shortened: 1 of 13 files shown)


evancharlton:
FYI: readers will not be able to see cognitedata/operational-intelligence#2050 so you should get in the habit of summarizing it here.

evancharlton (approved):
It looks like your prettier (or whatever) config went crazy and added a bunch of extra lines to the delta.

ghost:
> It looks like your prettier (or whatever) config went crazy and added a bunch of extra lines to the delta.

Yes, but I also modified the original lines so had to wrap the line

PR #48, opened 2019-02-27
Don't resolve Login.authorize on browser redirect

### LINKED ISSUE #47: Don't resolve sdk.Login.authorize on redirect
`sdk.Login.authorize` will return `Promise<undefined>` if it triggers a redirect

ghost:
Fixes #47

### DIFF
--- a/src/Login.ts
+++ b/src/Login.ts
@@ -174,9 +174,12 @@ export class Login {
     } catch (e) {
       //
     }
-    Login.loginWithRedirect(params);
-    // @ts-ignore (will do a browser redirect so nothing to care about)
-    return;
+
+    // Login.authorize should never resolve on browser redirect (to avoid return value undefined)
+    // @ts-ignore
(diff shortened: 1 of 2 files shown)


src/Login.ts
```diff
@@ -174,9 +174,11 @@ export class Login {
     } catch (e) {
       //
     }
-    Login.loginWithRedirect(params);
-    // @ts-ignore (will do a browser redirect so nothing to care about)
-    return;
+
+    // Login.authorize should never resolve on browser redirect (to avoid return value undefined)
+    return new Promise(() => {
```
valgussev, line 180: Will client receive Promise instead of undefined so it should still process it?

ghost: Yes, but they will receive a promise anyway (async function).

valgussev (approved):
As long as tests are passing lgtm

PR #53, opened 2019-03-01
Added option to cancel auto-authorize

### LINKED ISSUE #44: Be able to stop scheldueRenewal in Login.authorize

ghost:
```js
const authResult = await Login.authorize({...});
Login.stopAutoAuthorize(authResult);
```

Fixes #44

### DIFF
--- a/src/Login.ts
+++ b/src/Login.ts
@@ -85,7 +85,7 @@ function isString(value: any): boolean {
   return typeof value === 'string';
 }
 
-let cancelSchedule = () => {};
+const scheduleList = new Map<AuthResult, () => void>(); // access token -> function
 
 /**
  * @hidden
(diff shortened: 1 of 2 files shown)


src/Login.ts
```diff
@@ -104,6 +104,14 @@ export class Login {
     return url;
   }
 
+  public static stopAutoAuthorize(authResult: AuthResult): void {
```
valgussev, line 107: Why not just accessToken?

ghost: Yes I could but then we can't use WeakMap but maybe that is ok?

ghost: I will take a look

ghost:
Closed in favour of #54

PR #58, opened 2019-03-01
Use idToken for renewal

ghost:


src/Login.ts
```diff
@@ -331,14 +344,35 @@ export class Login {
     });
   }
 
-  private static async verifyAccessToken(
-    accessToken: string
+  private static async verifyTokens(
```
valgussev, line 347: I would rename this in something more meaningful since we are never verify accessToken anymore

valgussev (approved):

src/Login.ts
ghost: What do you mean? Which name?

valgussev: In `verifyTokens` method only `idToken` is verified. `accessToken` just passed in and out without the verfication, so it seems a method name is missleading

ghost: No, both are verified? https://github.com/cognitedata/cognitesdk-js/pull/58/files/75ff0a680cd54bff4db586965e7b7386468b9982#diff-44266178f674855837c2cb06417aec76R361

ghost: Actually the combination is verified

Issue #61, opened 2019-03-06
ScheduleRenewel is called constantly every 5 seconds


valgussev:
From time to time the problem occurs where `ScheduleRenewel` is getting triggered every 5s when `params` is passed into `authorize` method. Usually the problem can be solved but passing `undefined` as a `param` prop.

PR #63, opened 2019-03-06
v1.5.4 - Fix authorize problems

### LINKED ISSUE #62: 401 during calling `authorize`
If to pass `param` into `authorize` starting from some point of time an app receives back 401 error with the following requests stacking into an unresolved promise.
```
VM66:1 GET https://api.cognitedata.com/login/status 401 
```
```
https://api.cognitedata.com/login/token?token=ewogICJhbGciOiAiUlMyNTYiLAogICJ0eXAiOiAiSldUIgp9.ewogICJ1bmlxdWVfbmFtZSI6ICJ2YWxlcmlpLmd1c2V2QGNvZ25pdGUuY29tIiwKICAicHJvamVjdF9uYW1lIjogImFrZXJicCIsCiAgInNpZ25pbmdfa2V5IjogIjc0MDc4MWYzLTU2MTYtNGJmMC04NzY0LTAwMmZkNmZhNzFkMiIsCiAgImV4cGlyZV90aW1lIjogMTU1MTg3MTgyMgp9.RwYwilucNYp2jnrJoEkKWgFeq3ALg4KiJ4v16O8Tho5nMH9r%2B2B1QtXZlHr7ov3VCcrV7jf4g7%2Ba6GBNDS6svaWz8vqOltulGUp59i8pWMaplP6%2BvSzcr2V8U3PEtMwajNLxa6teD0kDoeLhIkY4h9%2Fe3tGNVerb0a9MnXuQjL7xTcALuqRTxlk8BWsd%2FlJxsbBW2LeevYftmDwA4ZGgh28rKM%2F%2BovuPq3Tq7nPS6%2Fr6AaY%2FoEuc3at%2FEcm1CYMgTVq97B5hfJKfU36MVZwCKAF75ASKui%2F1%2FgksMZWrS7PfWVeS6tBstT9z9XP1%2FGuGcSF%2FCBUCkASo9%2BQj7EXtuQ%3D%3D  401
```

ghost:
- Don't export an es-module file since `jwtDecode` doesn't work with it
- Clear `Authorization` before using endpoint `/login/token` to avoid `401`s
- Call `cancelSchedule` in `scheduleRenewal` to avoid infinite login loop
- Throw exception on invalid `idToken`

Should (hopefully) fix:
Fixes #62, #61

### DIFF
--- a/src/Login.ts
+++ b/src/Login.ts
@@ -3,7 +3,7 @@
 import { AxiosResponse } from 'axios';
 import * as jwtDecode_ from 'jwt-decode';
 import { parse, stringify } from 'query-string';
-import { configure, isBrowser, rawGet, setBearerToken } from './core';
+import { configure, instance, isBrowser, rawGet, setBearerToken } from './core';
 import { scheduleTask } from './helpers/scheduler';
 
 // https://github.com/jvandemo/generator-angular2-library/issues/221#issuecomment-387462303
@@ -235,9 +235,12 @@ export class Login {
   public static async validateJWT(token: string): Promise<TokenStatus> {
     const url = `${loginUrl()}/token`;
     const params = { token };
+    const authHeader = instance.defaults.headers.Authorization;
+    delete instance.defaults.headers.Authorization;
     const response = (await rawGet(url, { params })) as AxiosResponse<
       TokenStatusResponse
     >;
+    instance.defaults.headers.Authorization = authHeader;
     return response.data.data;
   }
 
@@ -309,6 +312,7 @@ export class Login {
 
     cancelSchedule = scheduleTask(
       () => {
+        cancelSchedule();
         Login.authorize(params, tokenCallback);
       },
       expireTimeInMs - timeLeftToRenewInMs,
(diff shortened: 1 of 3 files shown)


ghost (approved):
LGTM :+1:

src/Login.ts
```diff
@@ -235,9 +235,12 @@ export class Login {
   public static async validateJWT(token: string): Promise<TokenStatus> {
     const url = `${loginUrl()}/token`;
     const params = { token };
+    const authHeader = instance.defaults.headers.Authorization;
```
valgussev, line 238: We can use `setBearerToken('')` here instead

ghost: That doesn't remove the header. We also need to restore the previous `Authorization`-header after the request

valgussev: Just set `setBearerToken(accessToken);` only after `Login.validateJWT(idToken)`

ghost: It will not work since if you loggin the second time the Authorization header is already set. The request will fail if `Authorization: Bearer `. It needs to be completely removed.

valgussev (approved):

src/Login.ts
valgussev: Alright, hope you know what you are doing :)

Issue #72, opened 2019-03-27
413 (Payload to large)


valgussev:
for the ref https://github.com/cognitedata/operations-support/issues/1637

ghost:
@valerii-cognite This is solved?

valgussev:
From what I have - yes

PR #76, opened 2019-04-17
Remove id_token, access_token from url every time to avoid long login urls

ghost:
Ref this issue #72

### DIFF
--- a/src/Login.ts
+++ b/src/Login.ts
@@ -111,8 +111,8 @@ export class Login {
   //   1. Check if the token is valid (if not ignore jump to step 3)
   //   2. Return AuthResult and set bearer token
   // 2. Check if error is present in the URL (if yes, throw error with error_description)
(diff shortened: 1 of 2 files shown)


src/Login.ts
```diff
@@ -162,12 +162,13 @@ export class Login {
 
     // Step 2
     const authTokens = Login.parseQuery(window.location.search); // this line can throw
+    // Step 2.1
+    let url = window.location.href;
+    url = removeParameterFromUrl(url, ACCESS_TOKEN);
+    url = removeParameterFromUrl(url, ID_TOKEN);
+    window.history.replaceState(null, '', url);
```
valgussev, line 169: What if we passed in tokens and returned on the step 1.2? Should we also remove tokens from url?

ghost: If that is the case then we should ideally not ever have any id_token / access_token in the URL :)

valgussev (approved):

PR #79, opened 2019-04-24
Added option to authorise using popup window

ghost:
Now you can authorise using popup windows as seen in this example: https://github.com/cognitedata/cognitesdk-js/blob/f1/autorize_with_popup/examples/react/basic/src/App.js


martincognite (approved):
codecov is unhappy, other than that this looks great!

PR #81, opened 2019-05-02
Handle 401 from API on login and avoid 413

ghost:
- Handle errors from API when checking if an access token is valid
- Make sure `access_token` and `id_token` is removed from url before redirect with CDF

### DIFF
--- a/src/Login.ts
+++ b/src/Login.ts
@@ -156,11 +156,15 @@ export class Login {
       accessToken: string,
       idToken: string
     ): Promise<null | AuthResult> => {
-      const authResult = await Login.verifyTokens(accessToken, idToken);
-      if (authResult !== null) {
-        configure({ project: authResult.project });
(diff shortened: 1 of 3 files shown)


src/Login.ts
```diff
@@ -237,7 +241,17 @@ export class Login {
...
     // @ts-ignore
     return new Promise(() => {
-      Login.loginWithRedirect(loginParams);
+      // make sure tokens is removed from url
+      const removeTokens = (uri: string) =>
+        removeParameterFromUrl(
+          removeParameterFromUrl(uri, ACCESS_TOKEN),
+          ID_TOKEN
+        );
+      Login.loginWithRedirect({
```
valgussev, line 250: What if this fail as well?

valgussev (approved):

src/Login.ts
ghost: What do you mean?

PR #84, opened 2019-05-07
Don't leak tokens to other domains

### LINKED ISSUE #83: rawGet(), rawPost() may leak authentication headers
The `rawGet()`, `rawPost()` methods on the current SDK (1.7.2) accepts absolute URLs that are not pointing to CDF. When used like this, it will send along the authentication headers to the non-CDF domain, which seems to me like a security risk.

Sample code:
```js
// assume project, apiKey defined

const sdk = require('@cognite/sdk');
sdk.configure({ project, apiKey });
sdk.rawGet('http://localhost:8888/');

// localhost:8888 will receive an "Api-Key" header with the CDF API key
```

@f1cognite

ghost:
Fixes #83

### DIFF
--- a/src/__tests__/core.test.ts
+++ b/src/__tests__/core.test.ts
@@ -238,14 +238,14 @@ describe('Core', () => {
     const token = 'abc';
     const nonPresentTokenChecker = (config: AxiosRequestConfig) => {
       if (config.headers.Authorization !== undefined) {
-        return [404];
+        return [400];
       }
       return [200];
     };
 
     const presentTokenChecker = (config: AxiosRequestConfig) => {
       if (config.headers.Authorization !== `Bearer ${token}`) {
-        return [404];
+        return [400];
       }
       return [200];
     };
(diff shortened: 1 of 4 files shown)


src/core.ts
```diff
@@ -110,6 +110,23 @@ instance.interceptors.request.use(
...
 
+instance.interceptors.request.use((config: AxiosRequestConfig) => {
+  const url = config.url || '';
+  const baseUrl = config.baseURL || '';
+  const cogniteUrl = /^https:\/\/.*\.cognitedata.com/;
+  const isUnknownDomain =
+    // rawGet('https://example.com');
+    (/^http/.test(url) && !cogniteUrl.test(url)) ||
+    // rawGet('/path') with baseUrl !== cognite
+    (!cogniteUrl.test(baseUrl) && !cogniteUrl.test(url));
```
ghost: Lots of false positives and negatives here:

| URL | `isUnknownDomain` | Expected value | Comment |
|-|-|-|-|
| `http/path` | `true` | False | False positive
| `//example.com/path` | `false` | True | **False negative**
| `https://api.cognitedata.com.my-evil-domain.com/path` | `false` | True | **False negaitve**
| `https://api.cognitedata-com.com/path` | `false` | True | **False negaitve**
| `https://my-evil.domain.com/cognitedata.com/path` | `false` | True | **False negative**
| `ftp://my-evil-domain.com` | `false` | True | False negative<sup>1</sup>

<sub><sup>1</sup>This is less important though since major browsers (as far as I know) don't support non-local, non-HTTP(S) URLs in fetch requests.</sub>

Also, it will give false positives when `baseUrl` is set to a non-CDF domain (e.g. when testing locally with `localhost`).

ghost (changes requested):

src/core.ts
ghost: My suggested alternative check (with [`url`](https://www.npmjs.com/package/url) package, for instance):

```js
function isSameOrigin(baseUrl, newUrl) {
  const url = require('url');
  const baseUrlParsed = url.parse(baseUrl);
  const newUrlParsed = url.parse(url.resolve(baseUrl, newUrl));

  // check that protocol and hostname are the same
  return baseUrlParsed.protocol === newUrlParsed.protocol
    && baseUrlParsed.host === newUrlParsed.host;
}

const isUnknownDomain = !isSameOrigin(baseUrl, url);
```

src/__tests__/core.test.ts
```diff
@@ -255,14 +255,34 @@ describe('Core', () => {
...
+      ];
 
-      mock.onGet('/test').reply(presentTokenChecker);
-      await rawGet('/test');
+      const promises: Promise<any>[] = [];
+      tests.forEach(([url, shouldTokenPresent, mockUrl]) => {
+        const tester = shouldTokenPresent
+          ? presentTokenChecker
+          : nonPresentTokenChecker;
+        mock.onGet(mockUrl ? mockUrl : url).reply(tester);
```
ghost, line 281: What's the purpose of the `mockUrl` variable here?

ghost: When requesting `https://api.cognitedata.com.my-evil-domain.com/path` this will result into the url: `.my-evil-domain.com/path` for the request and I get a 404 on this line: https://github.com/cognitedata/cognitesdk-js/pull/84/files/368bed1847cbb5e5c96d6cfeed56795df29319c6..0ad8a761db349bf59761e7a0bdf07bbc01a8129a#diff-591ac955057de4587f59bc8a4b91b40dR281

ghost (approved):

PR #86, opened 2019-05-10
Uncomment Files integration test code

ghost:


polomani (approved):
LG
Although, later I would suggest to spit `'create,retrieve,update,delete` into independent tests.
In that case we see which one exactly is failing

ghost:
@polomani Totally agree! I will do that

PR #87, opened 2019-05-11
Added Raw API

ghost:


src/resources/raw.ts
```diff
@@ -0,0 +1,228 @@
...
+   * [Create tables in a database](https://doc.cognitedata.com/api/v1/#operation/createTables)
+   *
+   * ```js
+   * const tables = await client.raw.createTables('My company', [{ name: 'Customers' }]);
+   * ```
+   */
+  createTables: (
+    databaseName: string,
+    items: RawDBTable[],
+    // tslint:disable-next-line:bool-param-default
```
polomani, line 74: why did you disable it here?
doesn't `false` means the same for us as `undefined`?
(I mean we won't create db in both cases, so it might be ok to set `default=false`)

src/resources/raw.ts
```diff
@@ -0,0 +1,228 @@
...
+    listRows: (databaseName, tableName, query) => {
+      const endpoint = generateListEndpoint<ListRawRows, RawDBRow>(
+        instance,
+        rowPath(databaseName, tableName),
+        map,
+        false
+      );
+      if (query && query.onlyRowKeys) {
+        // @ts-ignore - we transform it
+        return endpoint({ columns: ',' });
```
polomani, line 188: what does this comma do here?

ghost: It says in the documentation: "Ordered list of column keys, separated by commas. Leave empty for all, use single comma to retrieve only row keys."

src/resources/raw.ts
ghost: I tried both:
- `ensureParent: boolean = false` (it fails since you can also have an initialiser in the function declaration and not in typescript declaration.
- `ensureParent: boolean | undefined` (but ts-lint also complained about this one). So thought it was a bug in ts-lint.

You can try out yourself on that branch, but it was always complaining

polomani (approved):

PR #92, opened 2019-05-15
Added Projects API

ghost:
We are missing update projects endpoint, but we have a test for it that is failing and waiting for them to fix: https://github.com/cognitedata/service-contracts/issues/227


ghost:
Wrong target branch

PR #93, opened 2019-05-15
Added Projects API

ghost:
We are missing update projects endpoint, but we have a test for it that is failing and waiting for them to fix: cognitedata/service-contracts#227

### DIFF
--- a/src/standardMethods.ts
+++ b/src/standardMethods.ts
@@ -287,13 +287,15 @@ export function generateSingleReplaceEndpoint<RequestType, RepsonseType>(
   resourcePath: string,
   metadataMap: MetadataMap
 ) {
-  return async function replace(change: RequestType): Promise<RepsonseType> {
+  return async function replace(
+    replacement: RequestType
+  ): Promise<RepsonseType> {
     const response = await rawRequest<RepsonseType>(
       axiosInstance,
       {
         method: 'put',
         url: resourcePath,
-        data: change,
+        data: replacement,
       },
       true
     );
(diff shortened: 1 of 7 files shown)


src/standardMethods.ts
```diff
@@ -280,3 +280,23 @@ export function generateInsertEndpoint<RequestParams>(
...
   };
 }
+
+/** @hidden */
+export function generateSingleReplaceEndpoint<RequestType, RepsonseType>(
+  axiosInstance: AxiosInstance,
+  resourcePath: string,
+  metadataMap: MetadataMap
+) {
+  return async function replace(change: RequestType): Promise<RepsonseType> {
```
polomani: maybe better name for `change`, like `replacement`?

src/types/types.ts
```diff
@@ -793,6 +793,110 @@ export interface ServiceAccountInput {
...
+export type ProjectName = string;
+
+/**
+ * The url name of the project. This is used as part of API calls. It should only contain letters, digits and hyphens, as long as the hyphens are not at the start or end.
+ * @example publicdata
+ */
+export type UrlName = string;
+
+/**
+ * A default group for all project users. Can be used to establish default capabilities.WARNING: this group may be logically deleted
```
polomani: space before warning

src/standardMethods.ts
ghost: Fixed

src/types/types.ts
ghost: Fixed

polomani (approved):

PR #94, opened 2019-05-15
Added API keys API

ghost:


src/__tests__/resources/apiKeys.integration.spec.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019 Cognite AS
+
+import { API } from '../../resources/api';
+import { NewApiKeyResponseDTO, ServiceAccount } from '../../types/types';
```
polomani: I believe we don't put `DTO` in the end of DTOs explicitly

src/standardMethods.ts
```diff
@@ -127,6 +127,29 @@ export function generateListNoCursorEndpoint<ResponseType>(
...
+        method: 'get',
+        url: resourcePath,
+        params,
+      },
+      true
+    );
+    return metadataMap.addAndReturn(response.data.items, response);
+  };
+}
+
```
polomani, line 152: you have the same method but with no params
why not making params optional?
problems dynamic typing?

src/types/types.ts
```diff
@@ -1018,6 +1018,36 @@ export interface AssetMapping3DBase {
...
+  id: CogniteInternalId;
+  serviceAccountId: CogniteInternalId;
+  createdTime: Date;
+  status: 'ACTIVE' | 'DELETED';
+}
+
+export interface ApiKeyListScope {
+  all?: boolean;
+  serviceAccountId?: CogniteInternalId;
+  includeDeleted?: boolean;
```
polomani, line 1155: it has some explanation in spec, why didn't add it here?

src/standardMethods.ts
ghost: Yes, dynamic typing.. 🤔 Haven't found a good way to do it

polomani (approved):

PR #96, opened 2019-05-15
Improved tests and support for Date objects

ghost:


src/__tests__/resources/revisions3D.integration.spec.ts
```diff
@@ -1,173 +1,216 @@
...
+import {
+  getSortedPropInArray,
+  retryInSeconds,
+  setupClient,
+  simpleCompare,
+} from '../testUtils';
+
+// suggested solution/hack for conditional tests: https://github.com/facebook/jest/issues/3652#issuecomment-385262455
+const describeIfCondition =
+  process.env.REVISION_3D_INTEGRATION_TEST === 'true'
```
polomani, line 26: add in readme maybe?

src/__tests__/resources/revisions3D.integration.spec.ts
```diff
@@ -1,173 +1,216 @@
...
+
+      const fileContent = readFileSync('src/__tests__/test3dFile.fbx');
+      [[model], file, assets] = await Promise.all([
+        // create 3D model
+        client.models3D.create([{ name: `Model revision test ${now}` }]),
+        // upload file
+        client.files.upload(
+          { name: `file_revision_test_${now}.fbx` },
+          fileContent,
+          true
```
polomani, line 63: oh this is so nice (hope this `true` actually works)

src/__tests__/testUtils.ts
```diff
@@ -71,3 +71,12 @@ export function getSortedPropInArray<T extends { [key: string]: any }>(
 ) {
   return arr.map(elem => elem[propName]).sort(simpleCompare);
 }
+
+export function string2arrayBuffer(str: string) {
+  const buf = new ArrayBuffer(str.length * 2); // 2 bytes for each char
```
polomani, line 77: nice copy-paste xD

src/axiosWrappers.ts
```diff
@@ -44,15 +45,18 @@ export function listenForNonSuccessStatusCode(
...
 ) {
   if (isSafeToRetry) {
     makeRequestSafeToRetry(requestConfig);
   }
-  return axiosInstance(requestConfig).catch(handleErrorResponse) as Promise<
-    AxiosResponse<ResponseType>
-  >;
+  requestConfig.data = transformDateInRequest(requestConfig.data);
+  const response = (await axiosInstance(requestConfig).catch(
+    handleErrorResponse
```
polomani: what I don't like here is that you mix `.then(), .catch()` syntax and `await, try{}` for promises. I mean, you take a style from both of them and it's a bit confusing to my mind.
I've used to use either one or another. But let's talk about this more.

src/resources/files.ts
```diff
@@ -49,7 +51,9 @@ export interface FilesAPI {
    */
   upload: (
     metadata: ExternalFilesMetadata,
-    fileContent?: FileContent
+    fileContent?: FileContent,
+    // tslint:disable-next-line:bool-param-default
+    waitUntilAcknowledged?: boolean
```
polomani, line 56: this is so ugly:(
hope we'll find better solution later

src/resources/files.ts
```diff
@@ -103,14 +107,40 @@ export interface FilesAPI {
...
+  return new Promise((resolve, reject) => {
+    try {
+      const retrieve = generateRetrieveSingleEndpoint<
+        CogniteInternalId,
+        FilesMetadata
+      >(axiosInstance, resourcePath, new MetadataMap());
+
+      const myInterval = setInterval(async () => {
+        if ((await retrieve(fileId)).uploaded) {
+          clearInterval(myInterval);
```
polomani, line 133: 1) will it wait forever if something bad happen while uploading? or if the file is too big?
2) what if status is `failed`?

src/resources/files.ts
```diff
@@ -134,6 +164,9 @@ function generateUploadEndpoint(
         data: fileContent,
       });
     }
+    if (waitUntilAcknowledged) {
+      await waitUntilFileIsUploaded(file.id, axiosInstance, resourcePath);
+    }
```
polomani, line 182: should't we return a newer information here? at least `uploaded` is true now

src/utils.ts
```diff
@@ -1,5 +1,6 @@
 // Copyright 2019 Cognite AS
 
+import { cloneDeepWith } from 'lodash';
```
polomani, line 3: do we actually need a lodash in here?
I mean the library is like 75 kb and we can reduce it by just having a few helper functions
All other places we use it are quite simple
At this place we can just do `JSON.parse(JSON.stringify(obj))` can't we?

src/utils.ts
```diff
@@ -87,3 +88,34 @@ export function makePromiseCancelable<T>(
   };
   return cancelablePromise as CancelablePromise<T>;
 }
+
+export function convertDateToTimestamp(date: Date) {
+  return date.getTime();
```
polomani: the name of the function is longer than it's actual code xD
can't we just do `date.getTime()` instead of `convertDateToTimestamp(date)` :) ?

src/utils.ts
```diff
@@ -87,3 +88,34 @@ export function makePromiseCancelable<T>(
...
+}
+
+export function convertTimestampToDate(timestamp: number) {
+  return new Date(timestamp);
```
polomani: same

src/utils.ts
```diff
@@ -87,3 +88,34 @@ export function makePromiseCancelable<T>(
...
+}
+
+export function transformDateInRequest(data: any) {
+  return cloneDeepWith(data, value => {
+    if (value instanceof Date) {
+      return value.getTime();
```
polomani, line 95: as I am saying, better than `convertDateToTimestamp(date)`

src/utils.ts
polomani: and to iterate recurcively
```
function eachRecursive(obj)
{
    for (var k in obj)
    {
        if (typeof obj[k] == "object" && obj[k] !== null)
            eachRecursive(obj[k]);
        else
            // check for date
    }
}
```

src/__tests__/resources/revisions3D.integration.spec.ts
ghost: I will

src/axiosWrappers.ts
ghost: I will clean up

src/resources/files.ts
ghost: So status can never be failed? But the upload process can fail, but that is catched before this function should be called.

src/resources/files.ts
ghost: Good catch! Yes. I will fix

src/utils.ts
ghost: So one thing is to clone, another thing is to replace objects like Date. `JSON.parse(JSON.stringify(obj))` will remove all Date objects with a string which will be harder to work with for us. And when including `cloneDeepWith` from lodash almost all build systems will apply tree shaking to only include what is needed in the final build

src/utils.ts
ghost: Hehe. True.

src/utils.ts
ghost: Agree

src/resources/files.ts
polomani: hmm ok

polomani (approved):

PR #97, opened 2019-05-15
Added Groups API

ghost:


src/__tests__/resources/groups.integration.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+            eventsAcl: { actions: ['READ'], scope: { all: {} } },
+          },
+        ],
+      },
+    ];
+    const response = await client.groups.create(groupsToCreate);
+    expect(response.length).toBe(1);
+    [group] = response;
+    expect(group.name).toBe(groupsToCreate[0].name);
+    await sleepPromise(10000);
```
polomani: 10 seconds for what?

polomani: why 10 seconds here?

src/types/types.ts
```diff
@@ -1166,6 +1166,205 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+
+export interface GenericAclAllScope {
+  all: {};
+}
+
+export interface GenericAclCurrentUserScope {
+  currentuserscope: {};
+}
+
+export type cognitegroups_aclAction = LIST | READ | CREATE | UPDATE | DELETE;
```
polomani: underscore?

src/types/types.ts
```diff
@@ -1166,6 +1166,205 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+export type cognitegroups_aclScope =
+  | GenericAclAllScope
+  | GenericAclCurrentUserScope;
+export type cognitegroups_aclAcl = Acl<
```
polomani: the naming is really bad.. It's like reading C code
what does `aclAcl` mean? underscore?

src/types/types.ts
```diff
@@ -1166,6 +1166,205 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+
+export interface CogniteAssetsAclIdScope {
+  assetIdScope: {
+    subtreeIds: CogniteInternalId[];
+  };
+}
+export type cognitetimeseries_aclAction = READ | WRITE;
+export type cognitetimeseries_aclScope =
+  | GenericAclAllScope
+  | CogniteAssetsAclIdScope;
```
polomani: Do we need to have Cognite as a prefix?

src/types/types.ts
```diff
@@ -1166,6 +1166,205 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+export type cogniteassets_aclAction = LIST | WRITE;
+export type cogniteassets_aclScope = GenericAclAllScope;
+export type cogniteassets_aclAcl = Acl<
+  cogniteassets_aclAction,
+  cogniteassets_aclScope
+>;
+
+export type cogniteevents_aclAction = READ | WRITE;
+export type cogniteevents_aclScope = GenericAclAllScope;
+export type cogniteevents_aclAcl = Acl<
```
polomani: what's the purpose of having cognite prefix?

src/types/types.ts
```diff
@@ -1218,123 +1218,80 @@ export interface GenericAclCurrentUserScope {
...
+export type ApiKeysAclScope = GenericAclAllScope | GenericAclCurrentUserScope;
+export type ApiKeysAcl = Acl<ApiKeysAclAction, ApiKeysAclScope>;
 
-export type cognitethreed_aclAction = READ | CREATE | UPDATE | DELETE;
-export type cognitethreed_aclScope = GenericAclAllScope;
-export type cognitethreed_aclAcl = Acl<
-  cognitethreed_aclAction,
-  cognitethreed_aclScope
->;
+export type ThreeDAclAction = READ | CREATE | UPDATE | DELETE;
```
polomani: how about `AclAction3D`?

src/types/types.ts
```diff
@@ -1218,123 +1218,80 @@ export interface GenericAclCurrentUserScope {
...
+export type ThreeDAcl = Acl<ThreeDAclAction, ThreeDAclScope>;
 
-export type cognitesequences_aclAction = READ | WRITE;
-export type cognitesequences_aclScope = GenericAclAllScope;
-export type cognitesequences_aclAcl = Acl<
-  cognitesequences_aclAction,
-  cognitesequences_aclScope
->;
+export type SequencesAclAction = READ | WRITE;
+export type SequencesAclScope = GenericAclAllScope;
```
polomani: Also might be better to have `AclScope` and `AclAction` as a prefix. In that case we can put 3d as a postfix and it will be consistent. Also it will be all grouped under `Acl` kinda. But let's discuss

polomani (changes requested):

src/types/types.ts
```diff
@@ -1166,6 +1166,162 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+
+export type AclActionFiles = READ | WRITE;
+export type AclScopeFiles = GenericAclAllScope;
+export type AclFiles = Acl<AclActionFiles, AclScopeFiles>;
+
+export type AclActionUsers = LIST | CREATE | DELETE;
+export type AclScopeUsers = GenericAclAllScope;
+export type AclUsers = Acl<AclActionUsers, AclScopeUsers>;
+
+export type ProjectsclAction = LIST | READ | CREATE | UPDATE;
```
polomani: this one left?

src/types/types.ts
```diff
@@ -1166,6 +1166,162 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+export type AclSecurityCategories = Acl<
+  AclActionSecurityCategories,
+  AclScopeSecurityCategories
+>;
+
+export type AclActionRaw = READ | WRITE | LIST;
+export type AclScopeRaw = GenericAclAllScope;
+export type AclRaw = Acl<AclActionRaw, AclScopeRaw>;
+
+export interface AssetsAclIdScope {
```
polomani: and this?

src/types/types.ts
```diff
@@ -1166,6 +1166,162 @@ export interface NewApiKeyResponse extends ApiKeyObject {
...
+export type AclActionEvents = READ | WRITE;
+export type AclScopeEvents = GenericAclAllScope;
+export type AclEvents = Acl<AclActionEvents, AclScopeEvents>;
+
+export type AclActionFiles = READ | WRITE;
+export type AclScopeFiles = GenericAclAllScope;
+export type AclFiles = Acl<AclActionFiles, AclScopeFiles>;
+
+export type AclActionUsers = LIST | CREATE | DELETE;
+export type AclScopeUsers = GenericAclAllScope;
```
polomani: and in that case this one as well

src/types/types.ts
polomani: and there should be `A` in `Acl`

polomani (changes requested):
sorry

polomani (approved):
LGTM, fixed the ones which didn't LGTM xD

PR #106, opened 2019-05-22
feat: Automatic chunking when posting resources

ghost:
When posting more than > 1000 timeseries, events, datapoints and assets the SDK will automatically split the the request up into chunks of 1000-elements do avoid breaking the limit in the API.

For assets it will also chunk it up to valid requests based on topological sort.

### DIFF
--- a/src/__tests__/resources/assets.unit.spec.ts
+++ b/src/__tests__/resources/assets.unit.spec.ts
@@ -1,0 +1,69 @@
+// Copyright 2019 Cognite AS
+
+import { assetChunker } from '../../resources/assets';
+
+describe('Asset unit test', async () => {
+  describe('assetChunker', () => {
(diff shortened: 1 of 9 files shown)


src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      dependencies.set(assetAB, assetA);
+      dependencies.set(assetAAA, assetAA);
+      dependencies.set(assetAAB, assetAA);
+
+      const visitedAssets = new Set();
+
+      chunks.forEach(chunk => {
+        chunk.forEach(asset => {
+          if (dependencies.has(asset)) {
+            expect(visitedAssets.has(dependencies.get(asset))).toBeTruthy();
```
polomani, line 62: gooood

polomani (approved):

PR #124, opened 2019-06-01
feat: Made client creation synchronous

### LINKED ISSUE #122: Simplify creation of a client
Today there are two ways of creating a client:
- `createClientWithApiKey`
- `createClientWithOAuth`

Both of the functions return a promise which makes it a bit harder to create a client than if it didn't return a promise. We should make the two functions synchronous-functions:
```js
const client = createClientWithApiKey({ project, apiKey });
```

This will be a **breaking change**

ghost:
Fixes #122

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ The SDK is written in native typescript, so no extra types needs to be defined.
 import * as sdk from '@cognite/sdk';
 
 async function quickstart() {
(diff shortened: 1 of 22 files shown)


ghost:
FYI @martincognite

polomani (approved):
cool
less code - less issues

PR #128, opened 2019-06-02
Feat/alphabetized types

polomani:
I made a script which sorts declarations in file alphabetically. (this feature is not supported by prettier yet, I didn't find any other solution)
And then ran it for types.ts.
This is needed to be able to check for interfaces changes easily.
And also makes easier to avoid conflicts.

### DIFF
--- a/scripts/alphabetizeTypes.js
+++ b/scripts/alphabetizeTypes.js
@@ -10,24 +10,26 @@ const FILENAME = argv.i;
 const HEADER = argv.h;
 
 (async () => {
-    const original = readFileSync(FILENAME).toString();
-    const parsedCode = await parser.parseFile(FILENAME).catch(err => {
+    try {
+        const original = readFileSync(FILENAME).toString();
+        const parsedCode = await parser.parseFile(FILENAME)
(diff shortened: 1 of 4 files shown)


scripts/alphabetizeTypes.js
```diff
@@ -0,0 +1,33 @@
...
+const comments = new Comments();
+const parser = new TypescriptParser();
+
+const FILENAME = argv.i;
+const HEADER = argv.h;
+
+(async () => {
+    const original = readFileSync(FILENAME).toString();
+    const parsedCode = await parser.parseFile(FILENAME).catch(err => {
+        console.error(err);
```
ghost, line 32: Could you wrap the entire function with a catch instead?

ghost (commented):
This is very cool!

ghost:
`yargs` doesn't support node9 and are failing the build.

scripts/alphabetizeTypes.js
polomani: yes

ghost (approved):

PR #133, opened 2019-06-03
test: Added test for handling invalid api-key

ghost:
Fixes #133

### DIFF
--- a/src/retryRequests.ts
+++ b/src/retryRequests.ts
@@ -76,7 +76,7 @@ export function addRetryToAxiosInstance(instance: AxiosInstance) {
       }
 
       if (!err.response) {
-        return true;
+        return false;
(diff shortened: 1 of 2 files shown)


src/retryRequests.ts
```diff
@@ -76,7 +76,7 @@ export function addRetryToAxiosInstance(instance: AxiosInstance) {
       }
 
       if (!err.response) {
-        return true;
+        return false;
```
polomani, line 79: what does this do? won't retry if there is a falsy response?

ghost: Just saying that it shouldn't retry the request if we didn't get a response from the API (we don't know the status code or the request failed)

polomani (approved):

Issue #157, opened 2019-06-13
client.assets.create() posts assets all at once, regardless their relations


polomani:
right now we chunk up assets if the array is >1000 and do `promise.all` for all promises.
Even though all assets are sorted (parents go before their children) it causes an error if there is a child from one chunk having a parent in another.

So for now we need to resolve all the requests sequentially. 

We might need some more advanced logic here later.

Issue #117, opened 2019-05-26
Handle errors during automatic chunk of requests


ghost:
If someone posts many resources and we automatically chunk the requests then we need to handle the case where one of the requests fails to succeed.

Suggestion:
If the request fails during automatic chunking we should throw an error explaining which resources was successfully posted and which one wasn't

ghost:
Take a look at the Python SDK for inspiration: https://cognite-docs.readthedocs-hosted.com/projects/cognite-sdk-python/en/latest/cognite.html#create-an-asset-hierarchy

PR #158, opened 2019-06-13
feat: throw multiError on create assets

polomani:
Solves #117 and #157 

So basically we chunk up `assets.create()` request if it has more than 1000 items.
(later we'll chunk other requests as well, but with different logic)
The error used to say only about the first error happened during those multiple requests.
Now we introduced `CogniteMultiError` which will contain data about all requests (again for this specific case it's always single one).
Although the first error still takes precedence, it will also contain information about what has been created and what has failed.

`PS. need to rebase when base branch is merged`

### DIFF
--- a/src/__tests__/resources/assets.integration.spec.ts
+++ b/src/__tests__/resources/assets.integration.spec.ts
@@ -1,6 +1,8 @@
 // Copyright 2019 Cognite AS
 
 import CogniteClient from '../../cogniteClient';
+import { CogniteError } from '../../error';
+import { CogniteMultiError } from '../../multiError';
 import { Asset } from '../../types/types';
 import { randomInt, setupLoggedInClient } from '../testUtils';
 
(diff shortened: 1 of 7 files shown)


martincognite (approved):
Nice!

src/__tests__/resources/assets.integration.spec.ts
```diff
@@ -35,11 +37,82 @@ describe('Asset integration test', () => {
...
+      extraProp: 'corrupted',
+      parentExternalId: newRootAsset.externalId,
+    };
+    const newChildAsset = {
+      ...childAsset,
+      parentExternalId: newRootAsset.externalId,
+    };
+    const childArr = new Array(999).fill(newChildAsset);
+    try {
+      expect(
```
ghost, line 77: Is the syntax correct here? Shouldn't `await` be outside of expect when you use `.rejects`?

src/__tests__/resources/assets.integration.spec.ts
```diff
@@ -35,11 +37,82 @@ describe('Asset integration test', () => {
...
+        await client.assets.create([newRootAsset, ...childArr, corruptedChild])
+      ).rejects.toBeTruthy();
+    } catch (e) {
+      expect(e).toBeInstanceOf(CogniteMultiError);
```
ghost, line 81: Will this run? Have you checked that `.rejects` in jest doesn't catch the exception?

src/error.ts
```diff
@@ -23,6 +23,7 @@ export class CogniteError extends Error {
       message += `\n${JSON.stringify(extra, null, 2)}`;
     }
     super(message);
+    Object.setPrototypeOf(this, CogniteError.prototype);
```
ghost, line 26: ?

src/multiError.ts
```diff
@@ -0,0 +1,64 @@
...
+
+/** @hidden */
+export interface MultiErrorRawSummary<RequestType, ResponseType> {
+  succeded: RequestType[][];
+  responses: AxiosResponse<ResponseType>[];
+  failed: RequestType[][];
+  errors: (Error | CogniteError)[];
+}
+
+export class CogniteMultiError<RequestType, ResponseType> extends Error {
```
ghost, line 14: Don't want to extend from `CogniteError`?

src/error.ts
polomani: this is a bit nasty way js works
https://stackoverflow.com/questions/51229574/why-instanceof-returns-false-for-a-child-object-in-javascript

ghost: I see. Maybe comment the link in the code?

src/__tests__/resources/assets.integration.spec.ts
polomani: I will refactor it a little bit

polomani: this check doesn't seem to work in any way

src/__tests__/resources/assets.integration.spec.ts
polomani: Yes, but I will add `expect.assertions(8);` to be sure

src/multiError.ts
polomani: `CogniteMultiError` doesn't necessarily have `status` prop
That's basically only one reason

src/error.ts
polomani: yep

PR #160, opened 2019-06-14
fix: ts errors during yarn docs

polomani:
typeScript freaks out during `yarn docs` for some reason
It sounds like a good reason to refactor it as well (and some tests of course)

### DIFF
--- a/src/resources/assets/assetUtils.ts
+++ b/src/resources/assets/assetUtils.ts
@@ -103,18 +103,17 @@ export async function promiseEachInSequence<RequestType, ResponseType>(
   inputs: RequestType[],
   promiser: (input: RequestType) => Promise<ResponseType>
 ) {
-  return inputs.reduce((previousPromise, input, index) => {
-    return previousPromise.then(result => {
-      return promiser(input)
-        .catch(err => {
-          return Promise.reject({
-            errors: [err],
-            failed: inputs.slice(index),
-            succeded: inputs.slice(0, index),
-            responses: result,
-          });
-        })
-        .then(Array.prototype.concat.bind(result));
-    });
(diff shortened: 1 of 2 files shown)


src/resources/assets/assetUtils.ts
```diff
@@ -103,18 +103,17 @@ export async function promiseEachInSequence<RequestType, ResponseType>(
...
-            errors: [err],
-            failed: inputs.slice(index),
-            succeded: inputs.slice(0, index),
-            responses: result,
-          });
-        })
-        .then(Array.prototype.concat.bind(result));
-    });
-  }, Promise.resolve(new Array()));
+  return inputs.reduce(async (previousPromise, input, index) => {
```
martincognite, line 106: I would imagine we have to `Promise.all` this, no?

`return Promise.all(inputs.reduce...)`

polomani: well no
`Promise.all` resolves all promises simultaneously, while we need to resolve them in sequence here.

polomani: No, reduce returns one element

polomani: which will look something like `Promise1().then(Promise2.then(Promise3.then(...)))`

martincognite: True, but then we could return an array of promises. Nevermind -- if this works then crack on

martincognite (approved):

src/resources/assets/assetUtils.ts
polomani: no no, the array won't work.
we need to send requests one by one in here :)

PR #163, opened 2019-06-16
F1/code samples improved

polomani:
needed for https://github.com/cognitedata/service-contracts/pull/235

Improved a script:
 - headers added
 - supports multiple links for single sdk method

Impoved typedocs comments:
 - added multiple links in places where one function corresponds with a few operationIds in API. Commented them out, so they are not visible in sdk typedocs. Only for `doc.cognitedata.com`
- fixed links to API in some places

Fixes https://github.com/cognitedata/cognitesdk-js/issues/162

### DIFF
--- a/scripts/postProcessDocs.js
+++ b/scripts/postProcessDocs.js
@@ -1,17 +1,27 @@
-const jsonDoc = require('../a.json');
+const jsonDoc = require('../docs.json');
 const _ = require('lodash');
+const fs = require('fs');
 
 const docRegEx = /https:\/\/doc.cognitedata.com\/api\/v1\/#operation\/([a-zA-Z0-9]+)/g;
+const header =  `import { CogniteClient } from '@cognite/sdk';\n\nconst client = new CogniteClient({ appId: '[YOUR APP NAME]' });\n\n`;
+
+const operationsWithoutHeader = ['redirectUrl'];
 
 const lookupTable = {};
 _.cloneDeepWith(jsonDoc, (value, _, object) => {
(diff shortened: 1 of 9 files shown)


scripts/postProcessDocs.js
```diff
@@ -0,0 +1,33 @@
+const jsonDoc = require('../docs.json');
+const _ = require('lodash');
+const fs = require('fs');
+
+const docRegEx = /https:\/\/doc.cognitedata.com\/api\/v1\/#operation\/([a-zA-Z0-9]+)/g;
+const header =
+`import { CogniteClient } from '@cognite/sdk';\n
```
ghost, line 7: String templates should support multilines without adding `\n` 🤔 
https://stackoverflow.com/a/36268846/4462088

ghost (commented):
Later we should also add a check if the operationIds was found in the spec and vice versa if we are missing some endpoints

scripts/postProcessDocs.js
polomani: yeah, I know, but I need 2 new lines, and that looks a bit messy
```
const header =
`import { CogniteClient } from '@cognite/sdk';

const client = new CogniteClient({ appId: '[APP NAME]' });
client.loginWithApiKey({
  project: '[PROJECT]',
  apiKey: '[API_KEY]'
});

`;
```

polomani: especially in the bottom

ghost (approved):

PR #176, opened 2019-06-25
feat: add chunking on update, retrieve, remove and insert

eirikdahlen:


src/standardMethods.ts
```diff
@@ -306,19 +336,25 @@ export function generateRetrieveLatestEndpoint<RequestParams, ResponseType>(
...
 ) {
   return async function insert(items: RequestParams[]): Promise<{}> {
-    const response = await rawRequest<ItemsResponse<{}>>(
-      axiosInstance,
-      {
-        method: 'post',
-        url: resourcePath,
-        data: { items },
-      },
+    let chunks: RequestParams[][] = [[]];
```
ghost: I see a lot of duplicate code / logic for all the chunking across functions. Anyway we can refactor this?

src/__tests__/standardMethods.spec.ts
```diff
@@ -133,4 +160,58 @@ describe('standard methods', () => {
...
+      const chunkCounter = countChunks(axiosMock, 'path/update');
+      const response = await update(changes);
+      expect(chunkCounter.value).toBe(3);
+      expect(response.length).toBe(changes.length);
+      expect(metadataMap.get(response)).toBeDefined();
+    });
+  });
+  describe('generateInsertEndpoint', () => {
+    test('automatic chunking for insert', async () => {
+      const metadataMap = new MetadataMap();
```
ghost: Can we refactor all of this? These lines are common for all the tests to reduce number of duplicate lines

ghost (changes requested):
At least one integration test for `generateRetrieveEndpoint`, `generateDeleteEndpoint`, `generateUpdateEndpoint` and `generateInsertEndpoint` would be great to verify that everything is working

src/__tests__/standardMethods.spec.ts
```diff
@@ -35,7 +59,7 @@ describe('standard methods', () => {
   });
 
   describe('generateCreateEndpoint', () => {
-    test('automatic chunking', async () => {
+    test('automatic chunking for create', async () => {
       const metadataMap = new MetadataMap();
       const items = new Array(3002).fill(null).map((_, index) => ({
```
ghost: Use the array function here?

src/__tests__/standardMethods.spec.ts
ghost: @eirikdahlen Still a lot of similar code left

src/standardMethods.ts
```diff
@@ -343,3 +364,14 @@ export function generateSingleReplaceEndpoint<RequestType, RepsonseType>(
     return metadataMap.addAndReturn(response.data, response);
   };
 }
+
+function doChunking<Type>(
```
ghost, line 372: Excellent!

src/__tests__/standardMethods.spec.ts
```diff
@@ -211,20 +201,13 @@ describe('standard methods', () => {
 
 async function tester(
```
eirikdahlen: @f1cognite changed back to the way we talked about on the phone, but still the code doesn't make sense as of now 😢We'll catch up tomorrow

src/__tests__/standardMethods.spec.ts
```diff
@@ -27,6 +30,26 @@ function numberGeneratorAPIEndpoint(config: AxiosRequestConfig) {
...
+      ...item,
+      id: Number(item.name),
+    }));
+    return [201, { items: responseItems }];
+  });
+  return counterObject;
+}
+
+function fillArray() {
+  return new Array(2003).fill(null).map((_, index) => ({
```
polomani: you don't need to fill it with nulls, as you map it right afterwards

src/__tests__/standardMethods.spec.ts
```diff
@@ -27,6 +30,26 @@ function numberGeneratorAPIEndpoint(config: AxiosRequestConfig) {
...
+    id: Number(index),
```
polomani: casting to number is redundant, index is already a number

src/__tests__/standardMethods.spec.ts
```diff
@@ -27,6 +30,26 @@ function numberGeneratorAPIEndpoint(config: AxiosRequestConfig) {
   }
 }
 
+function countChunks(axiosMock: MockAdapter, url: string) {
+  const counterObject = { value: 0 };
```
polomani, line 35: why do you wrap it in object? can't you just have a `number` variable?

polomani (changes requested):

polomani:
Yeah, I guess we just need to sit together for this one.

src/__tests__/standardMethods.spec.ts
ghost: Not if you want to return a dynamic number-variable from the function. Need a reference and not a plain value

ghost (approved):
LGTM 🚀

src/__tests__/standardMethods.spec.ts
```diff
@@ -133,4 +154,54 @@ describe('standard methods', () => {
       }
     });
   });
+  describe('generateDeleteEndpoint', () => {
```
polomani: I would suggest to move 3 of those up to the rest `generate...`  tests

src/standardMethods.ts
```diff
@@ -190,19 +187,26 @@ export function generateSimpleListEndpoint<RequestParams, ResponseType>(
...
-      },
+    const chunks = doChunking<IdType>(ids, chunkFunction);
+    const responses = await promiseAllWithData(
+      chunks,
+      input =>
+        rawRequest<ItemsResponse<ResponseType>>(axiosInstance, {
+          method: 'post',
+          url: `${resourcePath}/byids`,
+          data: { items: input },
+        }),
```
polomani:
```suggestion
        }, true),
```
is safe to retry = true

src/standardMethods.ts
```diff
@@ -190,19 +187,26 @@ export function generateSimpleListEndpoint<RequestParams, ResponseType>(
...
       true
```
polomani:
```suggestion
      false
```
runInSequence - false, we don't care about order of our requests here

src/standardMethods.ts
```diff
@@ -225,38 +229,52 @@ export function generateRetrieveSingleEndpoint<IdType, ResponseType>(
...
+      input =>
+        rawRequest<ItemsResponse<ResponseType>>(axiosInstance, {
+          method: 'post',
+          url: `${resourcePath}/delete`,
+          data: { items: input },
+        }),
       true
     );
-    return metadataMap.addAndReturn({}, response);
+    return metadataMap.addAndReturn({}, responses[0]);
```
polomani, line 251: I would rather merge with `Object.assign()` all of them to see the actual result.
Although it should not fail

src/standardMethods.ts
```diff
@@ -225,38 +229,52 @@ export function generateRetrieveSingleEndpoint<IdType, ResponseType>(
...
+    const chunks = doChunking<RequestType>(changes, chunkFunction);
+    const responses = await promiseAllWithData(
+      chunks,
+      input =>
+        rawRequest<Response>(axiosInstance, {
+          method: 'post',
+          url: `${resourcePath}/update`,
+          data: { items: input },
+        }),
+      true
```
polomani: do we need an order here? I think we can change this to `false`

src/standardMethods.ts
```diff
@@ -306,19 +324,22 @@ export function generateRetrieveLatestEndpoint<RequestParams, ResponseType>(
...
+      input =>
+        rawRequest<ItemsResponse<{}>>(axiosInstance, {
+          method: 'post',
+          url: resourcePath,
+          data: { items: input },
+        }),
       true
     );
-    return metadataMap.addAndReturn({}, response);
+    return metadataMap.addAndReturn({}, responses[0]);
```
polomani, line 346: The same here, I would merge responses to see an actual result

polomani (dismissed):

src/standardMethods.ts
f1cognite: What if we want to keep the order? If you request `[a,b,c]` and you end up with `[b,c,a]` then thats not good.

src/standardMethods.ts
ghost: What if we want to keep the order? If you request [a,b,c] and you end up with [b,c,a] then thats not good.

src/standardMethods.ts
ghost: What do you mean? `/delete` only returns an empty object...

src/standardMethods.ts
ghost: Hm. We prob. want to keep the order as the same as you requested?

src/standardMethods.ts
ghost: This also returns an empty response...

src/standardMethods.ts
polomani: the responses will be in the same order
this only says about the request order
if you put false it makes requests one by one, always waiting for the prev one

src/standardMethods.ts
polomani: answered below

src/standardMethods.ts
polomani: well, that's because we know it.. but, showing just one response doesn't feel right
what if one them was not empty?

src/standardMethods.ts
polomani: same, what if it did return something once? shouldn't we better know that something went the wrong way?

src/standardMethods.ts
polomani: Oh wait, yes, I am wrong

polomani: It does change the order 
assetUtils.ts:59: `responses.push(result);`

polomani: But that probably means that we need to have separate pr to fix that
Because sending requests sequentially just slows everything down here
One the other hand `promiseAllAtOnce` should keep the order, agree

src/__tests__/standardMethods.spec.ts
eirikdahlen: Yea I'll fix

src/standardMethods.ts
eirikdahlen: So what's the conclusion here?

src/standardMethods.ts
ghost: The return value here is `{}` and not `responses[0]`. `responses[0]` is used to get the headers/status code from the response (`{}`). But yes, we don't expose all headers from all the requests

src/standardMethods.ts
ghost: @polomani Are you actually sure it will mess up the order? `Promise.all` will be consistent?

src/standardMethods.ts
ghost: https://github.com/cognitedata/cognitesdk-js/pull/176#discussion_r298488697

src/standardMethods.ts
ghost: `Promise.all` preserves the order

src/standardMethods.ts
ghost: It also unnecessary makes the logic in our code more complex by doing that when it is clear from the API that is returns empty results.

src/standardMethods.ts
ghost: @eirikdahlen Ok, it should be `false`

src/standardMethods.ts
ghost: @eirikdahlen Yes, this should also be `false`

src/__tests__/standardMethods.spec.ts
ghost: `.fill` requires a value...

PR #181, opened 2019-06-26
[WIP] types generation using horiuchi/dtsgenerator

polomani:
To run the generation script: 
`sh scripts/generateTypes.sh`
That generates latest service contracts into `types/types.ts` interfaces

Known or possible issues:

1) [Object object] in generated examples https://github.com/horiuchi/dtsgenerator/issues/348
2) We renamed some interfaces, so they are different from the spec. So we either need to change the spec, or to fix the generator to support name mappings, or just manually make synonyms for them (fastest I guess). I did it manually for the ones that were breaking tests.
3) dtsgenerator makes a lot of nested namespaces. But you can choose "no namespace" option which generates file with no exports. So I workaround that by just find&replace (there is also a feature request about it on repo but it most likely won't be done soon)
4) Not able to check if interfaces are correct because our tests don't cover 100% of schemas (I think we need some automatic tester for that, which says if `schema==interface`).  Tests in dtsgenerator itself is not complete either, so we can't rely on them. 
5) We use our own branch of generator on `cognitedata/dtsgenerator#feat/typeMappings` which uses a few hacks like:
  - replace `number` with `Date` in specific places etc. (this one should be easy to refactor and make as a feature)
  - skip 1 particular broken schema. (this one is a totally silly hack)

We need to either upgrade them to features and suggest as a PR to dtsgenerator or use another way of doing it.
6) dtsgenerator produces some weird response interfaces which we don't use but they are still in the file (maybe filter them out?)
7) If we use this script we might want to move it somewhere else and have it in CI?
8) It does not generate an SDK completely as OpenAPITools/openapi-generator does. But then again openapi-generator doesn't work at all.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@types/query-string": "^6.1.0",
     "axios-debug-log": "^0.4.0",
     "axios-mock-adapter": "1.15.0",
-    "dtsgenerator": "^2.0.6",
+    "dtsgenerator": "cognitedata/dtsgenerator#feat/typeMappings",
     "es-check": "^5.0.0",
     "jest": "^23.6.0",
     "ls": "^0.2.1",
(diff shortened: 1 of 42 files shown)


polomani:
30.07.2019 We did an "SDK generation decision meeting" and decided to not do anything related to generation at least for next 6 months. After the end of this term we can review it once again

PR #182, opened 2019-06-27
fix: Improved stability and user-flow on popup logins

ghost:
- Use `postMessage` instead of hacky `postLoginTokens`
- Force the popup window to be a new window and not a tab
- Added new error-class for login problems

Validated in Chrome and Firefox


eirikdahlen (approved):
LGTM

PR #185, opened 2019-06-30
fix: Query param serialization for arrays

ghost:
Before this PR we would produce URLs like this: `events/list?assetIds[]=123&assetIds[]=456`
After this PR we produce it the Cognite way: `events/list?assetIds=[123,456]`


eirikdahlen (approved):
LGTM

Issue #186, opened 2019-07-02
Enhancement: Return timeseries data in v0.5


evenstensberg:
## Problem

As a user of the SDK it would be nice to use the data after creating it. Imagine a user wanting to create some timeseries and then add some sort of subscription based on that data. Currently, one needs to:

1) Create the timeseries
2) List all timeseries data and filter out those irrelevant

That means 2 calls to the SDK, which is not a great experience, for developers using the SDK.

# Proposed 

Returning the timeseries data from calling the sdk method `sdk.TimeSeries.create`.

https://github.com/cognitedata/cognitesdk-js/blob/master/src/TimeSeries.ts#L85-L86

evenstensberg:
https://github.com/cognitedata/cognitesdk-js/issues/8 is relevant, patch for this is probably okay

ghost:
Yes, this is a known inconsistency.

The background for this is that the timeseries api (0.5) didn't return created timeseries from the API (https://doc.cognitedata.com/api/0.5/#operation/postTimeSeries). All other APIs did this. We decided quite early that we wanted the SDK to be a 1:1 mapping between SDK-call & API-calls.

I am happy to announce that SDK (for v1) returns the created timeseries (https://cognitedata.github.io/cognitesdk-js/classes/timeseriesapi.html#create)

It will not prioritised to fix this for the SDK for 0.5 since we will deprecate 0.5 very soon. We encourage everyone to migrate to the SDK for v1.

evenstensberg:
Thanks for the update. Will update to v1 instead!

PR #188, opened 2019-07-02
enh: add return type for ts

evenstensberg:
Adds return type for timeseries as apart of #186 

## TODO

- API needs to be updated 

Edit: We can also do a get for these within the post, but that's probably a bad idea

### DIFF
--- a/src/TimeSeries.ts
+++ b/src/TimeSeries.ts
@@ -77,13 +77,14 @@ export interface TimeseriesSearchParams {
 const timeSeriesUrl = (): string => `${apiUrl(0.5)}/${projectUrl()}/timeseries`;
 
 export class TimeSeries {
-  public static async create(timeSeries: Partial<Timeseries>[]): Promise<void> {
+  public static async create(timeSeries: Partial<Timeseries>[]): Promise<Timeseries[]> {
     const body = {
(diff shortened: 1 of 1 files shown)


src/TimeSeries.ts
```diff
@@ -77,13 +77,14 @@ export interface TimeseriesSearchParams {
...
+  public static async create(timeSeries: Partial<Timeseries>[]): Promise<Timeseries[]> {
     const body = {
       items: timeSeries,
     };
     const url = timeSeriesUrl();
-    await rawPost(url, { data: body });
-    return;
+    const response = (await rawPost(url, { data: body })) as AxiosResponse<
+    TimeseriesResponse>;
+    return response.data.data.items;
```
ghost, line 87: This will throw? `response.data === {}` https://doc.cognitedata.com/api/0.5/#operation/postTimeSeries
See https://github.com/cognitedata/cognitesdk-js/issues/186#issuecomment-507727041

evenstensberg: You're right!

ghost:
@evenstensberg Should we close the PR?

evenstensberg:
Yes we can, sorry for late reply

PR #214, opened 2019-07-11
test: Increase sleep time for eventual consistency

ghost:
A lot of tests in Travis fails because of eventual consistency


src/__tests__/resources/timeseries.integration.spec.ts
```diff
@@ -74,7 +74,7 @@ describe('Timeseries integration test', () => {
 
   test('list from assetIds', async () => {
     // wait (eventual consistency delay)
-    await sleepPromise(2000);
+    await sleepPromise(5000);
```
polomani, line 77: didn't we have some kind of retry thing for this kind of cases?

polomani (approved):

PR #223, opened 2019-07-12
chore: auto update js code samples after package deploy

polomani:


scripts/afterRelease.sh
```diff
@@ -0,0 +1,24 @@
+brew install hub
+yarn typedoc --json docs.json
+node scripts/postProcessDocs.js
+
+packageVersion=$(node -e "const fs = require('fs');\
```
ghost: Replace this with `packageVersion=${jq .version ./package.json}`?

scripts/afterRelease.sh
```diff
@@ -0,0 +1,24 @@
...
+const packageFile = fs.readFileSync('./package.json');\
+const packageJson = JSON.parse(packageFile.toString());\
+console.log(packageJson.version)")
+branchName="feat/updateSDKJsExamples$packageVersion"
```
ghost, line 5: We can't get conflict here? So we only update when we release a new version?

scripts/afterRelease.sh
polomani: nice

scripts/afterRelease.sh
polomani: correct, we only update after we release. means, we won't have the same branch name

ghost (approved):
LGTM

PR #252, opened 2019-08-01
[DEVX-23] test: Replace most sleepPromise with retry-strategy

ghost:
Increase integration test stability by avoiding `sleepPromise` and rather retry the test until it succeeds. This should in theory reduce the test time.


polomani (approved):
NICE!

PR #267, opened 2019-08-22
fix: don't send expired tokens to authorize in scheduleRenewal call

ghost:
When `sdk.Login.authorize` is called with cached id&access-tokens it will use them and make sure to renew the tokens before they expires.
But today they are sent in as parameters to the second call to `sdk.Login.authorize` which is wrong (see diff).

This causes this network flow: https://docs.google.com/document/d/1RqiGrs25Udk-7spBuQ-wwS35BLdTUOFUd4ZX9HAiZg4/edit?usp=sharing

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cognite/sdk",
-  "version": "1.7.7",
+  "version": "1.7.8",
   "description": "Javascript client library for Cognite SDK",
   "contributors": [
     "Fredrik Anfinsen <fredrik.anfinsen@cognite.com>",
--- a/src/Login.ts
+++ b/src/Login.ts
@@ -387,7 +387,10 @@ export class Login {
     cancelSchedule = scheduleTask(
       () => {
         cancelSchedule();
-        Login.authorize(params, tokenCallback);
+        Login.authorize(
+          { ...params, idToken: undefined, accessToken: undefined },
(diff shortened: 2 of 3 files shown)


Meemaw (approved):
LGTM. This doesn't fix the "getting kicked out" problem though?

ghost:
Why not?

Meemaw:
From the timeline in doc it seems like we are just using the expired token for some time (5 seconds?) and this PR fixes that?

The bigger problem is that the user gets redirected from `/charts/<chartId>` to `/charts` after some time. Does this PR somehow fix that?

ghost:
The timeline in the doc showed us that the 5 seconds with 401 caused the griff-react component to end up in a bad state. With this PR the SDK will not throw 401s anymore caused my the renewal process.

Also is it true that the user was actually kicked out from `/charts<chartId>` to `/charts`? I haven't heard about that before

Meemaw:
Yes, I had a talk with one customer who was experiencing this. I had a felling that this was the bug that we are dealing with here. @AadneAadland was present in that call as well.

Meemaw:
Should we merge this?

ghost:
@Meemaw This package will auto-deploy to npm now so available in a few min. Would be great if you could bump the SDK version in Insight to check if this resolves the issue.

PR #281, opened 2019-09-09
[DEVX-7] fix: safely delete assets with chunking

ghost:


src/resources/assets/assetsApi.ts
```diff
@@ -109,8 +109,12 @@ export class AssetsAPI extends BaseResourceAPI<TypeAsset, Asset, AssetList> {
    * await client.assets.delete([{id: 123}, {externalId: 'abc'}]);
    * ```
    */
-  public async delete(ids: AssetIdEither[], params?: AssetDeleteParams) {
-    return super.deleteEndpoint(ids, params);
+  public async delete(ids: AssetIdEither[], params: AssetDeleteParams = {}) {
+    const paramsWithIgnoreUnknownIds = {
+      ...params,
+      ignoreUnknownIds: true,
```
polomani, line 115: it is not default on the API but it is on our side? why so?

ghost: If we are chunking automatically it has to be default from our side

ghost: The user of the SDK shouldn't care

polomani (approved):

PR #282, opened 2019-09-09
[DEVX-88] feat: expose logout.getUrl() endpoint

polomani:
https://cognitedata.atlassian.net/browse/DEVX-88

### DIFF
--- a/src/resources/logout/logoutApi.ts
+++ b/src/resources/logout/logoutApi.ts
@@ -1,0 +1,26 @@
+// Copyright 2019 Cognite AS
+
+import { MetadataMap } from '../../metadata';
+import { CDFHttpClient } from '../../utils/http/cdfHttpClient';
+import { BaseResourceAPI } from '../baseResourceApi';
+import { getLogoutUrl } from '../login';
+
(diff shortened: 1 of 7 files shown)


src/resources/logout/logoutApi.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * [Retrieve a logout url](https://docs.cognite.com/api/v1/#operation/logoutUrl)
+   *
+   * ```js
+   * // You can specify the url to send the user to after the logout is successful.
+   * // If no url is passed, you will end up at the IDPs log out page.
+   * const logoutUrl = await client.logout.getUrl('[url to redirect]');
+   * ```
+   */
+  public async getUrl(redirectUrl?: string) {
```
ghost: Wrap this into an object?

ghost (changes requested):

src/resources/logout/logoutApi.ts
polomani: @f1cognite  fixed

ghost (approved):
LGTM

PR #283, opened 2019-09-10
feat: use cached access token to skip auth flow

ghost:
Provide functionality for the SDK user to pass in a cached access token

### DIFF
--- a/src/cogniteClient.ts
+++ b/src/cogniteClient.ts
@@ -60,6 +60,10 @@ export const POPUP = 'POPUP';
 export interface OAuthLoginOptions extends Project {
   onAuthenticate?: OnAuthenticate | 'REDIRECT' | 'POPUP';
   onTokens?: OnTokens;
+  /**
+   * Provide optional cached access token to skip the authentication flow (client.authenticate will still override this).
+   */
+  accessToken?: string;
 }
 
 function validateAndReturnAPI<T>(api: T | undefined): T {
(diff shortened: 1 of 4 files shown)


valgussev:
So basically if `accessToken` is provided SDK under the hood will set `setBearerToken` without the need of calling `authenticate`?

ghost:
@valerii-cognite That is correct

valgussev (approved):

src/cogniteClient.ts
```diff
@@ -286,6 +291,11 @@ export default class CogniteClient {
       return didAuthenticate ? retry() : reject();
     });
 
+    const { accessToken } = options;
```
valgussev, line 294: please verify that `accessToken` contains same `project` as passed `project`

ghost: How can we verify this??

ghost: And why do we need this check? What scenario will that be?

ghost: The case is:
- You login to `akerbp` but provide access token for `publicdata`
- When you then call `client.login.status()` it will give you the status for the wrong project.
 That is confusing.

Any solutions @polomani ?

polomani: @f1cognite  how will all other endpoints behave? If they return 401 then we can just to fix the `login.status()` to return null

PR #284, opened 2019-09-11
[DEVX-110] fix: allow shallow copying for Asset and Timeseries

polomani:
https://cognitedata.atlassian.net/browse/DEVX-110

This is the best I can do here to help destructing them easier

So basically those tests show the way Asset and Timeseries behave while shallow copied.
I am still not sure I did the right thing here.

I mean it just makes our code messier, while for the user it's just a matter of calling `assets.map(asset=> asset.toJSON())`
or do something like `assets.map({id, name} => ({id, name}))`

So idk if we should be responsible for people destructing our classes in a poor way... 

Neither `Object.assign` or `spread` operator will ever be able to copy our custom `toJSON`, everything that goes after destruction is not under our control 

I don't like this code at all, btw


src/__tests__/resources/assetList.integration.specs.ts
```diff
@@ -41,10 +41,4 @@ describe('assetList', () => {
     await client.assets.delete(createdAssets.map(asset => ({ id: asset.id })));
     await client.events.delete(createdEvents.map(event => ({ id: event.id })));
   });
-
-  test('json stringify', async () => {
```
ghost, line 45: Why remove?

src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -198,4 +199,49 @@ describe('Asset unit test', () => {
...
+    const items = [
+      {
+        id: 1,
+      },
+      {
+        id: 2,
+      },
+    ];
+
+    test('assets.list().autoPagingToArray()', async () => {
```
ghost: Not meaningful test description

src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -198,4 +199,49 @@ describe('Asset unit test', () => {
...
+      nock(mockBaseUrl)
+        .post(new RegExp('/assets/list'), {})
+        .once()
+        .reply(200, { items });
+
+      const assets = await client.assets.list().autoPagingToArray({ limit: 2 });
+      expect(JSON.stringify(assets)).toBeTruthy();
```
ghost: `toBeTruthy` will I consider to be a weak test here. What do you test with that?

src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -198,4 +199,49 @@ describe('Asset unit test', () => {
...
+      const shallowCopyItems: TypeAsset[] = [...assets, assets[0]];
```
ghost: This is not spreading any objects? Just spreading an array which already works fine...

src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -198,4 +199,49 @@ describe('Asset unit test', () => {
...
+      expect(JSON.stringify(shallowCopyItems)).toBeTruthy();
+      const destructedAsset: TypeAsset = { ...assets[1] };
+      expect(destructedAsset.id).toBeDefined();
+      expect(() => JSON.stringify(destructedAsset)).toThrowError();
+      const destructedJsonAsset: TypeAsset = { ...assets[1].toJSON() };
+      expect(JSON.stringify(destructedJsonAsset)).toBeTruthy();
+    });
+
+    test('assets.list()', async () => {
```
ghost: I don't see the need for this test? Duplicate of the one above? If we are using autoPaginate or not shouldn't matter or no?

src/__tests__/resources/timeSeriesList.unit.spec.ts
```diff
@@ -99,4 +101,38 @@ describe('TimeSeriesList class unit test', async () => {
...
+    const items = [
+      {
+        id: 1,
+      },
+      {
+        id: 2,
+      },
+    ];
+
+    test('timeseries.list().autoPagingToArray()', async () => {
```
ghost: Bad test description. If this test fails what was the cause of the problem by looking at the logs?

src/resources/classes/timeSeries.ts
```diff
@@ -1,53 +1,44 @@
...
+  public description: string;
+  public securityCategories?: number[];
+  public createdTime: Date;
+  public lastUpdatedTime: Date;
+  public id: CogniteInternalId;
 
   constructor(client: CogniteClient, props: GetTimeSeriesMetadataDTO) {
-    super(client, props);
+    super(client);
+    this.externalId = props.externalId;
```
ghost, line 30: We could potentially add support in `BaseResource` for a list of prop names like `['externalId', 'name', 'isString', ...]`. And then use that list to create the `toJSON` return value and we could send the props to `BaseResource` as before.

ghost (changes requested):

src/resources/classes/timeSeries.ts
polomani: and the force casting to the type we need? 'cause we will lose the type information
also how will you able to remember to update that list of strings every time you update the interface?

src/__tests__/resources/assets.unit.spec.ts
polomani: the one above returns `Asset[]`, but this one returns `AssetList`
bug?

src/__tests__/resources/assets.unit.spec.ts
polomani: you are right, will remove it

src/__tests__/resources/assets.unit.spec.ts
polomani: basically just that it's not failing, maybe snapshot instead?

src/__tests__/resources/assetList.integration.specs.ts
polomani: moved to unit test intead

src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -198,4 +199,44 @@ describe('Asset unit test', () => {
...
+      {
+        id: 2,
+      },
+    ];
+
+    beforeEach(() => {
+      nock.cleanAll();
+      nock(mockBaseUrl)
+        .post(new RegExp('/assets/list'))
+        .thrice()
```
ghost, line 217: Why `thrice`? This runs before each and not before all

src/__tests__/resources/assets.unit.spec.ts
```diff
@@ -198,4 +199,44 @@ describe('Asset unit test', () => {
...
+      expect(() => JSON.stringify(assets[0])).not.toThrow();
+    });
+
+    test('change context for asset utility methods', async () => {
+      const assets = await client.assets.list().autoPagingToArray();
+      const utilMethod = assets[0].children;
+      const result = await utilMethod();
+      const resultAfterBind = await utilMethod.call(null);
+      expect({ ...result[0] }).toEqual({ ...resultAfterBind[0] });
+      expect([{ ...result[0] }, { ...result[1] }]).toEqual(items);
```
ghost, line 233: Excellent!

ghost (approved):

src/__tests__/resources/assets.unit.spec.ts
polomani: one of tests requires 3 calls, I didn't want to have a special case for that..

PR #286, opened 2019-09-17
Fix/api methods as arrow functions

polomani:


src/__tests__/cogniteClient.spec.ts
```diff
@@ -368,4 +368,72 @@ describe('CogniteClient', () => {
       });
     });
   });
+
+  describe.only('api endpoints smoke test', () => {
```
ghost:
```suggestion
  describe('api endpoints smoke test', () => {
```

src/resources/3d/assetMappings3DApi.ts
```diff
@@ -19,18 +18,18 @@ export class AssetMappings3DAPI extends BaseResourceAPI<AssetMapping3D> {
    * const mappings3D = await client.assetMappings3D.list(3244265346345, 32423454353545);
    * ```
    */
-  public list(
+  public list = (
     modelId: CogniteInternalId,
     revisionId: CogniteInternalId,
     scope?: AssetMappings3DListFilter
-  ): CursorAndAsyncIterator<AssetMapping3D> {
```
ghost, line 26: Do you want to remove all the explicit return types?

polomani: they don't make sense in most places, because the functions we call are already typed
(at least I didn't notice any difference)
But anyway, what was the idea on having them on you side?

src/__tests__/cogniteClient.spec.ts
polomani: oh, yeah

src/resources/3d/assetMappings3DApi.ts
ghost: Just to have explicit return values to give build errors when that isn't true anymore. Unsure how useful it is. So we can keep it as is for now (we have tests).

ghost (approved):

src/resources/3d/assetMappings3DApi.ts
polomani: agree here, will revert it back

PR #288, opened 2019-09-24
fix(OAuth): add warning when using OAuth without SSL

ghost:


polomani (approved):
this will trigger a release tho,
but I think it's ok, we should release patches fast

PR #289, opened 2019-09-25
Feat/sequences api

polomani:


src/types/types.ts
```diff
@@ -167,8 +170,101 @@ export interface Asset extends ExternalAsset, AssetInternalId {
...
+
+/**
+ * Describes a new column
+ */
+export interface ExternalSequenceColumn
+  extends ExternalSequenceColumnBase,
+    ExternalId {}
+
+interface ExternalSequenceColumnBase {
+  name?: ColumnName;
```
ghost: Maybe `SequenceColumnName` and `SequenceColumnDescription`?

src/types/types.ts
```diff
@@ -510,10 +635,23 @@ export type EXECUTE = 'EXECUTE';
 
 export type EventChange = EventChangeById | EventChangeByExternalId;
 
+export type SequenceChange = SequencePatch & IdEither;
```
ghost, line 638: Why not interface that extends? More consistent with other similar types

src/types/types.ts
```diff
@@ -724,6 +862,36 @@ export interface Filter {
...
+   * Columns to be included. Specified as list of column externalIds.
+   * In case this filter is not set, all available columns will be returned.
+   */
+  columns?: string[];
+}
+
+/**
+ * A request for datapoints stored
+ */
+export type SequenceRowsRetrieve = SequenceRowsRetriveData & IdEither;
```
ghost, line 886: Same here. Interface that extends

src/resources/baseResourceApi.ts
```diff
@@ -302,3 +313,11 @@ export abstract class BaseResourceAPI<
 type ListEndpoint<QueryType extends FilterQuery, WrapperType> = (
   query?: QueryType
 ) => Promise<HttpResponse<CursorResponse<WrapperType>>>;
+
+interface PostInParallelWithAutomaticChunkingParams<RequestType, ParamsType> {
```
ghost, line 317: 😆 ILikeVeryLongTypeNamesThatAreUnderstandableAndPrecise

ghost (changes requested):
Very good 🚀

src/types/types.ts
polomani: IdEither is a union type, when you do `SequenceRowsRetriveData & IdEither` it actually creates type, `(SequenceRowsRetriveData & InternalId) | (SequenceRowsRetriveData & ExternalId)`
How will you do it with extends? You cannot extend type

src/types/types.ts
polomani: answered below

src/resources/sequences/sequencesApi.ts
```diff
@@ -0,0 +1,170 @@
...
+  };
+
+  /**
+   * [Delete sequences](https://docs.cognite.com/api/v1/#operation/deleteSequences)
+   *
+   * ```js
+   * await client.sequences.delete([{id: 123}, {externalId: 'abc'}]);
+   * ```
+   */
+  public delete = (ids: IdEither[]) => {
```
ghost: Missing explicit return type

ghost (changes requested):

ghost (approved):
LGTM

PR #297, opened 2019-10-14
Fix/hide extra enumerable props

polomani:


polomani:
not too relevant anymore

PR #300, opened 2019-10-22
fix: serialisation for non-api errors

polomani:
from sentry https://sentry.io/organizations/cognite/issues/1240621332/?environment=production&project=1206818&query=is%3Aunresolved


valgussev (approved):
Seems about right but not sure whether it covers all the possible cases

polomani:
that's more like a guess, I tried to reproduce the initial error but didn't succeed with it.
so I believe this at least is a good start

PR #305, opened 2019-10-29
Fix SDK documentation link in README

ghost:
Link to SDK documentation is broken, so updating to (what I assume is) the correct link.


BugGambit:
This was fixed in #306

PR #310, opened 2019-11-01
feat: add asset list partition parameter

ghost:
This is a feature needed by Discovery. Currently, the JS sdk provides no easy way to fetch all assets of a tenant. However, for running Entity Matcher in discovery we would like to fetch everything quickly in parallel through partitions.

This solution simply enables the user to declare partitions. There's a more complex solution that can be useful that actually abstract out the fetching process itself.
```js
const results = await Promise.all(
      [...Array(n).keys()].map(async index => {
        let response = await sdk.assets.list({
          filter: { rootIds: [{ id: selectedRootAssetId }] },
          limit: 1000,
          // @ts-ignore
          partition: `${index + 1}/n`,
        });
        while (response.nextCursor) {
          response = await sdk.assets.list({
            filter: { rootIds: [{ id: selectedRootAssetId }] },
            limit: 1000,
            cursor: response.nextCursor,
            // @ts-ignore
            partition: `${index + 1}/n`,
          });
        }
...
)
```

### DIFF
--- a/src/__tests__/resources/assets.integration.spec.ts
+++ b/src/__tests__/resources/assets.integration.spec.ts
@@ -265,4 +265,10 @@ describe('Asset integration test', () => {
       .autoPagingToArray({ limit });
     expect(result.length).toBe(limit);
   });
+
+  test('list partition', async () => {
+    const limit = 5;
+    const result = await client.assets.list({ partition: '1/10', limit });
+    expect(result.items.length).toBe(limit);
+  });
 });
--- a/src/__tests__/resources/timeseries.integration.spec.ts
+++ b/src/__tests__/resources/timeseries.integration.spec.ts
@@ -106,6 +106,12 @@ describe('Timeseries integration test', () => {
       .autoPagingToArray({ limit: 100 });
   });
 
+  test('list partition', async () => {
+    await client.timeseries
+      .list({ includeMetadata: false, partition: '1/10' })
+      .autoPagingToArray({ limit: 100 });
+  });
+
   test('search', async () => {
     const name = 'test__constant_0_with_noise';
     const result = await client.timeseries.search({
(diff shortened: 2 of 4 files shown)


polomani:
Was it released? I don't see it in documentation

src/__tests__/resources/assets.integration.spec.ts
```diff
@@ -256,4 +256,10 @@ describe('Asset integration test', () => {
       .autoPagingToArray({ limit });
     expect(result.length).toBe(limit);
   });
+
+  test('list partition', async () => {
+    const limit = 5;
+    const result = await client.assets.list({ partition: '1/10', limit });
```
polomani, line 271: does it work well with autoPaging methods?

ghost: I dont think so. Partition is only used when you dont want pagination (fetch everything ignoring limits). So i dont think we should think about that case?

ghost:
@polomani @f1cognite @maciejmrowiec As discussed in the chat, this would enable `partition` for assets, events and timeseries.

src/__tests__/resources/timeseries.integration.spec.ts
```diff
@@ -106,6 +106,12 @@ describe('Timeseries integration test', () => {
       .autoPagingToArray({ limit: 100 });
   });
 
+  test('list partition', async () => {
+    await client.timeseries
+      .list({ includeMetadata: false, partition: '1/10' })
+      .autoPagingToArray({ limit: 100 });
```
polomani, line 112: you don't test a response here?

polomani (commented):
Tests are failing, did you try locally?

src/__tests__/resources/assets.integration.spec.ts
BugGambit: Partition should work excellent with pagination?
Essentially you just divide the data set in the backend before fetching so still need pagination for each partition

polomani:
@dlcognite `Unexpected field - filter.partition | code: 400` leave that on me, I can implement support for timeseries myself, I know what this issue. But the rest you can do

polomani (approved):

PR #314, opened 2019-11-11
Expose HttpError from main module

sebbert:


sebbert:
For the tests running on node 8, I get:
```
$ /home/travis/build/cognitedata/cognite-sdk-js/samples/nodejs/node_modules/.bin/check-node-version --node '>= 10'

node: 8.16.2

Error: Wanted node version >= 10 (>=10.0.0)

To install node, run `nvm install >= 10` or see https://nodejs.org/

error Command failed with exit code 1.

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

Done in 9.73s.

The command "yarn test-samples" exited with 0.
```

polomani:
@sebbert why do you need this HttpError? Don't we always wrap it in `CogniteError`?

sebbert:
@polomani Well, not in the case where the error response does not match what we expect to get back from CDF (see [handleErrorResponse](https://github.com/cognitedata/cognite-sdk-js/blob/v1/src/error.ts#L72)).

We're using `CogniteClient#get` etc. with `{ withCredentials: true }` to call the Jetfire backend using the global CogniteClient in the console, since it will automatically provide credentials in requests, and also refresh the bearer token when needed. Our error responses do not match that of CDF, so it will never end up wrapping in CogniteError.

I really don't like this solution, but it seems to be the only reasonable way to get this working for now. Perhaps we could add a flag to `HttpRequestOptions` which would force it to never wrap it in a `CogniteError`?

polomani (approved):
ah, in that case yes, it's an easiest solution. 
let's do this for now then
HttpError is ok to expose

Issue #316, opened 2019-11-15
The type AclScopeTimeseries is divergent with the API docs


nildo:
The type `AclScopeTimeseries` is currently defined as `AclScopeAll | AclScopeAssetsId`.

Today you can't use `AclScopeAssetsId` in groups creation anymore.

It is also missing the option of a `TimeSeriesIdScope`, according to the API v1 docs (https://docs.cognite.com/api/v1/#operation/createGroups).

Then, when using TypeScript, there is a typing error when assigning a group scoped by time series ids to a variable of type Group.

polomani:
hmm, it seems like documentation is not in sync with protobufs again
https://github.com/cognitedata/api-acls/blob/a85c3f37f4b9893ddb2dde6a59416a9b37e9ffe0/protos/timeseries_acl.proto

@ngerstle-cognite  could you confirm that `assetIdScope.subtreeIds` still exists?

and now also `assetIds` has been added?

ngerstle-cognite:
Subtree IDs exist, but can't be created anymore.

nildo:
Since the subtree IDs still exist, and just can't be created anymore, I think the PR #317 fixes the whole problem. timeseries ids exists and can be created, so, they just need to be added there.

PR #317, opened 2019-11-18
fix: add type for scope with time series ids

nildo:
This partially solves the issue #316 I opened.

A missing scope type `AclScopeTimeSeriesIds` was created. But the scope type `AclScopeAssetsId` that, according to the API docs, is not possible to use anymore is still there. I didn't remove it because I do not have the knowledge to be sure about the implications.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -67,6 +67,12 @@ export interface AclScopeAssetsId {
   };
 }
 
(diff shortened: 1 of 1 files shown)


polomani:
thank you!

polomani (approved):

PR #323, opened 2019-12-02
feat: add HTTP patch method to httpClient

Meemaw:
Support HTTP patch method for completeness.

We are using this in some of the Applications APIs and we are using SDK to get authentication for free.


polomani (approved):

polomani:
thank you!

PR #324, opened 2019-12-03
feat: add isString to DatapointsGetAggregateDatapoint

evancharlton:
Aggregates can never be string values, so define `isString: false` on
the DatapointsGetAggregateDatapoint type. This makes it easier to work
with the DatapointsGetDatapoint union type in applications because
they will have a consistent shape.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -533,6 +533,8 @@ export type DatapointsDeleteRequest =
 
 export interface DatapointsGetAggregateDatapoint extends DatapointsMetadata {
   datapoints: GetAggregateDatapoint[];
+  /** Aggregates cannot be strings, as this is unsupported by CDF. */
+  isString: false;
 }
 
 export type DatapointsGetDatapoint =


BugGambit (approved):

evancharlton:
Could this be considered a breaking change? It's hard to imagine, but maybe someone has some conditionals established based on the existence of this field?

(cc: @Meemaw )

evancharlton:
Actually, it doesn't even build so it must be a breaking change 😉 

I'll dig into it.

polomani:
@cognitedata/time-lords could we fix it in docs as well then?

evancharlton:
This causes a compile error in the `client.datapoints.retrieve` method because someone could -- in theory -- ask for an aggregate on a string time series (which is an API error). I don't know the best way to fix this, so I'm just going to cast in (again) in my app.

Please reopen this if you have some brilliant ideas on how to navigate the SDK type hierarchy! 😀

PR #326, opened 2019-12-09
feat: support assetSubtreeIds and assetExternalIds filters [DEVX-199]

polomani:
For:
- Events
- TimeSeries
- Assets (only assetSubtreeIds)

https://cognitedata.atlassian.net/browse/DEVX-199

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -802,6 +802,8 @@ export interface ExternalFilesMetadata {
   mimeType?: FileMimeType;
   metadata?: Metadata;
   assetIds?: CogniteInternalId[];
+  sourceCreatedTime?: Date;
+  sourceModifiedTime?: Date;
 }
 
 export interface ExternalId {
@@ -812,8 +814,11 @@ export interface FileChange {
   update: {
     externalId?: SinglePatchString;
     source?: SinglePatchString;
+    mimeType?: SinglePatchString;
     metadata?: ObjectPatch;
     assetIds?: ArrayPatchLong;
+    sourceCreatedTime?: SinglePatchDate;
+    sourceModifiedTime?: SinglePatchDate;
   };
 }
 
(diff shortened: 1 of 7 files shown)


src/types/types.ts
```diff
@@ -295,8 +295,20 @@ export type AssetExternalId = ExternalId;
 export interface AssetFilter extends Limit {
   filter?: {
     name?: AssetName;
+    /**
+     * Return only the direct descendants of the specified assets.
+     */
     parentIds?: CogniteInternalId[];
+    /**
+     * Return only the direct descendants of the specified assets.
```
Meemaw, line 303:
```suggestion
     * Returns only the direct descendants of the specified assets.
```

src/types/types.ts
```diff
@@ -295,8 +295,20 @@ export type AssetExternalId = ExternalId;
 export interface AssetFilter extends Limit {
   filter?: {
     name?: AssetName;
+    /**
+     * Return only the direct descendants of the specified assets.
```
Meemaw, line 299:
```suggestion
     * Returns only the direct descendants of the specified assets.
```

src/types/types.ts
```diff
@@ -824,10 +853,17 @@ export interface FileFilter extends Limit {
      * Only include files that reference these specific asset IDs.
      */
     assetIds?: CogniteInternalId[];
+    /**
+     * Only include files that have a related asset in a tree rooted at any of these root assetIds.
```
Meemaw, line 860: This comment is not consistent across types? In one place I see:

"The IDs of the root assets that the related assets should be children of." -- if the behaviour is the same they should probably have consistent comments.

Meemaw (approved):
There aren't many code changes so I don't have any real comments on that, but I added a few nitpicks about type comments -- try to be consistent with wording across resources.

src/utils.ts
```diff
@@ -109,6 +109,8 @@ export function transformDateInResponse<T>(data: T): T {
     'uploadedTime',
     'deletedTime',
     'timestamp',
+    'sourceCreatedTime',
+    'sourceModifiedTime',
```
valgussev, line 113: How do these two related to the topic?

valgussev (approved):

src/utils.ts
polomani: they are not really related, but they will go in a separate commit `feat(files): source created/modified time, rootAssetIds, assetSubtreeIds`, hence a separate item in the changelog as well.

Just wanted to save some time

valgussev: I see, than ready to merge

src/types/types.ts
polomani: agree

PR #336, opened 2020-01-16
chore(travis): fix semantic-release config to work with their v16

polomani:
semantic-release v16 works only with node 10. And since we spawn v8,9,10 instances on travis we can deploy only using one of them (and we should, the rest would fail anyway)

Also they changed the config syntax a bit, so I updated that


hhenriques (approved):
lgtm

PR #341, opened 2020-01-29
fix for datapoint delete type

sanderland:


larshagencognite (approved):

polomani (changes requested):
you have to write a commit message according to the rules
and I would mark it as a refactor, since this doesn't require a new build.

sanderland:
> you have to write a commit message according to the rules
> and I would mark it as a refactor, since this doesn't require a new build.

PR #342, opened 2020-01-30
fix: correct type on deleteDatapointsEndpoint method

sanderland:
Is this better @polomani ? your guide tells me to run yarn run commit, which doesn't exist


polomani (approved):
thanks, will fix

PR #344, opened 2020-02-09
feat: add getDefaultRequestHeaders() to get headers required by the API

larsmoa:
Add functionality to retrieve headers used for HTTP requests. This is useful to use CDF with libraries that doesn't alllow fetching data through the API and therefor require raw headers, e.g. Cesium.

### DIFF
--- a/src/cogniteClient.ts
+++ b/src/cogniteClient.ts
@@ -34,7 +34,11 @@ import { SequencesAPI } from './resources/sequences/sequencesApi';
 import { ServiceAccountsAPI } from './resources/serviceAccounts/serviceAccountsApi';
 import { TimeSeriesAPI } from './resources/timeSeries/timeSeriesApi';
 import { apiUrl, getBaseUrl, isUsingSSL, projectUrl } from './utils';
-import { HttpRequestOptions, HttpResponse } from './utils/http/basicHttpClient';
+import {
+  HttpHeaders,
+  HttpRequestOptions,
+  HttpResponse,
+} from './utils/http/basicHttpClient';
 import { CDFHttpClient } from './utils/http/cdfHttpClient';
 
 export interface ClientOptions {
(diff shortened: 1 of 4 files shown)


src/utils/http/basicHttpClient.ts
```diff
@@ -79,6 +79,10 @@ export class BasicHttpClient {
     return this;
   }
 
+  public getDefaultHeaders(): HttpHeaders {
+    return this.defaultHeaders;
```
polomani: need to clone here as well?
```suggestion
    return { ...this.defaultHeaders };
```

src/cogniteClient.ts
```diff
@@ -324,6 +328,29 @@ export default class CogniteClient {
...
+   * outside the SDK.
+   *
+   * ```js
+   * const customUrl = [...];
+   * const headers = client.getDefaultRequestHeaders();
+   * const result = await fetch(customUrl, { headers: new Headers(headers) });
+   * ```
+   */
+  public getDefaultRequestHeaders(): HttpHeaders {
+    return Object.assign({}, this.httpClient.getDefaultHeaders());
```
polomani: more compact way
```suggestion
    return { ...this.httpClient.getDefaultHeaders() };
```

src/cogniteClient.ts
```diff
@@ -324,6 +328,29 @@ export default class CogniteClient {
...
+  }
+
+  /**
+   * Returns the default HTTP request headers, including e.g. authentication
+   * headers that is included in all requests. Headers provided per-requests is not
+   * included in this list. This function is useful when constructing API requests
+   * outside the SDK.
+   *
+   * ```js
+   * const customUrl = [...];
```
polomani:
```suggestion
   * const customUrl = '...';
```
it has to be a string

src/cogniteClient.ts
```diff
@@ -324,6 +328,29 @@ export default class CogniteClient {
...
+   * const headers = client.getDefaultRequestHeaders();
+   * const result = await fetch(customUrl, { headers: new Headers(headers) });
```
polomani:
```suggestion
   * const result = await fetch(customUrl, { headers });
```
`new Headers(...)` can it work without a constructor? can you check plz? if so then we could simplify the example

src/cogniteClient.ts
```diff
@@ -324,6 +328,29 @@ export default class CogniteClient {
     this.httpClient.setBaseUrl(baseUrl);
   };
 
+  /**
+   * Returns the base-url used for all requests.
+   */
+  public getBaseUrl(): string {
+    return this.httpClient.getBaseUrl();
```
polomani, line 335: could you add a test for this one somewhere as well?

polomani (commented):
Looks good!
minor things to address

larsmoa:
Thx for the input - fixed.

polomani (approved):
thanks, I'll merge it myself

PR #348, opened 2020-02-17
WIP: Feat/timeseries datapoint pagination

hernil:
This feature should allow for requesting more than 10000 datapoints through the sdk. 
See the commit logs for details as they should be somewhat detailed. 

This is a WIP because it currently only supports one item per query. This will be fixed, but the PR is opened anyway as I would like input on a few different points: 

- This feature highjacks and manipulates response data. The rationale is that this makes the feature invisible to the user, but a downside can be that it can confuse a user looking at the network activity tab of their browser. 
- I'd like to make test cases for this code, but I'm not sure how well the test framework would handle throwing more than 10000 datapoints in the loop. 
- What is our stance on throwing exceptions? 
  - Spesifically should we allow queries that are seemingly allowed by the API but that logically does not hold up? Like a query where `{ start, end, limit, granularity }` are all provided but with conflicting valyues. 
- Sorting done in 64281a5 can add some overhead with larger sets of datapoints (about five seconds for ~1 million datapoints in my very basic testing). Should we be concerned or is this fine? 

So in summary, there's still quite fundamental functionality missing but I'd appriciacte any input on the progress so far! 😄

### DIFF
--- a/src/resources/dataPoints/multiQueryFetchHelpers.ts
+++ b/src/resources/dataPoints/multiQueryFetchHelpers.ts
@@ -1,0 +1,31 @@
+// Copyright 2019 Cognite AS
+
+import ms from 'ms';
+import { DatapointsMultiQuery } from '../..';
+import { PotentiallyUndefinedQueryValues } from './types';
+
+function cleanQueryInput(
+  q: DatapointsMultiQuery
+): PotentiallyUndefinedQueryValues {
+  function alwaysReturnNumber(value: string | number | Date): number {
+    if (typeof value === 'string') {
+      return ms(value);
+    }
+    if (value instanceof Date) {
+      return value.getMilliseconds();
+    }
+    return value;
+  }
+  function numberOrUndefined(
+    // tslint:disable-next-line:max-union-size
+    value: number | string | Date | undefined
+  ): number | undefined {
+    return typeof value === 'undefined' ? value : alwaysReturnNumber(value);
+  }
+  const start = numberOrUndefined(q.start);
+  const end = numberOrUndefined(q.end);
+  const limit = numberOrUndefined(q.limit);
+  const granularity: undefined | number = numberOrUndefined(q.granularity);
+
+  return { start, end, limit, granularity };
+}
--- a/src/resources/dataPoints/multiQueryFetchHelpers.ts
+++ b/src/resources/dataPoints/multiQueryFetchHelpers.ts
@@ -1,7 +1,14 @@
 // Copyright 2019 Cognite AS
 
 import ms from 'ms';
-import { DatapointsMultiQuery } from '../..';
+import {
+  DatapointsGetAggregateDatapoint,
+  DatapointsGetDatapoint,
+  DatapointsMultiQuery,
+  ItemsWrapper,
+} from '../..';
+import { API_MAX_DATAPOINTS_PER_REQUEST } from '../../constants';
+import { CDFHttpClient } from '../../utils/http/cdfHttpClient';
 import {
   ConsolidatedQueryValues,
   PotentiallyUndefinedQueryValues,
@@ -78,4 +85,34 @@ export function extrapolateValues(
     throw new Error('Invalid query composition');
   }
   return { start, end, limit, granularity } as ConsolidatedQueryValues;
+}
+
+export function createRequests(
+  originalQuery: DatapointsMultiQuery,
+  httpClient: CDFHttpClient,
+  path: string
+) {
+  const { start, end, granularity } = extrapolateValues(originalQuery);
+
+  const requests = [];
+  originalQuery.limit = API_MAX_DATAPOINTS_PER_REQUEST;
+  originalQuery.granularity = ms(granularity);
+  let newStart = start;
+  let newEnd = 0;
+  while (newEnd < end) {
+    const { ...newQuery } = originalQuery;
+    const potentialNewEnd =
+      newStart + API_MAX_DATAPOINTS_PER_REQUEST * granularity;
+    newEnd = potentialNewEnd < end ? potentialNewEnd : end;
+    newQuery.start = newStart;
+    newQuery.end = newEnd;
+    const request = httpClient.post<
+      ItemsWrapper<(DatapointsGetAggregateDatapoint | DatapointsGetDatapoint)[]>
+    >(path, {
+      data: newQuery,
+    });
+    requests.push(request);
+    newStart = newEnd;
+  }
+  return requests;
 }
--- a/src/resources/dataPoints/multiQueryFetchHelpers.ts
+++ b/src/resources/dataPoints/multiQueryFetchHelpers.ts
@@ -2,7 +2,10 @@
 
 import ms from 'ms';
 import { DatapointsMultiQuery } from '../..';
-import { PotentiallyUndefinedQueryValues } from './types';
+import {
+  ConsolidatedQueryValues,
+  PotentiallyUndefinedQueryValues,
+} from './types';
 
 function cleanQueryInput(
   q: DatapointsMultiQuery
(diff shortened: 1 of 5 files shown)


polomani:
> Specifically should we allow queries that are seemingly allowed by the API but that logically does not hold up? Like a query where `{ start, end, limit, granularity }` are all provided but with conflicting values.

- we didn't do any advanced checks in client so far, do such conflicts break the logic?

>  // It's polite to tell the user that their request is manipulated

- maybe, but not by injecting it in headers client-side. also we don't do it other similar places

>` if (limit > API_MAX_DATAPOINTS_PER_REQUEST) {`

- This check shouldn't be needed anywhere. Normally `createRequests(...)` thing should be able to handle any limit

- probably not a biggest problem here, but having any kind of helper files generally a bad idea (multiQueryFetchHelpers.ts). it would probably need to be somewhere in datapoints api, reusing its context

- I don't understand cleanQueryInput part. there is a request to get datapoints `this.httpClient.post(...)`, there is also other places where we chunk and merge afterwords. look at `promiseAllWithData` for instance, try to reuse stuff

>This feature highjacks and manipulates response data. The rationale is that this makes the feature invisible to the user, but a downside can be that it can confuse a user looking at the network activity tab of their browser

- highjack is fine, but do not modify inputs, you should rather destruct all input filters and create a multiple query out of it. something like:
```
const { granularity, limit, aggregate} = query;
const queries => map(...) => { granularity, limit: newLimitHere, ...}
promiseAllWithData(queries)
``` 
doing `originalQuery.granularity =...` doesn't look good

> I'd like to make test cases for this code, but I'm not sure how well the test framework would handle throwing more than 10000 datapoints in the loop

those should be unit tests, don't try to cover it in integration. please, have a look into similar test, like chunking or pagination

Overall, if you need this only for export data as csv (which is almost one use-case I believe) I would rather suggest implementing it on gearbox (or having some kind of helper outside of sdk). The reason for that is that it's quite hard to contribute sdk (requires deep knowledge of sdk)

hernil:
> this function looks waay complex.
> 
> why does it sort even? all datapoints are sorted on backend anyway
> 
Considering the async nature of the requests I figured there's no guarantee that the responses end up in the correct order. 
> why you need to go through all props if you only need `datapoints`? and even for datapoints it's just `[...datapointsA, ...datapointsB]` or I am missing something?
> 
Honestly, I just wasn't confident enough about what can end up in a response to simply dismiss everything other than the datapoints. 
> 
> and actually there is a function for that already, `mergeItemsFromItemsResponse`

Ah, that's great feedback. I'll have a look! 

> 
> 
> > Specifically should we allow queries that are seemingly allowed by the API but that logically does not hold up? Like a query where `{ start, end, limit, granularity }` are all provided but with conflicting values.
> 
> we didn't do any advanced checks in client so far, do such conflicts break the logic?

Putting aside the very real possibility that I've misunderstood something, consider the following: 
You query a for datapoints spanning a year with a granularity of 5 minutes. That should render 105120 datapoints. If you additionally pass on a limit of 20000 or even 120000 what are you expecting in return?  

> 
> > // It's polite to tell the user that their request is manipulated
> 
> maybe, but not by injecting it in headers client-side. also we don't do it other similar places
> 
Fair enough. Should I do it any other way or just remove it? 
> 
> > ` if (limit > API_MAX_DATAPOINTS_PER_REQUEST) {`
> 
> This check shouldn't be needed anywhere. Normally `createRequests(...)` thing should be able to handle any limit

That's fair. I just wanted to do as little as possible that could potentially break todays functionallity. Ie. just handle "normal requests" with a limit < 10000 as they always have been. As of now requests with a limit > 10000 will just fail. 

> 
> probably not a biggest problem here, but having any kind of helper files generally a bad idea (multiQueryFetchHelpers.ts). it would probably need to be somewhere in datapoints api, reusing its context

It was a matter of moving what ended up being a fairly large chunk of code related to one special feature out of the way and into its own scope to not overly complicate things in the existing file. 
> 
> I don't understand cleanQueryInput part. there is a request to get datapoints `this.httpClient.post(...)`, there is also other places where we chunk and merge afterwords. look at `promiseAllWithData` for instance, try to reuse stuff

Query-values can be all over the place as the sdk is very flexible. So `query.start` can be of type `string, number, Date or undefined`. I needed them to be numbers to do math needed for 
`extrapolateLimit` as a way to decide how to split up the number of requests. 
> 
> > This feature highjacks and manipulates response data. The rationale is that this makes the feature invisible to the user, but a downside can be that it can confuse a user looking at the network activity tab of their browser
> 
> highjack is fine, but do not modify inputs, you should rather destruct all input filters and create a multiple query out of it. something like:
> 
> 
> ```
> const { granularity, limit, aggregate} = query;
> const queries => map(...) => { granularity, limit: newLimitHere, ...}
> promiseAllWithData(queries)
> ```
> 
> doing `originalQuery.granularity =...` doesn't look good
> 
This seems more or less to be what I was going for. However there was some additional logic needed to calculate what the `newLimit` was going to be. I'm very much open for input on how to simplify the code! 😄 
> > I'd like to make test cases for this code, but I'm not sure how well the test framework would handle throwing more than 10000 datapoints in the loop
> 
> those should be unit tests, don't try to cover it in integration. please, have a look into similar test, like chunking or pagination
>
Will do! 
> 
> Overall, if you need this only for export data as csv (which is almost one use-case I believe) I would rather suggest implementing it on gearbox (or having some kind of helper outside of sdk). The reason for that is that it's quite hard to contribute sdk (requires deep knowledge of sdk)

I very much admit not having deep knowledge of the sdk so I'm leaning a bit on you guys! 😄  However this feature would bring the js sdk closer to being on par with the Python one in this area which is not an unwelcomed change if I have understood the signals conveyed by Murad correctly.

polomani:
> Considering the async nature of the requests I figured there's no guarantee that the responses end up in the correct order.

yeah, but you can keep request indexes and sort only responses, not data points

> Honestly, I just wasn't confident enough about what can end up in a response to simply dismiss everything other than the datapoints.

we'll leave this task for typescript

> That should render 105120 datapoints. If you additionally pass on a limit of 20000 or even 120000 what are you expecting in return?

limit takes precedence, so if there is more data points then limit says you cut off extra. if there is less then, you return what you have 
 
> Fair enough. Should I do it any other way or just remove it?

just remove, please

> It was a matter of moving what ended up being a fairly large chunk of code related to one special feature out of the way and into its own scope to not overly complicate things in the existing file.

then you need to abstract it :)
like there should be at least better name then just helper..
try to search for `new RevertableArraySorter(sortAssetCreateItems);` in the project

src/resources/dataPoints/multiQueryFetchHelpers.ts
```diff
@@ -0,0 +1,142 @@
...
+  granularity: number
+): number {
+  return Math.ceil((end - start) / granularity);
+}
+
+function extrapolateGranularity(start: number, end: number, limit: number) {
+  return Math.floor((start - end) / limit);
+}
+
+export function extrapolateValues(
```
sanderland, line 73: these do not make sense, the user should give all values except possible 'no limit', and if there are more than limit datapoints before end, it is cut off at limit.
also note granularity is a string and not a number

src/resources/dataPoints/multiQueryFetchHelpers.ts
```diff
@@ -0,0 +1,142 @@
...
+  const requests = [];
+  originalQuery.limit = API_MAX_DATAPOINTS_PER_REQUEST;
+  originalQuery.granularity = ms(granularity);
+  let newStart = start;
+  let newEnd = 0;
+  while (newEnd < end) {
+    const { ...newQuery } = originalQuery;
+    const potentialNewEnd =
+      newStart + API_MAX_DATAPOINTS_PER_REQUEST * granularity;
+    newEnd = potentialNewEnd < end ? potentialNewEnd : end;
```
sanderland, line 108: likewise, don't manipulate end.

sanderland:
there are a lot of misunderstandings here on how datapoints and aggregates work. 
please review
https://cognitedata.atlassian.net/wiki/spaces/TTSD/pages/1113227496/Paginating+datapoints

hernil:
If I've understood correctly there's a wish to have this functionality in Gearbox rather than here in the SDK so I'll make the necessary adjustments and open a new PR there :smile:

polomani:
@hernil yep, please have a look at algorithm provided by Sander

hhenriques:
Hi @hernil, we eventually want to have a set of helper functions for JS use. However, we aren't sure how we want that to look yet since we want to consider generating the sdks. So for now, all helper functions should be within gearbox. I'm closing the PR but keeping the branch so you can consider moving it to gearbox.

PR #350, opened 2020-02-25
feat: aggregate endpoint for assets/events

BugGambit:


src/types/types.ts
```diff
@@ -301,36 +301,58 @@ export type AssetDescription = string;
...
-    source?: AssetSource;
-    createdTime?: DateRange;
-    lastUpdatedTime?: DateRange;
-    /**
-     * Filtered assets are root assets or not
-     */
-    root?: boolean;
-    externalIdPrefix?: ExternalIdPrefix;
-  };
+  filter?: AssetFilterObject;
```
polomani: similar to `filter?: TimeseriesFilterProps;`
will be better:
```suggestion
  filter?: AssetFilterProps;
```

src/types/types.ts
```diff
@@ -802,6 +824,26 @@ export interface EventSearchRequest extends Limit {
...
+  /**
+   * Filter on events with strict matching.
+   */
+  filter?: EventFilter;
+}
+
+/**
+ * Response from event aggregate endpoint
+ */
+export interface EventAggregateResponse {
```
polomani: I am thinking maybe? because it's not a final response, it's just an item, what do you think?
```suggestion
export interface EventAggregate {
```

the same with assets

polomani (changes requested):

polomani (approved):

polomani:
remember to squash with `feat(assets-events): aggregate endpoint` or something

PR #356, opened 2020-03-12
Update 2.24 release changelog with data set information. 

maciejmrowiec:


hhenriques:
You need to define a subject on your commit and also a description. So something like "chore(dataset): update changelog"

maciejmrowiec:
hate f-pushing upstream ....

hhenriques (approved):

PR #361, opened 2020-03-18
feat(timeseries, sequences, files): added aggregate support

hhenriques:
Added aggregate support for timeseries, sequences and files.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -194,7 +194,7 @@ export interface Asset
 /**
  * Response from asset aggregate endpoint
  */
-export interface AssetAggregate {
+export interface AggregateResponse {
   /**
    * Size of the aggregation group
    */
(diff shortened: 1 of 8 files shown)


src/types/types.ts
```diff
@@ -194,7 +194,7 @@ export interface Asset
 /**
  * Response from asset aggregate endpoint
  */
-export interface AssetAggregate {
+export interface AggregateResponse {
```
polomani, line 194: Unfortunately we cannot just remove it now, since someone might have imported it.
But you can create synonym
`export interface AssetAggregate = AggregateResponse;`

and typedoc needs to be fixed
> Response from asset aggregate endpoint

not valid anymore

polomani (commented):
looks good, except that one thing

src/types/types.ts
polomani: also you need to do the same with
`EventAggregate` in that case

hhenriques: In hindsight, we should have one for each resource type. The API might diverge in the future.

polomani: that was an initial thought, yes.

polomani (approved):

PR #363, opened 2020-03-30
docs: more fetching examples

BugGambit:


samples/nodejs/quickstart.js
```diff
@@ -33,10 +39,18 @@ async function quickstart() {
     return false;
   });
 
+  const firstPage = await client.assets.list();
```
PhilippeBettler, line 42: How do we get the Array of items from firstPage or secondPage?

polomani (approved):

samples/nodejs/quickstart.js
BugGambit: @PhilippeBettler like this:
`const myAssets = firstPage.items`

PR #365, opened 2020-04-21
feat: add ongoing events filtering

mrtraser:


src/types/types.ts
```diff
@@ -738,7 +738,7 @@ export interface EventChangeById extends EventPatch, InternalId {}
 
 export interface EventFilter extends CreatedAndLastUpdatedTimeFilter {
   startTime?: DateRange;
-  endTime?: DateRange;
+  endTime?: DateRange | { isNull: boolean };
```
polomani: better to define new type or interface
and then you can add typedoc as well

polomani (commented):
they've also added `activeAtTime`, can you make support for that as well?

src/types/types.ts
mrtraser: smthing like `NullableDateRange`?

polomani: `NullableProperty`,  `NullableFilter`, or `NullableFilterProperty` would be more flexible

mrtraser: done

src/__tests__/resources/events.integration.spec.ts
```diff
@@ -192,6 +192,17 @@ describe('Events integration test', () => {
...
+
+    test('activeAtTime', async () => {
+      const { items } = await client.events.list({
+        filter: {
+          activeAtTime: { min: 101 },
+        },
+        limit: 10,
+      });
+
+      expect(items.length).toBe(1);
```
mrtraser: what if somebody adds an event to the tenant which will be at the same time range?
is it better to use `toBeGreaterThan(0)`?

polomani: .toBeGreaterThan(0) would be enough

polomani: in case someone adds some random event

BugGambit: What about https://github.com/cognitedata/cognite-sdk-js/pull/365/commits/fa6802afca6a8e5962708b190adfcbf67d972b03 ?

BugGambit: Ok. I can do that and revert https://github.com/cognitedata/cognite-sdk-js/pull/365/commits/fa6802afca6a8e5962708b190adfcbf67d972b03 if you want

polomani: https://github.com/cognitedata/cognite-sdk-js/commit/fa6802afca6a8e5962708b190adfcbf67d972b03 is fine

mrtraser: yep, a much better solution

polomani: I just thought you finished)

BugGambit: I reverted back. Test failed due to eventual consistency

BugGambit: I reverted back. Test failed due to eventual consistency

polomani (approved):
looks good

but
DO NOT MERGE BEFORE API RELEASE

PR #367, opened 2020-05-14
feat(assets,events): add support for dataSetId.isNull filter

dmivankov:


src/types/types.ts
```diff
@@ -1318,6 +1326,8 @@ export interface Node3DProperties {
   };
 }
 
+export type NullableFilter = { isNull: boolean };
```
polomani: I'd suggest using already existing type.  it's on the line below:`NullableProperty<T>`

polomani (commented):
try running `yarn lint --fix`, there are some issues

src/types/types.ts
dmivankov: NullableProperty is for two options T and isNull, for dataSetId.isNull there's no T option. Or is it legal to use NullablePropery<void> or something like that?

polomani: ah, my bad. then it's fine,
just fix the linter issue

BugGambit:
Shouldn't we include the other resources here? (sequences, time series, files etc.)

dmivankov:
Other resources don't implement it yet, hopefully will add that soon. Probably will use the same sdk PR

polomani:
@dmivankov also the tests are a bit exhaustive, i believe 1 test with either `isNull: true` or `false`  should be enough.

polomani:
@dmivankov it's much easier than it seems
just run `yarn lint --fix` locally

polomani:
@dmivankov looks good, are we waiting for something, or this can be released?

dmivankov:
Waiting for python sdk review with similar change, then we can release both sdks + docs at roughly the same time. But if not separate releases also should work

polomani (approved):
wait, so we can merge with python at the same time

andersfylling:
@dmivankov any updates on this?

PR #374, opened 2020-06-15
fix: export Revision3DStatus type

maksnester:
Need this type in the CDF console app to show some UI based on the revision status.


polomani:
@maksnester you need to change git commit message according to the rules
`fix: export Revision3DStatus enum` (I believe it's not much of a feature)

maksnester:
@polomani thanks. Reworded it.

polomani (approved):

PR #375, opened 2020-06-18
feat(synthetic time series): add synthetic time series query api call

haved:
Adds a new class for the synthetic time series api with one public method, accessed through the regular time series api class. Adds tests and a snippet for the docs.

### DIFF
--- a/src/__tests__/resources/timeseries.integration.spec.ts
+++ b/src/__tests__/resources/timeseries.integration.spec.ts
@@ -155,6 +155,15 @@ describe('Timeseries integration test', () => {
     expect(result[0].name).toBe(name);
   });
 
+  test('synthetic query', async () => {
+    const [ts1] = createdTimeseries;
+    const result = await client.timeseries.syntheticQuery([
+      { expression: `24 * TS{id:${ts1.id}}`, start: 0, end: 100, limit: 100 },
+    ]);
+    expect(result.length).toBe(1);
+    expect(result[0].datapoints.length).toBeLessThanOrEqual(100);
+  });
+
   test('delete', async () => {
     await client.timeseries.delete(createdTimeseries.map(({ id }) => ({ id })));
   });
--- a/src/__tests__/resources/timeseries.integration.spec.ts
+++ b/src/__tests__/resources/timeseries.integration.spec.ts
@@ -158,8 +158,20 @@ describe('Timeseries integration test', () => {
   test('synthetic query', async () => {
     const [ts1, ts2] = createdTimeseries;
     const result = await client.timeseries.syntheticQuery([
-      { expression: `24 * TS{externalId='${ts1.externalId}', aggregate='average', granularity='1h'}`, start: 0, end: 100, limit: 100 },
-      { expression: `map(TS{id=${ts2.id}}, ['A', 'B'], [1, 2], 0)`, start: "2d-ago", end: "0h-ago", limit: 4 }
+      {
+        expression: `24 * TS{externalId='${
+          ts1.externalId
+        }', aggregate='average', granularity='1h'}`,
+        start: 0,
+        end: 100,
+        limit: 100,
+      },
+      {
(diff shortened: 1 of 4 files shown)


src/__tests__/resources/timeseries.integration.spec.ts
```diff
@@ -155,6 +155,30 @@ describe('Timeseries integration test', () => {
...
+        expression: `24 * TS{externalId='${
+          ts1.externalId
+        }', aggregate='average', granularity='1h'}`,
+        start: 0,
+        end: 100,
+        limit: 100,
+      },
+      {
+        expression: `map(TS{id=${ts2.id}}, ['A', 'B'], [1, 2], 0)`,
+        start: '2d-ago',
```
polomani: I believe one is enough, let's try to not slow down the tests much

src/__tests__/resources/timeseries.integration.spec.ts
```diff
@@ -155,6 +155,30 @@ describe('Timeseries integration test', () => {
...
+        end: '0h-ago',
+        limit: 4,
+      },
+    ]);
+    expect(result.length).toBe(2);
+    expect(result[0].datapoints.length).toBeLessThanOrEqual(100);
+    expect(result[1].datapoints.length).toBeLessThanOrEqual(4);
+    expect(result[1].isString).toBe(false);
```
polomani: this is quite exhaustive
a single `expect(result.length).toBe(1);` is enough
we don't need to test all the api cases here, only make sure it gives 200 basically

src/resources/timeSeries/syntheticTimeSeriesApi.ts
```diff
@@ -0,0 +1,29 @@
...
+  SyntheticQueryResponse
+> {
+  public query = (
+    items: SyntheticQuery[]
+  ): Promise<SyntheticQueryResponse[]> => {
+    return this.querySyntheticEndpoint(items);
+  };
+
+  private get queryUrl() {
+    return this.url('query');
```
polomani: this is redundant, just move it to the line 20 inside `callRetrieveEndpoint`

src/resources/timeSeries/timeSeriesApi.ts
```diff
@@ -149,6 +158,24 @@ export class TimeSeriesAPI extends BaseResourceAPI<
...
 
+  /**
+   * [Synthetic Query](https://docs.cognite.com/api/v1/#operation/querySyntheticTimeseries)
+   *
+   * ```js
+   * await client.timeseries.syntheticQuery([
+   *   {
+   *     expression: "24 * TS{externalId='production/hour', aggregate='average', granularity='1d'}",
+   *     start: 0,
+   *     end: 0,
```
polomani:
```suggestion
   *     end: 0,
```
both start and end can't be zero.
maybe use relative time?

src/types/types.ts
```diff
@@ -1922,6 +1922,37 @@ export const SortOrder = {
...
 
+export interface SyntheticDateBase {
+  timestamp?: Date;
+}
+
+export interface SyntheticDataError extends SyntheticDateBase {
+  error: string;
+}
+
+export type SyntheticDataPoint = SyntheticDataValue | SyntheticDataError;
```
polomani:
```suggestion
export type SyntheticDatapoint = SyntheticDataValue | SyntheticDataError;
```
just to be consistent with other places

src/types/types.ts
```diff
@@ -1922,6 +1922,37 @@ export const SortOrder = {
  */
 export type SortOrder = 'asc' | 'desc';
 
+export interface SyntheticDateBase {
+  timestamp?: Date;
```
polomani: we can reuse `GetDatapointMetadata` instead of this, right?

polomani (approved):
do not merge before api release (friday is fine I guess?)

polomani:
and we need to squash as well

PR #376, opened 2020-06-19
chore: add codeSnippets/ to .gitignore

haved:


polomani (changes requested):
LG, but this is not a fix (it doesn't require a new release), but rather a `chore`

haved:
I changed the title of the pr, is that sufficient if it's squash merged?

polomani:
@haved we can rename yes, but just for the future
as we don't always squash

PR #377, opened 2020-06-22
feat: values aggregates for events

BugGambit:


src/types/types.ts
```diff
@@ -713,15 +718,28 @@ export type DeleteAssetMapping3D = AssetMapping3DBase;
...
 
+export interface EventAdvancedAggregate extends EventCountAggregate {
+  /**
+   * The field name(s) to apply the aggregation on. Currently limited to one field.
+   */
+  fields: ('type' | 'subtype' | 'dataSetId')[];
+  /**
+   * Type of aggregation to apply. values - Get unique values (upto max 1000) in the specified field ordered by frequency.
+   */
+  aggregate: 'values';
```
hhenriques: Did you consider `uniqueValues` instead of `values`? Think it's more explicit

src/types/types.ts
```diff
@@ -713,15 +718,28 @@ export type DeleteAssetMapping3D = AssetMapping3DBase;
 export type EXECUTE = 'EXECUTE';
 
 /**
- * Query schema for asset aggregate endpoint
+ * Query schema for event aggregate endpoint
  */
-export interface EventAggregateQuery {
+export type EventAggregateQuery = EventCountAggregate | EventAdvancedAggregate;
+
+export interface EventCountAggregate {
```
hhenriques: Why the change in name here? The `filter` is generic to all aggregates, not only `count` right?

src/types/types.ts
BugGambit: The schema is defined here: https://docs.google.com/document/d/1xrts3tu_7ldJDNiY6RJQ4Qt48onSC5R-8pmtiM9MViA/edit#bookmark=id.5x5u3i4p67l4

hhenriques: `values` seems a bit generic though. I first reviewed the test and just looking at the code, I couldn't understand what the result of the request was

src/types/types.ts
BugGambit: Fixed

src/types/types.ts
```diff
@@ -713,15 +718,30 @@ export type DeleteAssetMapping3D = AssetMapping3DBase;
...
   filter?: EventFilter;
 }
 
+export interface EventCountAggregate extends EventBaseAggregate {}
+
+export interface EventAdvancedAggregate extends EventBaseAggregate {
+  /**
+   * The field name(s) to apply the aggregation on. Currently limited to one field.
+   */
+  fields: ('type' | 'subtype' | 'dataSetId')[];
```
polomani, line 736: can we add a code snippet in the `eventsApi.ts:aggregate(...)` ?
even I don't understand how it works from the first glance

hhenriques: @polomani the plan is to have a function per aggregate type (so a single API can spawn multiple SDK functions). Not sure how snippets would work there

BugGambit (commented):
Awaiting @ponnapz api change for this one to pass the CI and be merged

src/types/types.ts
polomani: whose plain is that?
so far we had one function per endpoint why do you want to change that?
this will introduce a huge gap between api schemas and typescript types
what are the reasons?

src/resources/events/eventsApi.ts
```diff
@@ -56,11 +58,28 @@ export class EventsAPI extends BaseResourceAPI<CogniteEvent> {
...
+  /**
+   * [Aggregate events](https://docs.cognite.com/api/v1/#operation/aggregateEvents)
+   *
+   * ```js
+   * const aggregates = await client.events.uniqueValuesAggregate({ filter: { assetIds: [1, 2, 3] }, fields: ['subtype'] });
+   * console.log('Unique values: ', aggregates)
+   * ```
+   */
+  public uniqueValuesAggregate = (
+    query: EventUniqueValuesAggregate
```
polomani, line 75: this is a bit messy, I don't think we can have it.
it's not java, we don't need to complicate things

polomani: also this kills any idea of having even semi generated sdks

BugGambit: Why? Without it then the return type is really messy since it is dependent on the function call.
We discussed also the semi generated sdks scenario with this approach with @hhenriques

polomani:
https://github.com/cognitedata/cognite-sdk-js/pull/378
maybe it doesn't defend my argument about generated sdks, but still needs your opinion

polomani:
I closed that alternative pr.
I think we should change it to `client.events.aggregate.uniqueValues(...)` when we do a breaking change

src/types/types.ts
```diff
@@ -191,13 +191,22 @@ export interface Asset
   parentExternalId?: CogniteExternalId;
 }
 
-export interface AggregateResponse {
+export type AggregateResponse = CountAggregateResponse;
```
polomani: I don't think we should introduce aliases.
can't we just use AggregateResponse directly?

src/types/types.ts
```diff
@@ -713,15 +722,26 @@ export type DeleteAssetMapping3D = AssetMapping3DBase;
...
+export interface EventBaseAggregate {
   /**
    * Filter on events with strict matching.
    */
   filter?: EventFilter;
 }
 
+export interface EventCountAggregate extends EventBaseAggregate {}
+
+export interface EventUniqueValuesAggregate extends EventBaseAggregate {
```
polomani: same here, too many types

polomani (commented):
I would suggest to remove aliases, other than that looks good

src/types/types.ts
polomani: ok

polomani (approved):

polomani:
when squash
`feat(events): unique values aggregate`

polomani (approved):

PR #378, opened 2020-06-22
I1/event agggregates alternative solution

polomani:


src/resources/events/eventsApi.ts
```diff
@@ -53,12 +55,22 @@ export class EventsAPI extends BaseResourceAPI<CogniteEvent> {
...
-  ): Promise<EventAggregate[]> => {
-    return super.aggregateEndpoint(query);
+  public aggregate = <
+    T extends EventAggregateQuery | EventAggregateUniqueValues
+  >(
+    query: T
+  ) => {
+    return super.aggregateEndpoint<
+      T,
+      T extends EventAggregateUniqueValues
```
BugGambit, line 70: How will we easliy expand this for new aggregation methods?

src/__tests__/resources/events.integration.spec.ts
```diff
@@ -95,6 +95,19 @@ describe('Events integration test', () => {
     expect(aggregates[0].count).toBeDefined();
   });
 
+  test('values aggregate', async () => {
+    const aggregates = await client.events.aggregate({
+      filter: {
+        source: 'WORKMATE',
+      },
+      fields: ['type'],
+      aggregate: 'values',
```
hhenriques, line 104: Not a fan of the usability here. The user has to know what fields are compatible with each aggregate type.

polomani: but it's typed, you cannot randomly choose a string

polomani:
we'll go for methods approach

PR #379, opened 2020-06-23
feat: labels (only for assets)

BugGambit:
Need to add integration tests for asset.labels when labels is available in API v1.

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -594,7 +594,7 @@ export interface LabelDefinition extends ExternalLabelDefinition {
   createdTime: Date;
 }
 
-export interface LabelDefinitionFilter {
+export interface LabelDefinitionFilterRequest extends FilterQuery {
   filter?: {
     /**
      * Returns the label definitions matching that name.
@@ -607,11 +607,6 @@ export interface LabelDefinitionFilter {
     externalIdPrefix?: ExternalIdPrefix;
   };
 }
-
-export interface LabelDefinitionFilterRequest
-  extends LabelDefinitionFilter,
-    Cursor,
-    Limit {}
 
 /**
  * Filter on data sets with exact match
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -594,18 +594,20 @@ export interface LabelDefinition extends ExternalLabelDefinition {
   createdTime: Date;
 }
 
-export interface LabelDefinitionFilterRequest extends FilterQuery {
-  filter?: {
-    /**
-     * Returns the label definitions matching that name.
-     */
-    name?: string;
+export interface LabelDefinitionFilter {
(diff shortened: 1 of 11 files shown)


src/types/types.ts
```diff
@@ -543,6 +548,71 @@ export interface DataSetChangeByExternalId extends DataSetPatch, ExternalId {}
...
+   * A list of labels to remove to the resource
+   */
+  remove?: Label[];
+}
+
+export interface ExternalLabelDefinition extends Label {
+  /**
+   * Name of the label.
+   */
+  name: string;
```
vincent-cognite, line 585: no ? makes it required, right?

vincent-cognite:
no support for types in the js sdk? (just curious)

vincent-cognite (approved):

src/types/types.ts
hhenriques: `name` is required on the API right?

vincent-cognite: yes (I'm just typescript curious)

src/types/types.ts
```diff
@@ -543,6 +548,71 @@ export interface DataSetChangeByExternalId extends DataSetPatch, ExternalId {}
 
 export interface DataSetChangeById extends DataSetPatch, InternalId {}
 
+export type Label = ExternalId;
```
polomani, line 558: do we need this alias?

src/types/types.ts
```diff
@@ -543,6 +548,71 @@ export interface DataSetChangeByExternalId extends DataSetPatch, ExternalId {}
...
+     * Filter external ids starting with the prefix specified
+     */
+    externalIdPrefix?: ExternalIdPrefix;
+  };
+}
+
+export interface LabelDefinitionFilterRequest
+  extends LabelDefinitionFilter,
+    Cursor,
+    Limit {}
```
polomani: instead of `Cursor, Limit`
extends `FilterQuery`

polomani: maybe make it similar to `DataSetFilterRequest` so we don't empty types?
I mean move `filter?: LabelDefinitionFilter ` in here and unwrap the current `LabelDefinitionFilter`

polomani (commented):
looks good but I am not sure about having `blablaDefinition` everywhere
I would remove alias, use `ExternalId` type and name types as `blavlaLabel` without `Definition`

src/types/types.ts
BugGambit: Yes, for the UX

BugGambit:
> no support for types in the js sdk? (just curious)

@vincent-cognite No, it is not part of v1

polomani (approved):
don't merge before API release
remember to squash

PR #381, opened 2020-06-29
refactor: move structure to a monorepo with multiple packages

haved:
Turns one package into three:
 - `@cognite/sdk-core` in `packages/core`
 - `@cognite/sdk` in `packages/stable`
 - `@cognite/sdk-beta` in `packages/beta`

Using yarn workspaces, and lerna for management, CI automatic version bumping and releasing to npm.  
Samples are also workspaces now, but not public, and filtered out of most commands.

Changes travis to add node 14 and only do deploys and full build on one node version.  
Replaces sematic release with lerna.  
Docs on gh-pages includes a folder for beta.  
All packages get their own readme, also on npm.  
Update repo readme to link to packages.  
Only snippets from stable are exported to service contracts, but all are tested (tsc compiled).  
Typescript config magic (explained [here](./guides/monorepo.md)) to make Go To Definition work across packages.  
Codecoverage is counted as one percentage across all packages.


tsconfig.build.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.common",
+  "exclude": ["**/__tests__/**", "*.unit.spec.ts"]
```
polomani:
```suggestion
  "exclude": ["**/__tests__/**", "**/*.spec.ts", "**/*.test.ts"]
```
we have integrations as well

.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
```
polomani: can we move this to `.eslintrc.js`?
seems like overkill to have a separate file

.eslintrc.js
```diff
@@ -0,0 +1,38 @@
...
+        "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/ban-ts-comment": "off",
+        "react/prop-types": "off",
+        "@typescript-eslint/no-unused-vars": "off",
+        "prettier/prettier": "error",
+
+        //added manually while migrating from tslint
+        "@typescript-eslint/ban-types": "off",
+        "@typescript-eslint/no-empty-function": "off",
+        "no-constant-condition": "off" //makes while(true) illegal
```
polomani: please remove
`"@typescript-eslint/no-empty-function": "off"`
and `"no-constant-condition": "off"`
and use `// eslint-disable-line` instead

README.md
```diff
@@ -3,131 +3,60 @@
 </a>
 
 Cognite Javascript SDK
-==========================
-[![Build Status](https://travis-ci.org/cognitedata/cognite-sdk-js.svg?branch=master)](https://travis-ci.org/cognitedata/cognite-sdk-js)
-[![codecov](https://codecov.io/gh/cognitedata/cognite-sdk-js/branch/master/graph/badge.svg)](https://codecov.io/gh/cognitedata/cognite-sdk-js)
+======================
+[![Build Status](https://travis-ci.org/cognitedata/cognite-sdk-js.svg?branch=v1)](https://travis-ci.org/cognitedata/cognite-sdk-js)
+[![codecov](https://codecov.io/gh/cognitadata/cognite-sdk-js/branch/v1/graph/badge.svg)](https://codecov.io/gh/cognitedata/cognite-sdk-js)
```
polomani:
```suggestion
[![codecov](https://codecov.io/gh/cognitedata/cognite-sdk-js/branch/v1/graph/badge.svg)](https://codecov.io/gh/cognitedata/cognite-sdk-js)
```

README.md
```diff
@@ -3,131 +3,60 @@
...
 
 ## Development
 
-There is one integration test that requires a api key. In order to run this, you need an api key for the `cognitesdk-js` tenant. Talk to any of the contributors or leave an issue and it'll get sorted. Travis will run the test and has its own api key.
+The sdk is implemented as a package in a monorepo, together with core logic, beta versions and samples.
+See [monorepo.md](./guides/monorepo.md) for details.
+
+### Testing
+
+There is one integration test that requires an api key. In order to run this, you need an api key for the `cognitesdk-js` tenant. Talk to any of the contributors or leave an issue and it'll get sorted. Travis will run the test and has its own api key.
```
polomani:
```suggestion
This repo contains some integration tests that require a CDF api key for `cognitesdk-js` tenant. Talk to any of the contributors or leave an issue and it'll get sorted. Travis will run the test and has its own api key.
```

README.md
```diff
@@ -3,131 +3,60 @@
...
 
-## Contributing
+### Contributing
 
 Contributions welcome! See the [code of conduct](./CODE_OF_CONDUCT.md).
 
-## Release
+### Release
 
 Our releases are fully [automated](https://github.com/semantic-release/semantic-release).
```
polomani: this is not true anymore

README.md
```diff
@@ -137,6 +66,11 @@ Only basic steps are needed:
...
-## CHANGELOG
+### CHANGELOG
+
+Each package in the monorepo has its own changelog.
+- [@cognite/sdk](./packages/stable/CHANGELOG.md)
+- [@cognite/sdk-beta](./packages/beta/CHANGELOG.md)
+- [@cognite/sdk-core](./packages/core/CHANGELOG.md)
 
-You can find it [here](./CHANGELOG.md).
+You can find the changelog from before the monorepo [here](./CHANGELOG.md).
```
polomani: can we
copy everything from current changelog to `./packages/stable/CHANGELOG.md`
and then just reference it in a root log?

guides/monorepo.md
```diff
@@ -0,0 +1,88 @@
...
+and a root `tsconfig.json` with the path aliasing.
+The latter two extend the first.
+
+Package specific configuration is mostly specifying output folder, and extending the corresponding
+file in the project root. You only need to specify output folder in build configuration,
+since your IDE will reference by source through path aliases.
+
+Note that tests are excluded in `tsconfig.build.json`, but are otherwise included
+to make vscode handle Go To Definition from inside tests.
+This convention of `tsconfig.build.json` also includes the code snippet extraction,
```
polomani: what's that? a bit lost here

guides/monorepo.md
```diff
@@ -0,0 +1,88 @@
...
+**Note:** Scoped packages (e.g. `@cognite/sdk`) need special confirmation that they are public the first time they are published.
+This is specified in the individual `package.json` files: `"publishConfig": { "access": "public" }`.
+
+## Samples
+The `samples/` folder contains several samples, each top level folder being a private package in the workspace.
+Though tecnically in the workspace for ease of testing, they would also work as-is as stand alone dependants of the sdk.
+All samples' names end in `-sample`, to make filtering easy. `yarn test` does not test samples.
+
+**NOTE** The `samples/react/` folder contains samples that are not tested, because a) they don't have tests,
+and b) including the packages in the workspace would give lots of versioning conflicts with `react-scripts`.
```
polomani: this kind of documentation gets outdated very quickly, plz remove it

lerna.json
```diff
@@ -0,0 +1,15 @@
+{
+  "packages": [
+    "packages/**",
+    "samples/*"
+  ],
+  "version": "independent",
+  "npmClient": "yarn",
+  "command": {
+    "version": {
+      "ignoreChanges": ["**/__tests__/**", "**/*.unit.spec.ts"],
```
polomani: normally this shouldn't happen because we use `--conventional-commits`,
but if you really want to use it then we could add md files at least
```suggestion
      "ignoreChanges": ["**/*.md", "**/*.test.ts", "**/*.spec.ts"],
```

packages/beta/README.md
```diff
@@ -0,0 +1,33 @@
...
+but use the following command to install:
+```
+$ yarn add @cognite/sdk@npm:@cognite/sdk-beta
+```
+or with npm
+```
+$ npm install @cognite/sdk@npm:@cognite/sdk-beta --save
+```
+
+This will install the beta package as a dependency, but under the normal sdk name.
```
polomani: what's normal?)
just explain that it creates alias, so you can reference the sdk just
the same way using `@cognite/sdk`

packages/beta/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+With the beta package installed under an alias, you don't need to modify your code
+to access beta features. Import the `CogniteClient` as you normally would:
+```js
+import { CogniteClient } from '@congite/sdk';
+```
+
+## Documentation
+
+Though it looks like the normal `CogniteClient`, it's actually named `CogniteClassBeta`.
```
polomani: what's normal?) why do they need to know it's called that way?

packages/beta/README.md
```diff
@@ -0,0 +1,33 @@
...
+See the typedoc [here](https://cognitedata.github.io/cognite-sdk-js/beta/classes/cogniteclientbeta.html).
+
+\[beta API reference\](TODO: link to beta api reference here)
```
polomani: remove todo, we will merge it before we have any beta

packages/beta/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js",
+    "test-snippets": "yarn extract-snippets && yarn tsc -p codeSnippets/tsconfig.build.json"
+  },
+  "dependencies": {
+    "@cognite/sdk": "2.32.0",
+    "@cognite/sdk-core": "1.0.0",
+    "cross-fetch": "^3.0.4",
+    "lodash": "^4.17.11",
+    "query-string": "^5.1.1",
+    "url": "^0.11.0"
```
polomani: can you review the packages here?
I don't think we use cross-fetch or query-string directly

packages/beta/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "private": false,
+  "name": "@cognite/sdk-beta",
+  "license": "Apache-2.0",
```
polomani, line 4: you need to add license file in this folder and in stable as well

packages/beta/rollup.config.js
```diff
@@ -0,0 +1,45 @@
...
+        typescript({
+            tsconfig: './tsconfig.build.json',
+            typescript: require('typescript'),
+            clean: true,
+            exclude: '**/__tests__/**',
+        }),
+        json({
+            // All JSON files will be parsed by default,
+            // but you can also specifically include/exclude files
+            include: ['./package.json', '../stable/package.json'],
```
polomani: do we need `../stable/package.json`? we have it as a dependency now, shouldn't be needed during the build process anymore

packages/beta/src/__tests__/testUtils.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright 2020 Cognite AS
+
+import { Constants } from '@cognite/sdk-core';
+import CogniteClient from '../cogniteClient';
+
+export function setupClient(baseUrl: string = Constants.BASE_URL) {
+  return new CogniteClient({ appId: 'JS SDK integration tests', baseUrl });
```
polomani:
```suggestion
  return new CogniteClient({ appId: 'JS SDK integration tests (beta)', baseUrl });
```

packages/beta/src/cogniteClient.ts
```diff
@@ -0,0 +1,34 @@
...
+import { version } from '../package.json';
+
+export default class CogniteClientBeta extends CogniteClientStable {
+  /**
+   * Create a new SDK client (beta)
+   *
+   * @param options Client options
+   *
+   * To use the beta sdk, alias `"@cognite/sdk": "@cognite/sdk-beta@^<version>"` in `package.json`
+   * The beta SDK exports the client with the same name as stable, meaning you don't need to change any code.
```
polomani:
```suggestion
   * The beta SDK exports the client with the same name as stable, meaning you don't need to change any code.
```
```suggestion
   * The beta SDK exports the client with the same name as stable, meaning you don't need to change any imports.
```

packages/beta/src/cogniteClient.ts
```diff
@@ -0,0 +1,34 @@
...
+} from '@cognite/sdk';
+import { version } from '../package.json';
+
+export default class CogniteClientBeta extends CogniteClientStable {
+  /**
+   * Create a new SDK client (beta)
+   *
+   * @param options Client options
+   *
+   * To use the beta sdk, alias `"@cognite/sdk": "@cognite/sdk-beta@^<version>"` in `package.json`
```
polomani:
```suggestion
   * For smooth transition between stable sdk and beta, you may create an alias `"@cognite/sdk": "@cognite/sdk-beta@^<version>"` in `package.json`
```

packages/beta/src/cogniteClient.ts
```diff
@@ -0,0 +1,34 @@
...
+   * // can also specify a base URL
+   * const client = new CogniteClient({ ..., baseUrl: 'https://greenfield.cognitedata.com' });
+   * ```
+   */
+  constructor(options: ClientOptions) {
+    super(options);
+  }
+
+  protected get version() {
+    return `${version} (@cognite/sdk-beta)`;
```
polomani: the header format is `CogniteJavaScriptSDK:${this.version}`, so it should be
```suggestion
    return `${version}-beta`;
```
instead

packages/beta/src/index.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright 2020 Cognite AS
+
+export * from '@cognite/sdk';
+export * from './types';
+export {
+  default as CogniteClient,
+  default as CogniteClientBeta,
```
polomani: why 2 exports?

packages/beta/package.json
polomani: and core ofc

packages/beta/src/types.ts
```diff
@@ -0,0 +1,3 @@
+// Copyright 2020 Cognite AS
+
+export * from '@cognite/sdk';
```
polomani, line 3: this is not great

can we export just types from `@cognite/dist/src/sdk/types`?

packages/core/README.md
```diff
@@ -0,0 +1,8 @@
+A core library used by cognite Javascript SDKs, published as `@cognite/sdk-core`.
```
polomani, line 1: mention that this can be reused to create custom sdks, alphas for instance

packages/core/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "private": false,
+  "name": "@cognite/sdk-core",
+  "license": "Apache-2.0",
+  "repository": "cognitedata/cognite-sdk-js",
+  "homepage": "https://github.com/cognitedata/cognite-sdk-js#readme",
+  "main": "dist/index.js",
+  "types": "dist/index.d.js",
```
polomani: do you know why stable/beta has `"types": "dist/src/index.d.js",` but this one no?

polomani: can we make so none of them has src part?

packages/stable/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "esCheck": "es-check es5 './dist/index.js'",
+    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js",
+    "test-snippets": "yarn extract-snippets && yarn tsc -p codeSnippets/tsconfig.build.json"
+  },
+  "dependencies": {
+    "@cognite/sdk-core": "1.0.0",
+    "cross-fetch": "^3.0.4",
+    "lodash": "^4.17.11",
+    "query-string": "^5.1.1",
+    "url": "^0.11.0"
```
polomani: again check the dependencies, we might not use `cross-fetch` here at least

samples/react/authentication-with-popup/package.json
```diff
@@ -1,11 +1,12 @@
 {
-  "name": "authentication-with-popup",
-  "version": "0.1.0",
+  "name": "authentication-with-popup-sample",
+  "version": "1.0.0",
   "private": true,
   "dependencies": {
+    "@cognite/sdk": "^2.31.1",
```
polomani:
```suggestion
    "@cognite/sdk": "^2",
```
does this work? or then it doesn't fetch the local one?

.eslintrc.js
```diff
@@ -0,0 +1,38 @@
...
+    },
+    env: {
+        browser: true,
```
polomani, line 45: `"file-header": [true, "Copyright \\d{4} Cognite AS"]`
does this feature exist in es lint?
it is important to add

packages/stable/src/types.ts
```diff
@@ -1,4 +1,17 @@
...
+  CogniteInternalId,
+  Cursor,
+  ExternalId,
+  FilterQuery,
+  IdEither,
+  InternalId,
+  Limit,
+} from '@cognite/sdk-core';
+
+export * from '@cognite/sdk-core';
```
polomani: maybe we could have types as a module in core?

polomani: we shouldn't export everything from here

packages/core/src/baseCogniteClient.ts
```diff
@@ -439,56 +349,38 @@ export default class CogniteClient {
     this.httpClient.addOneTimeHeader(X_CDF_SDK_HEADER, value);
   }
 
-  private initAPIs = () => {
-    const defaultArgs: [CDFHttpClient, MetadataMap] = [
+  protected initAPIs() {}
+
+  protected get version() {
+    return 'unreleased';
```
polomani: since now we are publishing it on npm we should give the correct version (just in case)

you could try to do the same as on beta
but with `-core` tag instead

samples/README.md
```diff
@@ -11,7 +11,7 @@ This folder contains examples on how to use the SDK in different frameworks.
 
 - Clone this repository locally. See guide [here](https://help.github.com/en/articles/cloning-a-repository).
 - Install NodeJS on your machine. You can download NodeJS [here](https://nodejs.org/en/download/).
-- Open this folder in your terminal.
+- Open a sample project in a terminal.
```
polomani:
```suggestion
- Open a sample project directory in a terminal.
```
?

samples/nodejs/package.json
```diff
@@ -1,5 +1,11 @@
 {
+  "name": "nodejs-sample",
+  "version": "1.0.0",
+  "private": true,
   "license": "Apache-2.0",
+  "dependencies": {
+    "@cognite/sdk": "^2.32.0"
```
polomani:
```suggestion
    "@cognite/sdk": "^2"
```
should be enough?

samples/typescript/package.json
```diff
@@ -1,11 +1,14 @@
...
+  "name": "typescript-sample",
+  "version": "1.0.0",
+  "private": true,
   "license": "Apache-2.0",
   "scripts": {
     "tsc": "tsc",
     "test": "yarn tsc && node build/quickstart.js"
   },
   "dependencies": {
+    "@cognite/sdk": "^2.31.1",
```
polomani:
```suggestion
    "@cognite/sdk": "^2",
```

scripts/extractCodeSnippets.js
```diff
@@ -1,87 +1,64 @@
...
 const fs = require('fs');
-const tsconfig = require('../tsconfig.json');
 
-function stripMarkdownCodeSnippet(rawCode) {
-  return rawCode
-    .replace('```js', '')
-    .replace('```', '')
-    .trim();
-}
+const docRegEx = /https:\/\/(doc.cognitedata.com|docs.cognite.com)\/api\/v1\/#operation\/([a-zA-Z0-9]+)/g;
```
polomani: have you rebased on v1?

scripts/prSnippets.sh
```diff
@@ -0,0 +1,36 @@
...
+
+echo "Copying stable SDK's code snippets"
+cp "../${stablePackage}/codeSnippets/index.json" ./versions/v1/js-sdk-examples.json
+if ! git diff --quiet ; then
+    echo "service contracts code snippets have changed. making pull request"
+    git add ./versions/v1/js-sdk-examples.json
+    git commit -m "$message"
+    #git push origin "$branchName"
+    #hub pull-request -m "$message"
+    #echo "pull request made"
```
polomani: well you need to uncomment this

scripts/prSnippets.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/bash
+# exit when any command fails (DON'T -x !!)
```
polomani, line 2: there is `/dev/null 2>&1` on git command is it still not safe enough?

scripts/prSnippets.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "Cloning service contract repo"
+git clone https://$GITHUB_TOKEN@github.com/cognitedata/service-contracts.git >/dev/null 2>&1
+cd service-contracts
+
+echo "Checking out branch $branchName"
+git checkout -b "$branchName"
+
+echo "Copying stable SDK's code snippets"
+cp "../${stablePackage}/codeSnippets/index.json" ./versions/v1/js-sdk-examples.json
+if ! git diff --quiet ; then
```
polomani: why this complications? the pr will not be opened if there no changes anyway
you are overengineering it

scripts/versionAndPush.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "Checking out \$TRAVIS_BRANCH: $TRAVIS_BRANCH"
+git checkout $TRAVIS_BRANCH
+
+git config --global user.email "cognite-cicd@users.noreply.github.com"
+git config --global user.name "Cognite CICD"
+
+# Find the new remote with github token, careful not to leak secrets
+OLD_REMOTE="$(git remote get-url origin)"
+REPO_PATH=$(echo $OLD_REMOTE | sed -En "s/^.*github.com.(.*)/\1/p")
+git remote set-url origin "https://$GITHUB_TOKEN@github.com/$REPO_PATH" > /dev/null 2>&1
```
polomani: is it really needed? any other options?

scripts/versionAndPush.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/bash
+# exit upon error (DON'T -x !!)
+set -e
+
+# Don't be in detached head mode
+echo "Checking out \$TRAVIS_BRANCH: $TRAVIS_BRANCH"
+git checkout $TRAVIS_BRANCH
```
polomani, line 7: how can you possibly be in detached mode?

scripts/ciScript.sh
```diff
@@ -0,0 +1,22 @@
...
+    echo "Only doing build and test"
+    yarn --ignore-engines
+    yarn global add lerna@3.1.4 #Supports node 8
+    lerna run build --stream
+    lerna run test --stream
+else
+    yarn
+    yarn global add codecov@3.7.0
+    yarn commitlint-travis
+    yarn validateDocLinks || true
```
polomani, line 15: what's with this one again?

packages/stable/README.md
```diff
@@ -0,0 +1,93 @@
...
+  const assets = await client.assets
+    .list()
+    .autoPagingToArray({ limit: 100 });
+}
+quickstart();
+```
+
+## Documentation
+
+ - [API documentation](https://doc.cognitedata.com)
```
polomani:
```suggestion
```
this link is useless here

packages/stable/README.md
```diff
@@ -0,0 +1,93 @@
...
+ - [API reference documentation](https://doc.cognitedata.com/api/v1)
+ - [CogniteClient typedoc](https://cognitedata.github.io/cognite-sdk-js/classes/cogniteclient.html)
```
polomani:
```suggestion
 - [JS SDK reference documentation](https://cognitedata.github.io/cognite-sdk-js/classes/cogniteclient.html)
```

packages/stable/README.md
```diff
@@ -0,0 +1,93 @@
+Cognite Javascript SDK
+======================
+The package `@cognite/sdk` provides convenient access to the stable [Cognite API](https://doc.cognitedata.com/dev/)
+from applications written in client- or server-side javascript.
+
+The SDK supports authentication through api-keys (_for server-side applications_) and bearer tokens (for web applications).
```
polomani, line 6: add a link to auth guide

samples/README.md
```diff
@@ -11,7 +11,7 @@ This folder contains examples on how to use the SDK in different frameworks.
...
 - Install dependencies by running `$ npm install` or with yarn `$ yarn`.
```
polomani:
```suggestion
- Install dependencies by running `npm install` or with yarn `yarn`.
```
I know it's not your code, just cleaning up a bit

README.md
```diff
@@ -3,131 +3,60 @@
...
+### Testing
+
+There is one integration test that requires an api key. In order to run this, you need an api key for the `cognitesdk-js` tenant. Talk to any of the contributors or leave an issue and it'll get sorted. Travis will run the test and has its own api key.
 Set the environment variable `COGNITE_PROJECT` to `cognitesdk-js` and `COGNITE_CREDENTIALS` to your api-key.
 
 Run all tests:
 
 ```bash
 $ yarn
+$ yarn build
```
polomani: could you please remove dollar sign for all commands in `*.md` files?
some people don't know that you shouldn't copy it
(i know it's not your code, but we are touching it anyway)

packages/beta/README.md
```diff
@@ -0,0 +1,33 @@
...
+With the beta package installed under an alias, you don't need to modify your code
+to access beta features. Import the `CogniteClient` as you normally would:
+```js
+import { CogniteClient } from '@congite/sdk';
+```
+
+## Documentation
+
+Though it looks like the normal `CogniteClient`, it's actually named `CogniteClassBeta`.
+See the typedoc [here](https://cognitedata.github.io/cognite-sdk-js/beta/classes/cogniteclientbeta.html).
```
polomani:
```suggestion
See the reference doc [here](https://cognitedata.github.io/cognite-sdk-js/beta/classes/cogniteclientbeta.html).
```
try to not use a lot of library/technology specific words

polomani: why not relative url?

packages/core/src/index.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright 2020 Cognite AS
+import * as Constants from './constants';
+import * as GraphUtils from './graphUtils';
+import * as TestUtils from './testUtils';
+
+export { MetadataMap } from './metadata';
+export { BaseResourceAPI } from './baseResourceApi';
+export { default as BaseCogniteClient } from './baseCogniteClient';
+export * from './baseCogniteClient';
+export * from './types';
```
polomani: types should be a separate module so we can export them from stable/beta safely

polomani: so make the same as `Constants` basically

polomani (changes requested):
good stuff!
but there are many changes, that's why there will be many suggestions / clarifications need

besides the comments, could you please run something that'll EOF in files that missing it?

meanwhile I'll try to import and use it in some project, to see if we forgot anything

README.md
haved: I think the stable changelog is completely generated by lerna every time, so I'm not sure how easy it is to include previous changes. Will look into it

polomani:
I've also imported the package to gearbox.js and it seems that we messed up somewhere with AssetClass / Asset exports.
could you have a look at that?

UPD: never mind, I just had to link `sdk-core` as well

lerna.json
haved: If you change the readme and want that to be reflected on npm, then we should maybe just let people use conventional commits to do that if they like. I agree that bumping versions on npm because of a readme is weird, but bumping due to tests changing is probably worse, as the tests are in no way included in the package.

packages/beta/package.json
haved: Looking at index.js from dist/ tells us only `@cognite/sdk` is `require()`d. Should we keep `@cognite/sdk-core` as a dependency? If we ever use it it should be a direct dependency, and lerna will make sure stable and beta never depend on different versions. Should we just keep it as a direct dependency?

packages/core/package.json
haved: I notices that they differed, but never found out why. Will have another look

README.md
haved: I changed it to a link to lerna version

guides/monorepo.md
haved: Replaced with new paragraph

samples/react/authentication-with-popup/package.json
haved: The react samples are removed from the workspaces either way, so it doesn't get the local one. However, we should probably just say `^2` because lerna is no longer going to update that. We could in theory have yarn workspaces not include react, and have lerna workspaces include react, but I'll have to see if that breaks something.

haved: But are you suggesting we use ^2 in every sample in general?

.eslintrc.js
haved: added with https://github.com/Stuk/eslint-plugin-header

README.md
polomani: I pretty much doubt they generate it from scratch, at least that's not how `semantic-release` used to work

lerna.json
polomani: we definitely won't bump the patch version just because of readme (I've never heard ppl doing that), it'll wait until the next release
so for me it's no difference between readme and tests tbh

polomani: so it's just up to developers if they're following the semantic versioning (which they must)

samples/react/authentication-with-popup/package.json
polomani: > But are you suggesting we use ^2 in every sample in general?

yes

>The react samples are removed from the workspaces either way
```
  "workspaces": [
    "packages/**",
    "samples/*"
  ],
```
are they?

packages/stable/src/types.ts
haved: I'm having a hard time forward-exporting part of `@cognite/sdk-core`. I think we may have to manually specify what should be forwared.

scripts/extractCodeSnippets.js
haved: Yes, there were lots of merge conflicts, can make sure that everything from v1 is kept

haved: I changed the file, it's now based on the 10 day old bugfix version from Fredrik

scripts/prSnippets.sh
haved: I kept the if, but do if git commit instead, as the commit will fail if no changes have occured. I think it's good to keep the  echos seing as the script is run without `set -x`, and git commands are piped to null

scripts/prSnippets.sh
haved: The piping hides output of commands, -x will print the commands that run, with macros expanded

lerna.json
haved: I added *.md

Unrelated: Should we not use `**/__test__/**` as a check for tests?
I know jest only cares about `.test.ts(x?)` and `.spec.ts(x?)`, but there can be other files related to tests in __test__, so I think we should keep it ignored, both in lerna and in tsconfig.

On that note: Do we ever use `.tsx` tests? They would be picked up by jest but not lerna, so we should probably make them agree

packages/beta/rollup.config.js
haved: Don't think we do.. removed it

samples/react/authentication-with-popup/package.json
haved: Yes, as they are two folders deep. Only one asterisk on `samples/*`
It was to avoid having all of react and react-scripts as dev dependencies

scripts/versionAndPush.sh
haved: It avoids hard coding the source repo, but I agree it's very ugly. We do hard code a lot of links, so we might as well just write `cognitedata/cognite-sdk-js.git` and make it easy to find-replace if the path ever does change.

haved: I did that

scripts/versionAndPush.sh
haved: When travis clones, it does
```
git checkout -qf 67c3cc9b5c52a5b47707eabb47da25922256a5a4
```
so we have to attach to a branch again.
This script is only run on deploys, so we don't have to worry about pull request builds having the branch name of the target.

scripts/ciScript.sh
haved: Validating doc links is going to fail, because we have absolute links to e.g. beta doc pages.
This makes the command run, and if it fails, true is ran, which always has exit code 0, so travis won't stop.
The travis log will show which links failed. This is maybe not a permanent solution, but we need to deploy first

packages/beta/README.md
haved: Because this readme will appear on github and on npm as well. We could have a separate readme as the front page of typedoc, but then we would still need absolute links for this npm/Github README.

packages/core/src/index.ts
haved: ES6 imports and exports are very strict, so even if we do export it as Types, we still can't forward export that from stable in a nice way (no Types.Types). Will look some more

guides/monorepo.md
```diff
@@ -0,0 +1,90 @@
...
+
+Then you can run the samples' tests.
+```
+$ yarn test-samples
+```
+
+## Documentation
+Since each package gets its own npm page, they all have `README.md` files.
+These are also used as front pages of typedoc pages.
+This means all links must be absolute.
```
BugGambit: 👍

packages/beta/README.md
```diff
@@ -0,0 +1,32 @@
...
+$ npm install @cognite/sdk@npm:@cognite/sdk-beta --save
+```
+
+This will install the package `@cognite/sdk-beta` as a dependency, but aliased as `@cognite/sdk`.
+In `package.json`, it will look like this:
+```
+    "@cognite/sdk": "npm:@cognite/sdk-beta@^X.X.X"
+```
+
+This will dowload `@cognite/sdk-beta` and pretend it is `@cognite/sdk`.
```
BugGambit:
```suggestion
This will download `@cognite/sdk-beta` and pretend it is `@cognite/sdk`.
```

README.md
haved: I put the old changelog in stable. Will see what lerna does once we merge this.

packages/beta/package.json
haved: I have decided to keep core as a dependency

scripts/versionAndPush.sh
haved: Actually, I use `$TRAVIS_REPO_SLUG`

lerna.json
```diff
@@ -0,0 +1,15 @@
+{
+  "packages": [
+    "packages/**",
+    "samples/*"
+  ],
+  "version": "independent",
+  "npmClient": "yarn",
+  "command": {
+    "version": {
+      "ignoreChanges": ["**/*.md/", "**/__tests__/**", "**/*.spec.ts", "**/*.test.ts"],
```
polomani, line 10: this list is never ending, we should rely on semantic messages and common sense instead

lerna.json
polomani: doesn't matter imho, can be changed later

lerna.json
haved: Doing some testing in https://github.com/haved/cognite-sdk-js

packages/beta/README.md
```diff
@@ -0,0 +1,32 @@
...
+This will download `@cognite/sdk-beta` and pretend it is `@cognite/sdk`.
+With the beta package installed under an alias, you don't need to modify your code
+to access beta features. Import the `CogniteClient` as you normally would:
+```js
+import { CogniteClient } from '@congite/sdk';
+```
+
+## Documentation
+
+When you import `CogniteClient` from the beta sdk, you will in reality get the subclass `CogniteClientBeta`.
```
polomani, line 29: again, why this matter? it's just an internal name

packages/stable/src/types.ts
polomani: probably fine, can you do that plz? (list the exports I mean)

scripts/extractCodeSnippets.js
```diff
@@ -71,17 +72,26 @@ codeSnippets.forEach((snippets, operationId) => {
...
-console.log(
-  `TS config for code snippets saved to: ${snippetsFolder}/tsconfig.json`
-);
+// We make a tsconfig.build.json that extends the normal tsconfig.build.json
+tsconfig.extends = '../../../tsconfig.build.json';
+const tsbuildconfigpath = path.join(snippetsFolder, './tsconfig.build.json');
+fs.writeFileSync(tsbuildconfigpath, JSON.stringify(tsconfig, null, 2) + '\n');
+console.log(`TS build config for code snippets saved to: ${tsbuildconfigpath}`);
+
+// We also make a tsconfig.json for IDEs only that includes path aliases
```
polomani: what is this? `tsconfig` is temporary, it's just to run typescript once

packages/stable/README.md
```diff
@@ -0,0 +1,93 @@
...
+    .list()
+    .autoPagingToArray({ limit: 100 });
+}
+quickstart();
+```
+
+## Documentation
+
+ - [API reference documentation](https://doc.cognitedata.com/api/v1)
+ - [CogniteClient typedoc](https://cognitedata.github.io/cognite-sdk-js/classes/cogniteclient.html)
```
polomani: what about this again?

polomani (commented):
a few things left

(besides lerna bumping the patch every time)

scripts/extractCodeSnippets.js
polomani: I also get this
```
Error: Cannot write file '/Users/***/cognite-sdk-js/packages/core/package.json' because it would overwrite input file.
```

lerna.json
polomani: then we'll go for a release branch

polomani: can you create a wildcard? `release/*`?

packages/beta/README.md
haved: Because that is the name you see in the documentation. Not good, but right now it is

scripts/extractCodeSnippets.js
haved: I get that warning every time, don't know what it is about.
About the tsconfig, I make two of them, with the only difference being what tsconfig it inherits from, such that Go To Definition works from snippets as well

packages/stable/README.md
haved: I forgot to push, fixed now

scripts/extractCodeSnippets.js
polomani: > Go To Definition works from snippets as well

why do we need that at all?
code snippets need to be just tested and pushed to contracts

polomani: > I get that warning every time, don't know what it is about

I guess you are literally trying to write to the file that exists

packages/stable/src/types.ts
```diff
@@ -1,4 +1,30 @@
...
+  ExternalId,
+  FilterQuery,
+  IdEither,
+  InternalId,
+  Limit,
+  LogoutUrl,
+  CursorResponse,
+  ItemsWrapper,
+  ClientOptions,
+} from '@cognite/sdk-core';
```
polomani, line 35: why there are two similar exports?

> refactor: export enough of cores types for stable and beta to compile

also, all current types from core should be exported (because it used to be like that)

(unless you have objections)

packages/beta/README.md
polomani: ok

packages/stable/src/types.ts
haved: I agree all current types should be exported, but it looks very bad to export * from core in types.ts. I also agree the overlap between imports and exports looks weird. We could perhaps in typescript 3.8 do export types, but even then, not all types from core are supposed to be exported from types.ts in stable (such as BaseCogniteClient, yes it's a type, but not in this sense). Do we have to manually export every type in `core/src/types.ts`?

scripts/versionAndPush.sh
```diff
@@ -2,6 +2,13 @@
 # exit upon error (DON'T -x !!)
 set -e
 
+if ! [[ $TRAVIS_COMMIT_MESSAGE =~ "[release]" ]]; then
```
polomani:
```suggestion
if ! [[ $TRAVIS_COMMIT_MESSAGE =~ "chore: release" ]]; then
```
maybe? then it'll be compatible with angular rules

CONTRIBUTING.md
```diff
@@ -24,24 +24,35 @@ only use squashing if the resulting squash commit can accuratly describe the cha
...
+
+Updating and uploading npm packages only happens if the HEAD commit of the main branch
+contains `[release]` in its description. When CI/CD sees this, it will use lerna to update
+package versions of changed packages based on commit messages, and add the 
+changes to the changelogs. The changes are comitted to the main branch
+with the new versions as git tags, and the new package versions are uploaded to npm.
+
+We restrict new npm releases to `[release]`-tagged commits becuase lerna is
+quite agressive in its versioning. Changes to any file not ignored by lerna will
+cause a PATCH bump. Markdown files and tests are ignored, but chaning anything else,
```
polomani:
```suggestion
cause a PATCH bump. Markdown files and tests are ignored, but changing anything else,
```

CONTRIBUTING.md
```diff
@@ -24,24 +24,35 @@ only use squashing if the resulting squash commit can accuratly describe the cha
...
+like a comment in a source file, will trigger a new version,
+irrespective of conventional commits.
+
+This does *not* mean you should store unfinished work on the main branch.
+Another package may be ready for release, and once anyone pushes a `[release]`
+commit, all changed packages are updated.
+
+Also keep in mind that the `[release]` commit has to be the HEAD of
```
polomani: maybe we just advice to trigger the push only for administrators?
so it won't be as a part of pr but rather the direct push

polomani: we have now a bit of a problem now with people merging to master without us knowing it

packages/stable/src/types.ts
polomani: I meant export all types that are in `@cognite/sdk-core/src/types.ts` ofc
not everything))
types.ts is basically consisting of types that were previously exported anyway (it's around 20 of them)

scripts/versionAndPush.sh
polomani: maybe some regex `chore: .*[release].*` (not sure about syntax)

lerna.json
haved: We are no longer using release branch, and even if you want to patch old major versions, you don't use lerna

scripts/extractCodeSnippets.js
haved: I removed the `tsconfig.json` from the snippets folder, which was good because IDEs use the parent folder `tsconfig.json`. The snippets still have their own `tsconfig.build.json`

scripts/versionAndPush.sh
haved: The CI/CD version bump commit is called `chore(release): publish new package versions [skip ci]`.
We have to make a commit before this commit that triggers it. Preferably they look a bit different, but it should maybe be clear the first triggered the last. I agree we can require a certain prefix

haved: However, sometimes you may want to use a squash commit and also trigger a build, like:
```
feat(assets): Add endpoint [release]
```

polomani: agree

packages/stable/src/types.ts
haved: I did a new pull request for this, just to avoid messing with the base of 3 other PRs

CONTRIBUTING.md
```diff
@@ -37,22 +37,18 @@ with the new versions as git tags, and the new package versions are uploaded to
...
+cause a PATCH bump. Markdown files and tests are ignored, but changing anything else,
 like a comment in a source file, will trigger a new version,
 irrespective of conventional commits.
 
 This does *not* mean you should store unfinished work on the main branch.
-Another package may be ready for release, and once anyone pushes a `[release]`
-commit, all changed packages are updated.
+Another package may be ready for release, and once a `[release]`
+commit is pushed, all changed packages are updated.
+Repository administators should be in controll of `[release]` commits.
```
polomani:
```suggestion
Repository administrators should be in control of `[release]` commits.
```

.eslintrc.js
```diff
@@ -0,0 +1,49 @@
...
+        sourceType: "module"
+    },
+    rules: {
+        "@typescript-eslint/explicit-module-boundary-types": "off",
+        "@typescript-eslint/no-inferrable-types": "off",
+        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/ban-ts-comment": "off",
+        "react/prop-types": "off",
+        "@typescript-eslint/no-unused-vars": "off",
```
mrtraser: Is it any reason to have unused vars?

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
+Contributing
+============
+
+Contributions are welcome, and this document details how changes can be made and submitted,
+and eventually included in a relase. We use monorepo tooling, and a git setup for CI/CD.a PATCH bump
```
mrtraser:
```suggestion
and eventually included in a release. We use monorepo tooling, and a git setup for CI/CD PATCH bump 
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+
+ - For fixes, start the commit with `fix:`. This will bump PATCH.
+ - For features, start the commit with `feat:`. This will bump MINOR.
+ - For breaking changes, add an exclamation mark, e.g. `feat!:` or `fix!:`. This will bump MAJOR.
+ - For extra details in the changelog, you can specify a scope, e.g. `feat(assets):`.
+ - For other changes, use `refactor:`, `test:`, `docs:`, `chore:` etc.
+
+## Pull request
+Make a pull request from your branch to the main branch. When merging the pull request,
+only use squashing if the resulting squash commit can accuratly describe the change as a single conventional commit.
```
mrtraser:
```suggestion
only use squashing if the resulting squash commit can accurately describe the change as a single conventional commit.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+Documentation is built and deployed, and code snippets
+are exported to the service contract repo as a pull request.
+
+Updating and uploading npm packages only happens if the HEAD commit of the main branch
+contains `[release]` in its description. When CI/CD sees this, it will use lerna to update
+package versions of changed packages based on commit messages, and add the 
+changes to the changelogs. The changes are comitted to the main branch
+with the new versions as git tags, and the new package versions are uploaded to npm.
+
+We restrict new npm releases to `[release]`-tagged commits becuase lerna is
```
mrtraser:
```suggestion
We restrict new npm releases to `[release]`-tagged commits because lerna is
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+quite agressive in its versioning. Changes to any file not ignored by lerna will
```
mrtraser:
```suggestion
quite aggressive in its versioning. Changes to any file not ignored by lerna will
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+
+To add a release commit to a clean working tree, use the command
+```bash
+git checkout v1
+git pull
+git commit --allow-empty -m "chore: trigger [release]"
+git push
+```
+
+Also keep in mind that the `[release]` commit has to be the HEAD of
```
mrtraser:
```suggestion
Also, keep in mind that the `[release]` commit has to be the HEAD of
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,109 @@
...
+main, and travis only runs on the HEAD. If HEAD has changed by the time
+the versiong happens, travis will fail.
```
mrtraser:
```suggestion
the versioning happens, travis will fail.
```

guides/monorepo.md
```diff
@@ -0,0 +1,106 @@
...
+All dependencies can be installed from the repository root:
+```
+yarn
+```
+To `build`, `clean`, `test`, `lint` or `lint:fix`, run the command like so:
+```
+$ yarn <command>
+```
+If in the root folder, the operation is run in all packages (excluding samples), in topological order.
+Remember to build all packages before testing, and the enviroment variables needed to run certain tests.
```
mrtraser:
```suggestion
Remember to build all packages before testing, and the environment variables needed to run certain tests.
```

guides/monorepo.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Versioning and releasing
+See [Contributing](../CONTRIBUTING.md).
+
+### About Lerna version
+Our setup has lerna individually update versions of packages based on the messages of commits affecting each package.
+It will also document the changes to each package's `CHANGELOG.md`.
+You can make the changelog clearer by specifying scope in commit messages, like `feat(assets):`
+
+If no changes were made in a package, or the commits only affect files ignored by `lerna version` (see lerna.json),
```
mrtraser:
```suggestion
If no changes are made in a package, or the commits only affect files ignored by `lerna version` (see lerna.json),
```

packages/core/src/utils.ts
```diff
@@ -76,6 +78,86 @@ export function promiseCache<ReturnValue>(
...
+      failed,
+      errors,
+    };
+  }
+
+  return responses;
+}
+
+/** @hidden */
+// TODO: refactor
```
mrtraser: what do we need to refactor here?

polomani: I think it was there before, better to remove. not relevant

scripts/ciScript.sh
mrtraser: can we have some comment about it here?

polomani (approved):
I've been following this from the beginning, so I don't think there is much sense for me to start over.

@mrtraser Just need a stamp from you I guess
@BugGambit if you have time

mrtraser (approved):
Have very minor suggestions on PR, put it above.

Quite impressed with monorepo solution, looks promising as for me, well done 👍

packages/core/src/utils.ts
polomani: removed the comment

.eslintrc.js
polomani: fixed

PR #382, opened 2020-06-29
Added application domains to the project update object

henne90gen:


frxstrem:
Add it here as well:

https://github.com/cognitedata/cognite-sdk-js/blob/83c67ec7c693eb4b0e7df0670299a7b27e99743e/src/types/types.ts#L1454-L1456

src/types/types.ts
```diff
@@ -2174,6 +2176,11 @@ export type UrlName = string;
  */
 export type ValidDomains = string[];
 
+/**
+ * A set of domains hosting applications used with the project. If missing, a default set of domains will be used. Once application domains are set, they may be individually removed, but cannot be removed entirely in order to fall back to the default list.
```
hhenriques:
```suggestion
 * A set of domains used within the project for application hosting. If missing, a default set of domains will be used. Once application domains are set, they may be individually removed. However, the removal of all available domains means that login to the project won't be permitted.
```

hhenriques (approved):
Looks good. Just a minor suggestion regarding documentation

src/types/types.ts
polomani: @hhenriques you know it is copied from service contracts docs, right?

henne90gen: I adjusted the description to match the service contract exactly.

polomani:
could you guys look into changelog now and not do such thing next time?
https://github.com/cognitedata/cognite-sdk-js/blob/v1/CHANGELOG.md

polomani:
same here
https://github.com/cognitedata/cognite-sdk-js/releases

polomani:
never mind, fixed myself.
but commit history is quite ugly, no ones fault. but be more careful next time

PR #383, opened 2020-07-01
docs(labels): fix code snippet api url

polomani:
fyi @fredrik955 I agree that `createLabelDefinitions` is better, however it is not like that in api reference :(


BugGambit:
Autch! I can look into that

polomani:
in favour of https://github.com/cognitedata/service-contracts/pull/970

PR #386, opened 2020-07-06
fix: import only used parts of lodash

maksnester:
Noticed that sdk brings the whole lodash and decided that it's something that easy to fix. 

In our demo examples these changes save ~70KB. Not really a size for reveal viewer because it loads tons of data, but maybe it will be useful to other apps. 

I had to enable `esModuleInterop` in tsconfig, to make these `lodash/isFunction` imports work. As a side effect import statement for `nock` module is also simplified. 

About lodash imports, they also [recommend](https://www.npmjs.com/package/lodash) to import what you use:

> // Cherry-pick methods for smaller browserify/rollup/webpack bundles.
> var at = require('lodash/at');
> var curryN = require('lodash/fp/curryN');


polomani:
@mrtraser you did research, right? it seems to be a matter of setting rollup/treeshaking in your app.
@maksnester have you tried using it?

maksnester:
@polomani yes, I tried that inside our examples for the reveal. Examples have a create-react-app based setup, which means the same setups are quite often used. Here is the analyzer tool report (1 - v1, 2 - this branch):

image

image

mrtraser:
@maksnester, have you tried to keep the existing import approach with `loadash-es` instead?

maksnester:
@mrtraser no, I haven't

maksnester:
@mrtraser @polomani OK, I tried to simply replace `from 'lodash'` with `from 'lodash-es'` and here are results for my examples build:

image

As you can see, no tree-shaking magic happens. Result bundle even bigger with `lodash-es`

maksnester:
FYI: es-check reports circular dependency

```
(!) Circular dependency: src\resources\login.ts -> src\resources\login.ts
```

polomani:
@maksnester weird, on the bottom here https://medium.com/@martin_hotell/tree-shake-lodash-with-webpack-jest-and-typescript-2734fa13b5cd it says that `lodash-es` with `esModuleInterop` should work

polomani:
@maksnester I tried a few approaches and yours the only one that works, indeed.
we have a lot of structural changes, so I had to reapply your commit manually. sorry

rebased now

PR #387, opened 2020-07-07
Fix/date parsing

### LINKED ISSUE #333: Date conversion is too intrusive
The SDK will traverse the response body and convert any field that matches a certain set of names to javascript Date objects (https://github.com/cognitedata/cognite-sdk-js/blob/b95c4e769920526f9f20d5e1e8e5070b5b34f7e4/src/utils.ts#L105L120).

This is not necessarily valid to do in the general case.

For example, say I have a RAW table which contains a column named `timestamp`, which stores *seconds* since unix epoch. The Date constructor takes a value in *milliseconds*  since epoch, so this will not yield the correct result. And even if it did, we don't want the SDK to make assumptions about raw data.

Instead, I think we should tailor this logic for each resource type, and only convert dates where we know it is safe to do so.

haved:
This is just a sample of a possible approach to selective date parsing in APIs, to fix #333. The fix is applied only to the asset and apiKey APIs to begin with


packages/core/src/baseResourceApi.ts
```diff
@@ -56,17 +57,37 @@ export abstract class BaseResourceAPI<
...
+      : httpClient;
+  }
 
   protected getMetadataMap() {
     return this.map;
   }
 
+  protected get cdfHttpClient(): CDFHttpClient {
+    if (this.httpClient instanceof CDFHttpClient) return this.httpClient;
+    return this.httpClient.cdfHttpClient;
```
polomani: maybe you don't like current abstractions but this doesn't make it any better

I believe you should either add logic to `postRequest`/`preRequest` in `CDFHttpClient`, or even inherit it, but not this

packages/core/src/dateParser.ts
```diff
@@ -0,0 +1,151 @@
...
+  async delete<ResponseType>(
+    path: string,
+    options?: HttpRequestOptions
+  ): Promise<HttpResponse<ResponseType>> {
+    return this.parseResponse(
+      await this.httpClient.delete(path, this.parseRequest(options))
+    );
+  }
+
+  async patch<ResponseType>(
```
polomani: try to have it in one place, there are some methods that cover all cases at the same time

5 times copy-paste is not good

packages/stable/src/api/events/eventsApi.ts
```diff
@@ -17,6 +21,14 @@ import {
 } from '../../types';
 
 export class EventsAPI extends BaseResourceAPI<CogniteEvent> {
+  /**
+   * Specify that dates should be parsed in requests and responses
+   * @hidden
+   */
+  protected getDateParser() {
+    return new DateParser(['items'], ['createdTime', 'lastUpdatedTime']);
```
polomani: ideally you just need to pass some kind of array, this is too imperative
(if you can't find better way, it's ok)

scripts/endpointDateFinder.js
```diff
@@ -0,0 +1,185 @@
...
+  console.log(
+    'Also prints out places where you arrive at {} objects or objets with string keys'
+  );
+  console.log('Must be run in a folder with a ./codeSnippets/docs.json file');
+  console.log('Example usage in packages/stable/ folder:');
+  console.log('$ node ../../endpointDateFinder.js CogniteClient client');
+  process.exit(0);
+}
+
+const result = findAllDateReturns(
```
polomani: do you want to commit this? it's completely unmaintainable

packages/stable/DateReturns.txt
```diff
@@ -0,0 +1,101 @@
...
+>> Date >> (await client.timeseries.create())[0].lastUpdatedTime
+>> Date >> (await client.timeseries.list()).createdTime
+>> Date >> (await client.timeseries.list()).lastUpdatedTime
+>> Date >> (await client.timeseries.retrieve())[0].createdTime
+>> Date >> (await client.timeseries.retrieve())[0].lastUpdatedTime
+>> Date >> (await client.timeseries.search())[0].createdTime
+>> Date >> (await client.timeseries.search())[0].lastUpdatedTime
+>> Date >> (await client.timeseries.update())[0].createdTime
+>> Date >> (await client.timeseries.update())[0].lastUpdatedTime
+>> Date >> (await client.viewer3D.retrieveRevealRevision3D()).createdTime
```
polomani: is this going to be committed?

scripts/endpointDateFinder.js
polomani: it also uses code snippets that are not supposed to a be a source of truth

UPD: not code snippets, typedoc, sorry

packages/core/src/dateParser.ts
```diff
@@ -0,0 +1,152 @@
...
+   * where lead can be any field from the lead parameter,
+   * and the date is a field from the dates parameter.
+   * Lead can also be the name of an array, which will allow all elements of that array
+   *
+   *  lead[5].lead.date
+   *
+   * @param lead the keys that lead to dates
+   * @param dates the keys that can be parsed as Date objects
+   */
+  constructor(private lead: string[], private dates: string[]) {}
```
polomani:
```suggestion
  constructor(private path: string[], private datePropNames: string[]) {}
```
and now you don't even need to document it anymore

packages/core/src/dateParser.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/* Old keys used for date parsing:
+  'createdTime',
+  'lastUpdatedTime',
+  'uploadedTime',
+  'deletedTime',
+  'timestamp',
+  'sourceCreatedTime',
+  'sourceModifiedTime'
```
polomani: why is it here?

polomani (commented):
overall what we are trying to do here is basically just to pass the prop names from each api to the function that'll convert those props to Dates.
it seems like there too much code for this here, and it's overcomplicated

also
```
constructor(private lead: string[], private dates: string[]) {}
```
better to be generically typed, otherwise you might misspell some prop name and it will be hard to notice

packages/stable/DateReturns.txt
haved: I decided to remove the script and this file. It was nice when digging through returns, but it's not maintainable at all

packages/core/src/dateParser.ts
haved: Just as a reminder of old names used while writing changes. We never actually received any "timestamp" fields, but we use it a lot in requests. Removed now

haved:
Mainly a fix of #333, but also fixes CDF-7248

packages/stable/src/types.ts
```diff
@@ -2173,3 +2173,14 @@ export interface Versioned3DFile {
...
+export type DatePropPath = 'items' | 'columns' | 'datapoints';
+export type DatePropName =
+  | 'createdTime'
+  | 'lastUpdatedTime'
+  | 'sourceCreatedTime'
+  | 'sourceModifiedTime'
+  | 'uploadedTime'
+  | 'deletedTime'
+  | 'timestamp';
+export type DatePropFilter = [DatePropPath[], DatePropName[]];
```
polomani: we `export * from './types';` in `index.ts`, so we cannot have it here

polomani: this is a file with API types only

packages/stable/src/api/timeSeries/timeSeriesApi.ts
```diff
@@ -46,6 +47,14 @@ export class TimeSeriesAPI extends BaseResourceAPI<
     );
   }
 
+  /**
+   * Specify what fields in json responses should be parsed as Dates
```
polomani: why is this documented in every single api? one place should be enough

packages/core/src/login.ts
```diff
@@ -63,15 +67,17 @@ export function getIdInfoFromApiKey(
...
+  });
 }
 
 /** @hidden */
 export function getIdInfoFromAccessToken(
   httpClient: CDFHttpClient,
   accessToken: string
 ): Promise<null | IdInfo> {
-  return getIdInfo(httpClient, {
+  return getIdInfo(httpClient.get.bind(httpClient), {
```
polomani, line 70: this comes all the way from `createAuthenticateFunction` -> `Login.loginSilently(httpClient, ...)` maybe do it there instead?

packages/core/src/login.ts
```diff
@@ -63,15 +67,17 @@ export function getIdInfoFromApiKey(
   httpClient: CDFHttpClient,
   apiKey: string
 ): Promise<null | IdInfo> {
-  return getIdInfo(httpClient, { [API_KEY_HEADER]: apiKey });
+  return getIdInfo(httpClient.get.bind(httpClient), {
```
polomani: it doesn't seem we use this function?
If not, better to remove

packages/stable/src/api/3d/revisions3DApi.ts
```diff
@@ -29,6 +30,14 @@ export class Revisions3DAPI extends BaseResourceAPI<Revision3D> {
     this.nodes3DApi = new Nodes3DAPI(resourcePath, httpClient, map);
   }
 
+  /**
+   * Specify what fields in json responses should be parsed as Dates
+   * @hidden
+   */
+  protected getDateProps(): DatePropFilter {
+    return [['items'], ['createdTime']];
```
polomani: since we don't have Date parsing tests for all apis, have you checked all the apis and their props manually as well?
(I know you had a generator) but it might've missed something

polomani (commented):
I rebased the branch
Overall looks good, some minor comments

packages/stable/src/types.ts
polomani: fixed myself

packages/stable/src/api/timeSeries/timeSeriesApi.ts
polomani: removed

packages/core/src/login.ts
polomani: unrelated

packages/stable/src/api/3d/revisions3DApi.ts
polomani: checked that, might have missed something, hard to find out

packages/core/src/login.ts
polomani: removed

packages/stable/src/api/sequences/sequencesApi.ts
```diff
@@ -39,6 +39,16 @@ export class SequencesAPI extends BaseResourceAPI<Sequence> {
     );
   }
 
+  /**
+   * @hidden
+   */
+  protected getDateProps() {
+    return this.pickDateProps(
```
BugGambit, line 46: This interface is unintuitive to me. What does this mean 🤔 ?

polomani: `this.pickDateProps` is typed, so that should help a bit
but you can try to suggest something else
we can make `['items']` as a default path I guess

BugGambit: After we talked yesterday then I understand the approach due to typing

packages/core/src/__tests__/login.spec.ts
```diff
@@ -138,32 +136,6 @@ describe('Login', () => {
       },
     };
 
-    test('successful getIdInfoFromApiKey', async () => {
```
BugGambit, line 141: Why are we removing this?

BugGambit (approved):

packages/core/src/__tests__/login.spec.ts
polomani: it's never used? as you can see I removed it and nothing happened

BugGambit: so you are sure no one of our users uses it 🤔 ?

polomani: it wasn't exported from the main module and it was hidden from ref docs

polomani: idk why we have it

polomani: but less code is always better)

PR #389, opened 2020-07-12
fix: move retryable endpoint validation to per package

haved:
Allow each package to make their own `RetryValidator`s, and specify their own list of retryable endpoints. Given to the `BaseCogniteClient` through overriding a protected method. Thus Beta can override some endpoints, but still rely on stable's list as `super.getRetryValidator()`.


packages/core/src/httpClient/retryValidator.ts
```diff
@@ -0,0 +1,73 @@
...
+) => boolean;
+
+/** @hidden */
+export type EndpointList = { [key in HttpMethod]?: string[] };
+
+const DEFAULT_MAX_RETRY_ATTEMPTS = 5;
+
+export const createRetryValidator = (
+  endpointsToRetry: EndpointList,
+  maxRetries: number
```
polomani: why not to default `= DEFAULT_MAX_RETRY_ATTEMPTS` here?

polomani (approved):
looks good to me, but I think 5 should be a default value
(don't merge before monorepo pr)

packages/stable/src/retryValidator.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright 2020 Cognite AS
+
+import {
+  createRetryValidator,
+  EndpointList,
+  RetryValidator,
+} from '@cognite/sdk-core';
+import { HttpMethod } from '@cognite/sdk-core';
+
+const MAX_RETRY_ATTEMPTS = 5;
```
polomani: we can make it default

PR #391, opened 2020-07-15
refactor: fix non-logical coupling

BugGambit:
A Label and LabelDefinition is not connected so let's not extend from label in the definition


polomani:
can we close this in favour of this https://github.com/cognitedata/cognite-sdk-js/pull/395?

PR #395, opened 2020-07-17
feat: better label type names

BugGambit:
BREAKING CHANGE:

This PR will rename the following types:
- `LabelsPatch` -> `AttachedLabelsPatch`
- `Label` -> `AttachedLabel`
- `AttachedLabelFilter` -> `LabelFilter`
- `LabelContainsAnyFilter` -> `AttachedLabelContainsAnyFilter`
- `LabelContainsAllFilter` -> `AttachedLabelContainsAllFilter`
- `ExternalLabelDefinition` -> `ExternalLabel`
- `LabelDefinition` -> `Label`
- `LabelDefinitionFilter` -> `LabelFilter`
- `LabelDefinitionFilterRequest` -> `LabelFilterRequest`


packages/stable/src/types.ts
```diff
@@ -571,30 +571,37 @@ export interface DataSetChangeByExternalId extends DataSetPatch, ExternalId {}
 
 export interface DataSetChangeById extends DataSetPatch, InternalId {}
 
-export type Label = ExternalId;
+export type AttachedLabel = ExternalId;
```
polomani, line 574: Now this looks even worse, I would suggest to get rid of it and use external id instead :)

polomani (commented):
So we ended up removing 'Definition' part after all)

PR #398, opened 2020-07-22
chore: remove unused dev deps

polomani:


package.json
```diff
@@ -55,10 +55,8 @@
     "husky": "^3.0.2",
     "jest": "^23.6.0",
     "lerna": "^3.22.1",
-    "ls": "^0.2.1",
     "markdown-link-check": "^3.8.1",
     "nock": "^10.0.6",
-    "parse-comments": "^1.0.0",
```
BugGambit, line 61: We don't use the parser anymore 🤔 ?

polomani: we removed it, not everyone agrees that we should alphabetically sort interfaces.
and also we keep forgetting to run it)

PR #400, opened 2020-07-23
Chore/upgrade packages

polomani:
things changed:
- don't build & run tests on node 8 (not breaking by itself but we cannot be confident that we are compatible anymore)
- bump jest to v26
- typescript 3.9
- removed unused dependencies
- bumped typedoc as there was a bug :
```
Error: Cannot write file './packages/stable/package.json' because it would overwrite input file
```
Due to broken support for some tsconfig.json features
- bumped rollup


samples/typescript/package.json
```diff
@@ -8,10 +8,10 @@
     "test": "yarn tsc && node build/quickstart.js"
   },
   "dependencies": {
-    "@cognite/sdk": "^2",
-    "typescript": "^3.3.4000"
+    "@cognite/sdk": "^2"
```
BugGambit, line 11: fyi we should bump this to v3

BugGambit (approved):
Approved, but see comment

samples/typescript/package.json
polomani: We would have to do this after release. Normally lerna bumps package versions automatically but in this case we run `lerna version --no-private` which doesn't touch private packages. We made it that way because otherwise we'll get a bunch of "typescript-sample@2.3.1" git tags whenever stable changes.

So we have to update manually to x^ whenever major release happens, but otherwise it is fine, it uses local instance.

haved (approved):
Approved, I added `"include": ["src/"]` to some `tsconfig.build.json` files to prevent warnings about overriding input files. (When `"excluded"` is specified, the `"outDir"` will not be exluded by default).

PR #401, opened 2020-07-24
chore: set initial versions to 0.0.0

polomani:
lerna will bump them to v1 automatically

### DIFF
--- a/packages/beta/package.json
+++ b/packages/beta/package.json
@@ -6,7 +6,7 @@
   "homepage": "https://github.com/cognitedata/cognite-sdk-js#readme",
   "main": "dist/index.js",
   "types": "dist/src/index.d.js",
-  "version": "2.32.0",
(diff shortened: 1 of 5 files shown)


haved (commented):
LGTM, but the first publish through lerna must be done with explicit major bumping. Otherwise core and beta will never leave 0.x.x. The command will probably be like this:
```
yarn lerna version major --conventional-commits --no-private
```
**Note:** this must be run in the master branch, or use `--allowBranch <branch>`.  
This will make a release commit with updated versions and changelogs,
tag the commit and push it to github.
A release will not be created, so use the Github UI to turn the tag into an actual release.
Finally:
```
yarn lerna publish from-package
```
You need to be logged into npm to run this.

### Alternativly:
We just add the `major` specifier to the ci script for only the initial release, then remove it

PR #405, opened 2020-07-28
feat(datapoints): add unit property to retrieve response object

haved:
Also add isStep where applicable.
The spec says isStep is a required response field in aggregates but optional in normal double value datapoints. Is this correct?

### DIFF
--- a/packages/stable/src/types.ts
+++ b/packages/stable/src/types.ts
@@ -672,6 +672,9 @@ export type DatapointsDeleteRequest =
   | (ExternalId & DatapointsDeleteRange);
 
 export interface DatapointAggregates extends DatapointsMetadata {
+  /**
+   * false for aggregate time series values
+   */
   isString: false;
   /**
    * Whether the timeseries is a step series or not
@@ -683,6 +686,9 @@ export interface DatapointAggregates extends DatapointsMetadata {
 export type Datapoints = StringDatapoints | DoubleDatapoints;
 
 export interface DoubleDatapoints extends DatapointsMetadata {
+  /**
+   * false for timeseries with double values
+   */
   isString: false;
   /**
    * Whether the timeseries is a step series or not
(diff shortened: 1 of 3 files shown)


packages/stable/src/types.ts
```diff
@@ -672,26 +672,23 @@ export type DatapointsDeleteRequest =
...
+  isStep: boolean;
   datapoints: DatapointAggregate[];
 }
 
 export type Datapoints = StringDatapoints | DoubleDatapoints;
 
 export interface DoubleDatapoints extends DatapointsMetadata {
-  /**
-   * Whether the time series is string valued or not.
-   */
```
polomani, line 683: can you add `isString: boolean;` to `DatapointsMetadata` and add tsdoc there?

packages/stable/src/types.ts
```diff
@@ -672,26 +672,23 @@ export type DatapointsDeleteRequest =
   | (ExternalId & DatapointsDeleteRange);
 
 export interface DatapointAggregates extends DatapointsMetadata {
+  isString: false;
+  isStep: boolean;
```
polomani, line 679: I think we should document `isStep`
you will not get duplicated tsdoc because this property is not common on all union members

polomani (changes requested):

packages/stable/src/types.ts
haved: That will give it a triple comment when hovering over, but it's probably better in total

polomani: not if it's in `DatapointsMetadata `

packages/stable/src/types.ts
haved: It's common in two of them, so you get it twice. Will see how the typedoc looks

packages/stable/src/types.ts
haved: Yes, I know it sounds wrong but it splits the same comment into three fields with the same name. I think it's better to give each a comment, in such a way that even reading them all concatenated makes sense

polomani (approved):

PR #406, opened 2020-07-28
docs: write docstring about CursorAndAsyncIterator

haved:
As a small improvement to the reference documentation, where some requests return this object without full explanation of how it can be used.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -58,6 +58,29 @@ export interface LogoutUrlResponse {
   data: { url: LogoutUrl };
 }
 
+/**
+ * A Promise to a response that can be awaited like normal,
+ * but at the risk of not getting all results due to API limits.
+ * In which case a nextCursor field is provided to request the next page.
+ * Helper methods are provided to abstract away the pagination.
+ *
+ * Example using `client.timeseries.list`:
+ * ```
+ * const response = client.timeseries.list({ filter: { assetIds: [ASSET_ID] } });
+ * ```
+ *
+ * You can iterate through all returned items like so:
+ * ```
+ * for await (const value of response) {
+ *   // do something to value
+ * }
+ * ```
+ *
+ * Or get an array of up to 1000 items like so:
+ * ```
+ * const timeseries = await response.autoPagingToArray({limit: 1000});
+ * ```
+ */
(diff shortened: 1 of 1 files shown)


packages/core/src/types.ts
```diff
@@ -58,6 +58,29 @@ export interface LogoutUrlResponse {
...
+ * You can iterate through all returned items like so:
+ * ```js
+ * for await (const value of response) {
+ *   // do something to value
+ * }
+ * ```
+ *
+ * Or get an array of up to 1000 items like so:
+ * ```js
+ * const timeseries = await response.autoPagingToArray({ limit: 1000 });
```
polomani: but this is a thing. you can get more than 1000
`client.timeseries.list({ filter: { assetIds: [ASSET_ID] }, limit: 1000 }). autoPagingToArray({ limit: 5000 });` will iterate 5 times

you can also use `autoPagingToArray({ limit: Infinity });` to fetch all

packages/core/src/types.ts
```diff
@@ -58,6 +58,29 @@ export interface LogoutUrlResponse {
   data: { url: LogoutUrl };
 }
 
+/**
+ * A Promise to a response that can be awaited like normal,
+ * but at the risk of not getting all results due to API limits.
```
polomani, line 63: example?

packages/core/src/types.ts
```diff
@@ -58,6 +58,36 @@ export interface LogoutUrlResponse {
...
+ * Helper methods are provided to abstract away the pagination.
+ *
+ * Example using `client.timeseries.list`:
+ * ```js
+ * const response = client.timeseries.list({ filter: { assetIds: [ASSET_ID] }, limit: 1000 });
+ * ```
+ *
+ * You can get up to 1000 elements with normal await:
+ * ```js
+ * const items = await response;
```
polomani:
```suggestion
 * const { items, nextCursor } = await response;
```
I believe it's like this

haved: Hmm, can one actually await a CursorAndAsyncIterator? It doesn't have a then function (?)

polomani: it does `CursorAndAsyncIterator<T> = Promise<ListResponse<T[]>> & CogniteAsyncIterator<T>`

polomani (approved):

PR #421, opened 2020-08-14
fix: exports from packages, add guide for making derived SDKs

haved:
Also adds some exports to core and stable, so there are some `feat` commits. (You could change them if you don't want a MINOR bump). Fixes casing in an api class (should not a breaking change, as the type wasn't exported).

### DIFF
--- a/guides/derived_SDKs.md
+++ b/guides/derived_SDKs.md
@@ -1,0 +1,374 @@
+# Derived SDKs HOW-TO
+
+This document serves as a guide for working with SDK packages.
+First there is an overview, describing the most important packages
+in the repository and how they interact. Then there are guides for
+creating your own SDK packages, how to add APIs to the SDK, how to
+make a derived SDK, and how to override APIs from the derived SDK.
+
+## Overview of packages
+
+### @cognite/sdk-core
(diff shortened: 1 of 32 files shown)


haved:
I was going to slack to joke about needing a review by tomorrow before I lose access, but it's already gone :(

guides/derived_SDKs.md
```diff
@@ -0,0 +1,374 @@
...
+}
+
+export class CoolThingAPI extends BaseResource<CoolThing> {
+
+    /**
+    * [Create cool things](https://doc.cognitedata.com/api/v1/#operation/postCoolThing)
+    *
+    * ```js
+    * const coolthings = [
+    *   { name: 'Shrub', coolness: 40 },
```
polomani, line 118: 🤣

polomani (approved):
looks cooler then cactus to me!

polomani:
ps: we might need to squash it to `fix:` or something, since it's not just a documentation

hhenriques (approved):
Great write up!

mrtraser (approved):
cool way to explain things, well done!

PR #422, opened 2020-08-21
feat: add logger to sdk-core

mrtraser:
Added ability to provide logger function to the `sdk-core`. For now, the logger will be called only in places inside `loginSilently` method, which don't expose error and catch it inside sdk. Can be used to cover with logging other places in the code.

### DIFF
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -133,9 +133,10 @@ export default class BaseCogniteClient {
   }
 
   public authenticate: () => Promise<boolean> = async () => {
-    throw Error(
-      'You can only call authenticate after you have called loginWithOAuth'
-    );
+    const message = `You can only call authenticate after you have called loginWithOAuth`;
+
+    this.logger.log(message);
+    throw Error(message);
   };
 
   public get project() {
@@ -238,6 +239,7 @@ export default class BaseCogniteClient {
       httpClient: this.httpClient,
       onAuthenticate,
       onTokens,
+      logger: this.logger,
     });
 
     this.httpClient.set401ResponseHandler(async (_, retry, reject) => {
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -16,7 +16,7 @@ import {
   HttpResponse,
 } from './httpClient/basicHttpClient';
 import { CDFHttpClient } from './httpClient/cdfHttpClient';
-import { Logger, LoggerFunc } from './logger';
+import { logger, LoggerFunc } from './logger';
 import {
   createAuthenticateFunction,
   OnAuthenticate,
@@ -92,7 +92,6 @@ export default class BaseCogniteClient {
   private hasBeenLoggedIn: boolean = false;
   private loginApi: LoginAPI;
   private logoutApi: LogoutApi;
-  private logger: Logger;
   /**
    * Create a new SDK client
    *
@@ -129,13 +128,16 @@ export default class BaseCogniteClient {
     this.metadata = new MetadataMap();
     this.loginApi = new LoginAPI(this.httpClient, this.metadataMap);
     this.logoutApi = new LogoutApi(this.httpClient, this.metadataMap);
-    this.logger = new Logger(options.debug);
+    if (options.debug) {
+      logger.attach(this.projectName).enable(this.projectName);
+    }
   }
 
   public authenticate: () => Promise<boolean> = async () => {
     const message = `You can only call authenticate after you have called loginWithOAuth`;
 
-    this.logger.log(message);
+    logger.log(this.projectName, message);
+
     throw Error(message);
   };
 
@@ -239,7 +241,6 @@ export default class BaseCogniteClient {
       httpClient: this.httpClient,
       onAuthenticate,
       onTokens,
-      logger: this.logger,
     });
 
     this.httpClient.set401ResponseHandler(async (_, retry, reject) => {
@@ -365,12 +366,12 @@ export default class BaseCogniteClient {
     this.httpClient.addOneTimeHeader(X_CDF_SDK_HEADER, value);
   }
 
-  public attachLogger(logger: LoggerFunc) {
-    this.logger.attach(logger).enable();
+  public attachLogger(loggerFunc: LoggerFunc) {
+    logger.attach(this.projectName, loggerFunc).enable(this.projectName);
   }
 
   public detachLogger() {
-    this.logger.detach().disable();
+    logger.detach(this.projectName);
   }
 
   protected initAPIs() {
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -16,6 +16,7 @@ import {
   HttpResponse,
 } from './httpClient/basicHttpClient';
 import { CDFHttpClient } from './httpClient/cdfHttpClient';
+import { Logger, LoggerFunc } from './logger';
 import {
   createAuthenticateFunction,
   OnAuthenticate,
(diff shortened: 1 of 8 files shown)


packages/core/src/login.ts
```diff
@@ -239,7 +242,12 @@ function parseTokenQueryParameters(query: string): null | AuthTokens {
     [ERROR_DESCRIPTION_PARAM]: errorDescription,
   } = parse(query);
   if (error !== undefined) {
-    throw Error(`${error}: ${errorDescription}`);
+    throw new CogniteLoginError(`Failed to parse token query parameters`, {
+      accessToken,
+      idToken,
```
valgussev: Can we get `query` here instead or in addition to tokens?

packages/core/src/login.ts
```diff
@@ -81,7 +81,8 @@ export interface AuthenticateParams {
 /** @hidden */
 export async function loginSilently(
   httpClient: CDFHttpClient,
-  params: AuthenticateParams
+  params: AuthenticateParams,
+  onLoginFailed?: (error: CogniteLoginError) => void
```
valgussev: It is currently not exposed externally, right? We need to access `onLoginFailed ` through `authenticate`

mrtraser: Not sure that my understanding is correct, but what actually changed is that you can provide a callback for `loginSilently` method to handle errors that might appear. `loginSilently` works at it was before

packages/core/src/login.ts
mrtraser: yep, I'll add `query` as well

packages/core/src/login.ts
valgussev: we are not calling `loginSilently` directly, instead we call `authenticate` which in turn calls `loginSilently`

polomani: this function is never called by the user. how would the pass this parameter?

mrtraser: my false, it can be redone by providing additional parameter to `authenticate` function

mrtraser: @valgussev , is it enough to have parameter `debug:  boolean` in `loginWithOAuth` with prompting errors to console?

valgussev: we would want to use mixpanel / sentry or any other remote logger, console log is not sufficient, exposing callback is prefered

valgussev (commented):
I'm struggling to find the way for a caller (application) to provide a `LoggerFunc` while calling `authenticate` method, could you please clarify that?

mrtraser:
Yep, you can provide logger to the `sdk` instance. For instance
```typescript
const client = CogniteClient({...});
client.attachLogger(loggerFunction);

//... do stuff

client.authenticate()
```
If authentication failed without throwing error to the application (error handled by sdk), `loggerFunction` will be triggered

packages/core/src/logger.ts
```diff
@@ -0,0 +1,53 @@
...
+  data?: LoggerEventData;
+}
+
+const defaultLogger = (event: LoggerEvent) => console.log(event);
+
+export class Logger {
+  private logger: LoggerFunc;
+  private active: boolean;
+
+  constructor(active = false, logger?: LoggerFunc) {
```
polomani:
```suggestion
  constructor(private active = false, private logger: LoggerFunc = defaultLogger) {
```

polomani: and then can remove the body

packages/core/src/logger.ts
```diff
@@ -0,0 +1,53 @@
...
+    this.logger = logger || defaultLogger;
+    this.active = active;
+  }
+
+  public log(event: string | LoggerEvent) {
+    if (!this.active) {
```
polomani: why not move those 2 lines underneath inside this block instead?

packages/core/src/login.ts
```diff
@@ -110,8 +112,8 @@ export async function loginSilently(
     if (tokens !== null) {
       return tokens;
     }
-  } catch (_) {
-    // don't do anything.
+  } catch (error) {
+    logger.log(error);
```
polomani: maybe we should make logger as a singleton? so we don't need to pass it here at all?
same in constructor.
(it's a suggestion, I don't have strong preferences here, can be this way as well I guess)

packages/core/src/loginError.ts
```diff
@@ -1,8 +1,12 @@
 // Copyright 2020 Cognite AS
 
 export class CogniteLoginError extends Error {
-  constructor(message?: string) {
+  public data: { [key: string]: any };
+  constructor(message?: string, data?: { [key: string]: any }) {
     super(message || 'Not able to login');
+
+    this.data = data || {};
```
polomani: default value should be in the constructor

polomani (commented):
not sure if we should do it as a singleton instead

valgussev:
What if application want to be subscribed only to a specific type of Errors (i.e. auth)? How to differentiate from the rest?

packages/core/src/logger.ts
mrtraser: done

packages/core/src/login.ts
mrtraser: yep, that's a good point, redone as a singleton

packages/core/src/login.ts
polomani: is this safe from a security point of view to throw error exposing the access token?

mrtraser:
@valgussev , we can have prop `type` inside `data` field of triggered error. See `LoggerEventData`. Is it usable? Or better to have it as a part of `LoggerEvent` interface?

packages/core/src/login.ts
mrtraser: we can expose only query instead, which contains access token as well. Access token is also present in header of each request, so don't know should it be a problem or not

packages/core/src/baseCogniteClient.ts
```diff
@@ -126,12 +128,17 @@ export default class BaseCogniteClient {
     this.metadata = new MetadataMap();
     this.loginApi = new LoginAPI(this.httpClient, this.metadataMap);
     this.logoutApi = new LogoutApi(this.httpClient, this.metadataMap);
+    if (options.debug) {
+      logger.attach(this.projectName).enable(this.projectName);
```
valgussev, line 132: Why do we want this per `projectName` and not let's say per sdk client instance?

valgussev (commented):
> @valgussev , we can have prop type inside data field of triggered error. See LoggerEventData. Is it usable? Or better to have it as a part of LoggerEvent interface?

Will it be sdk responsible for differentiating errors and warnings?

packages/core/src/baseCogniteClient.ts
mrtraser: right, but we need to identify sdk instance somehow to have a reference inside singleton. If we have logger instance per sdk, as it was before, we need to pass logger to the methods as a separate parameter, which is no so convenient than have singleton instance instead.

mrtraser:
> Will it be sdk responsible for differentiating errors and warnings?

yes, don't see other options

mrtraser:
Updated PR in part, which is related to adding `type` property to `CogniteLoginError`

polomani (commented):
I am not sure we need to set up the instance of the logger per each sdk instance.

Maybe it's enough to have just a pure singleton?

and use it this way (or similar)

```
import { Debug } from '@cognite/sdk'

...

Debug.activate();
Debug.setLogger(console.log || customLogger);
```

@valgussev thoughs?

valgussev:
I don't see it a the problem on my side to specify it per each sdk instance since we are using 1 instance most of the time

mrtraser:
It will work for all instances at the same time in case of a pure singleton. If it's ok, then I can simplify implementation

andersfylling:
Just to bump the thread: I'm closing this due to age. If anyone wants to continue the work, fix the issues or come with new ideas, please do and re-open this PR / make a new one. But for now, I don't see a reason to keep this open.

PR #423, opened 2020-09-03
feat: add relationships resource api

mrtraser:


packages/beta/package.json
```diff
@@ -18,7 +18,7 @@
     "esCheck": "es-check es5 './dist/index.js'",
     "docs": "typedoc --options typedoc.js --tsconfig tsconfig.json src/index.ts",
     "docs:bundle": "yarn docs && mkdir -p ../../docs/beta && cp -r docs/* ../../docs/beta/",
-    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js",
+    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js --beta",
```
polomani: I think we don't really need to extract beta snippets (because there is no beta docs.cognite.com) yet

unless something has changed in our plans

mrtraser: `test-snippets` is a part of CI. It will be also easier to move features (including code snippets) from `beta` to `stable`

scripts/extractCodeSnippets.js
```diff
@@ -4,6 +4,7 @@ const snippetsFolder = path.join(process.cwd(), './codeSnippets');
 const jsDoc = require(snippetsFolder + '/docs.json');
 const _ = require('lodash');
 const fs = require('fs');
+const isBeta = !!~process.argv.indexOf('--beta');
```
polomani: I would rather have something like `--package-name "@cognite/sdk-beta"`

packages/stable/src/__tests__/api/labels.int.spec.ts
```diff
@@ -1,12 +1,13 @@
 import { ExternalLabelDefinition } from '@cognite/sdk';
+import { randomInt } from '@cognite/sdk-core/src/testUtils';
 import CogniteClient from '../../cogniteClient';
 import { setupLoggedInClient } from '../testUtils';
```
polomani:
```suggestion
import { setupLoggedInClient, randomInt } from '../testUtils';
```

polomani: instead of line above

scripts/extractCodeSnippets.js
polomani: then we could reuse it for other packages

packages/beta/src/types.ts
```diff
@@ -1,5 +1,55 @@
...
+  targetType: RelationshipResourceType;
+  startTime?: Timestamp;
+  endTime?: Timestamp;
+  confidence?: number;
+  dataSetId?: CogniteInternalId;
+  labels?: Label[];
+}
+
+export interface RelationshipsFilterRequest extends FilterQuery {
+  filter: RelationshipsFilter;
```
polomani:
```suggestion
  filter?: RelationshipsFilter;
```

mrtraser: nice catch!

packages/stable/src/__tests__/api/labels.int.spec.ts
mrtraser: done

scripts/extractCodeSnippets.js
mrtraser: done

polomani (approved):
Looks quite easy to add new apis, good sign
(at least so far)

packages/beta/src/__tests__/api/relationships.int.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    ]);
+    [event] = await client.events.create([{ externalId: eventName }]);
+  });
+
+  afterAll(async () => {
+    await client.assets.delete([{ externalId: assetName }]);
+    await client.events.delete([{ externalId: eventName }]);
+  });
+
+  describe('create', () => {
```
hhenriques, line 58: could you please test with a `datasetId` and also with `labels`

PR #424, opened 2020-09-08
chore: initialize the subsurface wells js sdk extension

Sollimann:
Create the subsurface wells SDK extension from the Cognite SDK 

In this PR:

[SBS-2299] - Initialize the wells js sdk and add integration test for cogniteclient connection

[SBS-2299]: https://cognitedata.atlassian.net/browse/SBS-2299

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 * @cognitedata/core-api @cognitedata/devx-externals
+* @cognitedata/subsurface-data
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,2 @@
 * @cognitedata/core-api @cognitedata/devx-externals
-* @cognitedata/subsurface-data
+/packages/wells/ @cognitedata/subsurface-data
--- a/packages/wells/README.md
+++ b/packages/wells/README.md
@@ -34,5 +34,5 @@ cd /cognite-sdk-js/packages/wells
 
 ```bash
 yarn build
-COGNITE_PROJECT=<project-tenant> COGNITE_CREDENTIALS=<your-api-key> yarn test
+COGNITE_WELLS_PROJECT=<project-tenant> COGNITE_WELLS_CREDENTIALS=<your-api-key> yarn test
 ```
--- a/packages/wells/README.md
+++ b/packages/wells/README.md
@@ -1,0 +1,38 @@
+<a href="https://cognite.com/">
+    <img src="./cognite_logo.png" alt="Cognite logo" title="Cognite" align="right" height="80" />
+</a>
+
+# Cognite Wells JS SDK (derived from stable)
+
+This package provides an SDK derived from `@cognite/sdk`, aka
+[stable](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/README.md).
+
+It is recomended to install this package under the same name as `@cognite/sdk`.
+This allows you to change SDK versions without changing your imports.
+See the [beta readme](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/beta/README.md) for details.
+
+### build
+
+```bash
+yarn
+yarn build
+```
+
+### Testing the wells package only
+
+This repo contains some integration tests that require a CDF api key for `subsurface-test` tenant.
+Talk to any of the contributors or leave an issue and it'll get sorted.
+Travis will run the test and has its own api key.
+
+Run all tests:
+
+navigate to wells package root directory:
+
+```bash
+cd /cognite-sdk-js/packages/wells
+```
+
+```bash
+yarn build
+COGNITE_PROJECT=<project-tenant> COGNITE_CREDENTIALS=<your-api-key> yarn test
+```
--- a/packages/wells/package.json
+++ b/packages/wells/package.json
@@ -1,0 +1,31 @@
+{
+  "private": true,
+  "name": "@cognite/sdk-wells",
+  "license": "Apache-2.0",
+  "repository": "cognitedata/cognite-sdk-js",
+  "homepage": "https://github.com/cognitedata/cognite-sdk-js#readme",
+  "main": "dist/index.js",
+  "types": "dist/src/index.d.js",
+  "version": "0.0.0",
+  "scripts": {
+    "clean": "rm -rf dist/ docs/ codeSnippets/",
+    "test": "jest --config=../../jest.config.js --testPathPattern=/wells/",
+    "lint": "eslint 'src/**/*.{js,ts}'",
+    "lint:fix": "yarn lint --fix",
+    "prepublishOnly": "yarn build",
+    "build": "yarn clean && rollup -c && yarn esCheck",
+    "watch": "rollup -cw",
+    "esCheck": "es-check es5 './dist/index.js'",
+    "docs": "typedoc --options typedoc.js --tsconfig tsconfig.json src/index.ts",
+    "docs:bundle": "yarn docs && mkdir -p ../../docs/wells && cp -r docs/* ../../docs/wells/",
+    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js",
+    "test-snippets": "yarn extract-snippets && yarn tsc -p codeSnippets/tsconfig.build.json"
+  },
+  "dependencies": {
+    "@cognite/sdk": "^3.0.0",
+    "@cognite/sdk-core": "^1.0.0"
+  },
+  "publishConfig": {
+    "access": "public"
+  }
+}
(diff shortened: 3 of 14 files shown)


Sollimann:
probably some key here missing for the integration test?

Sollimann:
travis-ci shows this: 
image

so, it fails in the second test here:
image

I would assume it is because the current api-key is probably not authorized for the subsurface-test project?

polomani:
@Sollimann you don't have to have these tests, but yes, you need to configure your project.
From testUtils.ts
```
project: process.env.COGNITE_PROJECT as string,
apiKey: process.env.COGNITE_CREDENTIALS as string,
```
you may rename variable names to something like `COGNITE_PROJECT_WELLS` and `COGNITE_CREDENTIALS_WELLS`
and define them

mrtraser:
But how @Sollimann can define them? Does he have access to the repo variables?

.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
-* @cognitedata/core-api @cognitedata/devx-externals
\ No newline at end of file
+* @cognitedata/core-api @cognitedata/devx-externals
+* @cognitedata/subsurface-data
```
Sollimann: Is this how you wanted me to add subsurface codeowners?

Sollimann:
> @Sollimann you don't have to have these tests, but yes, you need to configure your project.
> From testUtils.ts
> 
> ```
> project: process.env.COGNITE_PROJECT as string,
> apiKey: process.env.COGNITE_CREDENTIALS as string,
> ```
> 
> you may rename variable names to something like `COGNITE_PROJECT_WELLS` and `COGNITE_CREDENTIALS_WELLS`
> and define them

Is this something infra would have to do? add env variables `COGNITE_PROJECT_WELLS` and `COGNITE_CREDENTIALS_WELLS` to the repo?

polomani:
> Is this something infra would have to do? add env variables COGNITE_PROJECT_WELLS and COGNITE_CREDENTIALS_WELLS to the repo?

you should be able to do that yourself in travis configuration https://travis-ci.org/github/cognitedata/cognite-sdk-js/settings

mrtraser:
not everyone has access to travis settings section I believe, including me

.github/CODEOWNERS
polomani:
```suggestion
/packages/wells/ @cognitedata/subsurface-data
```

Sollimann:
@polomani @mrtraser the CI passed. I added credentials in the Travis CI. Could I get a review? :)

polomani (commented):
This pr will not publish anything to npm, so it should be pretty safe to merge it.


Regarding the commit messages:
`feat: added code subsurface-data codeowners` feat keyword usually bumps the major and adds that message as a feature to changelog. You don't need to worry about it now, especially because you are on version 0, where every change is breaking, but I'd rather squash this PR to something like `chore: initial commit` so it is not visible in changelog.

Package name:
Something to reconsider when we publish to npm, I would suggest something more descriptive?

packages/wells/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+navigate to wells package root directory:
+
+```bash
+cd /cognite-sdk-js/packages/wells
+```
+
+```bash
+yarn build
+COGNITE_PROJECT=<project-tenant> COGNITE_CREDENTIALS=<your-api-key> yarn test
```
polomani:
```suggestion
COGNITE_WELLS_PROJECT =<project-tenant> COGNITE_WELLS_CREDENTIALS=<your-api-key> yarn test
```

packages/wells/src/__tests__/cogniteClient.int.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+    client = setupLoggedInClient();
+  });
+  // test that the client behaves as stable
+  test('assets list', async () => {
+    const response = await client.assets.list();
+    expect(response.items.length).toBeGreaterThan(0);
+  });
+  test('raw get assets', async () => {
+    const response = await client.get(
+      '/api/v1/projects/subsurface-test/assets'
```
polomani, line 24: could we make those tests optional for the developers that are not working on wells sdk?
see `packages/stable/src/__tests__/api/revisions3D.int.spec.ts`

We could check if variables are set and run only in that case.
Sot when you run tests for entire project locally and don't set wells variables it'll not fail.

(maybe some different approach, but same idea)

polomani (changes requested):
one small thing remembered

Sollimann:
> This pr will not publish anything to npm, so it should be pretty safe to merge it.
> 
> Regarding the commit messages:
> `feat: added code subsurface-data codeowners` feat keyword usually bumps the major and adds that message as a feature to changelog. You don't need to worry about it now, especially because you are on version 0, where every change is breaking, but I'd rather squash this PR to something like `chore: initial commit` so it is not visible in changelog.
> 
> Package name:
> Something to reconsider when we publish to npm, I would suggest something more descriptive?

reworded the commits from feat to chore.

In terms of naming we have been talking about having the subsurface-* prefix in front of everything. However, in the context of wells the word subsurface would be kind of redundant as wells does not make sense outside the context of subsurface (as far as I know). We just want to be able to run client.wells.assets() request. If you have a suggestion for naming, please let me know :)

packages/wells/src/__tests__/cogniteClient.int.spec.ts
Sollimann: Done :)

packages/wells/src/__tests__/cogniteClient.int.spec.ts
```diff
@@ -2,8 +2,16 @@
 
 import CogniteClient from '../cogniteClient';
 import { setupLoggedInClient } from './testUtils';
+import { isNullOrUndefined } from 'util';
```
polomani: `isNullOrUndefined` is deprecated
```
process.env.COGNITE_WELLS_PROJECT && process.env.COGNITE_WELLS_CREDENTIALS
```
should be enough

Sollimann: changed :+1:

packages/wells/src/cogniteClient.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Create a new SDK client (derived)
+   * @param options Client options
+   */
+  constructor(options: ClientOptions) {
+    super(options);
+  }
+
+  protected get version() {
+    return `${version}-derived`;
```
polomani:
```suggestion
    return `${version}-wells`;
```

Sollimann:
For some reason the first commit having {version}-wells failed in the integration test in the stable package. I changed the version in the packages/wells to have tag {version}-wells-derived. Is that ok?

it might have been a duplicate version that caused it.

polomani:
@Sollimann 

> ${version}-wells failed in the integration test in the stable package

That is not possible. Tests must have failed just because of some random issue, re-run usually helps for that

`${version}-wells-derived` will be sent with every request as a part of `x-cdp-sdk` header, so I would suggest to have something that makes sense there. Maybe `wells/${version}`or even `${stableVersion}/wells/${version}` would be more informative and convenient for gathering metrics? :)

I am still not sure what the package is all about, so cannot suggest much. But we need to be able to differentiate it from other packages in our logs

polomani:
@mrtraser need a stamp from you

mrtraser (approved):

polomani (approved):

packages/wells/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "docs:bundle": "yarn docs && mkdir -p ../../docs/wells && cp -r docs/* ../../docs/wells/",
+    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js",
+    "test-snippets": "yarn extract-snippets && yarn tsc -p codeSnippets/tsconfig.build.json"
+  },
+  "dependencies": {
+    "@cognite/sdk": "^3.0.0",
+    "@cognite/sdk-core": "^1.0.0"
+  },
+  "publishConfig": {
+    "access": "public"
```
polomani, line 29: I didn't notice you are going to publish it already.
Are you ok with the name? Could it be something more descriptive?

Imagine we have multiple sdks, this one will be called `@cognite/sdk-wells` in your case

Btw if you choose to change it, you'd need to change folder name as well. and other occurrences

polomani (changes requested):

packages/wells/package.json
Sollimann: @rek @kiril-me @jucc any input here? Keep the same sdk name?

polomani: also if you want to publish it `"private"` supposed to be set to `false` (line 2)

Sollimann: I think we can keep "private": true and "access": "private" for now. And publish once the SDK has some use to it.

Sollimann: From what I read, it would be enough to set "private": true? 
image

polomani: @Sollimann 

> From what I read, it would be enough to set "private": true?

you might be right, it's ok then.

polomani (approved):
we can change the package name later

PR #427, opened 2020-09-17
fix: fileAPI upload endpoint

mrtraser:
Fixes `upload` endpoint of `FilesAPI` by preventing content-type header changes based on `isJson` check of the request body. 
Previously, `isJson` returns `true` for `ArrayBuffer` request body type and set `content-type: application/json` header instead of provided content-type header in params.
[release]

### DIFF
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -2,7 +2,6 @@
 
 import isObject from 'lodash/isObject';
 import isArray from 'lodash/isArray';
-import { isBuffer } from 'util';
 import { API_VERSION, BASE_URL } from './constants';
 import { CogniteError } from './error';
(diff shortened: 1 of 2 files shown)


hhenriques (approved):

packages/core/src/utils.ts
```diff
@@ -51,7 +50,8 @@ export function convertToTimestampToDateTime(timestamp: number): Date {
 export function isJson(data: any) {
   return (
     (isArray(data) || isObject(data)) &&
-    !isBuffer(data) &&
+    !Buffer.isBuffer(data) &&
```
maksnester, line 53: Passes the test (runs in node) but crashes in a browser because there is no global `Buffer` in browsers.

PR #428, opened 2020-09-21
Wells sdk/fetch wells

Sollimann:
In this PR:
[SBS-2321] - Implement get by id, get by name and list by name prefix 
[SBS-2254] - Implement the getWellsByPolygon method

left some cards in the well model Kanban board to add try-catch and unit tests for wells.

[SBS-2321]: https://cognitedata.atlassian.net/browse/SBS-2321
[SBS-2254]: https://cognitedata.atlassian.net/browse/SBS-2254


packages/wells/src/client/api/wells.ts
```diff
@@ -0,0 +1,178 @@
...
+   * ```js
+   * var wells: Well[] = Wells.mapToWell(asset);
+   * ```
+   * @param assets
+   */
+  static mapToWell = (assets: Asset[]): Well[] => {
+    return assets.map(asset => {
+      return <Well>(<unknown>{
+        name: asset.name,
+        wellHeadLocation: Wells.mapToWellHeadLocation(asset),
```
drsphelps, line 40: the well's id from CDF should also be added to the Well object if it's been fetched from CDF

packages/wells/src/__tests__/fixtures/wellFixtures.ts
```diff
@@ -0,0 +1,88 @@
...
+      // WRP_AZIMUTH_DSDSUNIT: 'lat_dms',
+      // WRP_EW_DSDSUNIT: 'm',
+      // WRP_INCLINATION_DSDSUNIT: 'lat_dms',
+      // WRP_MD_DSDSUNIT: 'm',
+      // WRP_NS_DSDSUNIT: 'm',
+      // WRP_OFFSET: '0.0',
+      WRP_TVD: '67.4',
+      WRP_TVD_DSDSUNIT: 'm',
+      source: 'EDM',
+      type: 'Well',
```
drsphelps, line 11: Should the mock Asset also have fields for x, y, and crs if it's going to be used to create wells?

packages/wells/src/client/api/wells.ts
```diff
@@ -0,0 +1,178 @@
...
+      wkt: polygon,
+      crs: crs,
+    };
+
+    const geometryRelBody = {
+      geometry: geometryBody,
+      relation: SpatialRelationshipNameDTO.Within,
+    };
+
+    const points = await geospatialClient.findSpatial({
```
drsphelps: Does the JavaScript version of the GeospatialSDK limit the number of of returned points to 1000 like the Python SDK? If it does, do we want to limit the number of wells to 1000? Also, some of the points returned may not correspond to wells, so the number of returned wells may be below the limit even if there are enough in CDF

drsphelps (approved):
A couple of minor comments, although for the most part it looks like it implements the same behaviour as the Python SDK

packages/wells/src/client/api/wells.ts
Sollimann: Yea, it is limited by the service contract: [service contract](https://docs.cognite.com/api/playground/#operation/searchSpatial) I added pagination now limited to 1000 pr page

PR #432, opened 2020-10-06
fix: beta re-export bug

mrtraser:
For now, beta package can't redefine `CogniteClient` instance of the sdk stable version in commonJS module builder by Rollup.JS

image

This PR fixes this issue by disabling external live bindings (https://rollupjs.org/guide/en/#outputexternallivebindings)

### DIFF
--- a/packages/beta/rollup.config.js
+++ b/packages/beta/rollup.config.js
@@ -9,6 +9,7 @@ export default {
       dir: 'dist',
       format: 'cjs',
       sourcemap: true,
+      externalLiveBindings: false,
     },
   ],
   external: [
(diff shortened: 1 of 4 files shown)


polomani (approved):
we can try

mrtraser:
Faced with `CORS` issue regarding `version: beta` header:


Screenshot 2020-10-06 at 15 58 35


So, `version: beta` header is not present in `OPTION` response `Access-Control-Allow-Headers` header list (but it should be there). It should be configured on backend side I guess (has been already requested)

**For the reference:**


header in request

```
access-control-request-headers: api-key,content-type,version,x-cdp-app,x-cdp-sdk
```

header in response

```
access-control-allow-headers: accept, accept-language, content-type, content-language, authorization, api-key, x-upload-content-type, x-cdp-sdk, x-cdp-app, x-cdp-clienttag, x-user-agent
```

polomani:
I put it here so we don't forget.
This fix needs to be released together with the "timestamp to date conversion issue in relationships".

so when ppl start using it it's all fixed

hhenriques:
@polomani what "timestamp to date conversion issue in relationships"?

mrtraser:
Added fix for timestamp conventional as well

PR #433, opened 2020-10-14
feat: publish wells sdk

Sollimann:
release the first version (v0.0.1) of the wells sdk


packages/wells/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "private": true,
+  "private": false,
```
Sollimann, line 2: I have set ` "publishConfig": {"access": "public"}` and `"private": false` in the package.json file. Is this the right way to publish the package?

packages/wells/src/client/api/wells.ts
```diff
@@ -143,22 +144,29 @@ export class Wells extends AssetsAPI {
...
+    limit = 1000,
+    offset = 0,
+  }: {
+    polygon: string;
+    source?: string;
+    crs?: string;
+    layerName?: string;
+    limit?: number;
+    offset?: number;
+  }): Promise<Well[]> => {
```
Sollimann, line 161: is this how you wanted it? @rek

Sollimann: Having i.e `geometry = 'POLYGON ((-4.86423 63.59999, 19.86423 63.59999, 19.86423 52.59999, -4.86423 52.59999, -4.86423 63.59999))'`

you can now do like: `client.wells.getWellsByPolygon({polygon: geometry})`

rek: it would be nice if you handled the WKT transformation so we can just send GeoJSON, that way when @kiril-me supports that from `geospatial-sdk-js` no changes are required.

you could also detect a string (wkt) or an object (geojson) and set the query based on that.

also it's still missing a few fields we need. i put up a quick PR here to should how discover would use it: https://github.com/cognitedata/discover/pull/718/files

PR #437, opened 2020-10-19
feat: export dataclasses, types and Wells to SDK

Sollimann:
So, I just recently published the first version of the cognite/sdk-wells to npm [@cognite/sdk-wells](https://www.npmjs.com/package/@cognite/sdk-wells). I then tried to create a small javascript project to import the sdk and run the client. However, there seems to be an issue occurring in the `index.ts` of the wells-sdk when doing the import `const { CogniteClient } = require("@cognite/sdk-wells");` where I get a `exports.CogniteClient = CogniteClient; TypeError: Cannot set property CogniteClient of #<Object> which has only a getter`. This is how the code-snippet looks: 

image

and this is the stack trace when executing the import at **line 2** of `app.js`:

image

I then looked in the `README.MD` for the [derived-sdk](https://github.com/cognitedata/cognite-sdk-js/blob/master/guides/derived_SDKs.md) and noticed that I had forgotten to export the new classes and interfaces, so I have added them in now. I think that might cause some troubles when using the sdk, but is it the root cause of the error @cognitedata/core-api ? Because it seems to me that the root cause is that the `CogniteClient` does not contain a `setter()` as described by:

image

It might be that the lack of exports in the `index.ts` is the root cause, and that the TypeError is the most current exception and that there is a prior exception in the inner exception property that is really where things start failing. What is your thoughts here? FYI this is the CogniteClient extension for the wells package: 
image

### DIFF
--- a/packages/wells/src/index.ts
+++ b/packages/wells/src/index.ts
@@ -2,3 +2,9 @@
 
 export * from '@cognite/sdk';
 export { default as CogniteClient } from './client/cogniteClient';
+export { Wells } from './client/api/wells';
+
+// dataclasses and types
+export * from './client/model/Well';
+export * from './client/model/GeoJson';
+export * from './client/model/WellHeadLocation';
--- a/packages/wells/README.md
+++ b/packages/wells/README.md
@@ -27,29 +27,39 @@ COGNITE_WELLS_PROJECT=<project-tenant>
 COGNITE_WELLS_CREDENTIALS=<your-api-key>
 ```
 
-##### setup client
+#### setup client
 
 ```bash
-let client: CogniteClient = setupLoggedInClient()
+const { CogniteClient } = require("@cognite/sdk-wells");
+
+let client = new CogniteClient({
+  appId: `JS SDK test (${name})`,
+  baseUrl: "https://api.cognitedata.com",
+});
+
+client.loginWithApiKey({
+  project: process.env.COGNITE_WELLS_PROJECT,
+  apiKey: process.env.COGNITE_WELLS_CREDENTIALS,
+});
 ```
 
-##### run a query
+#### run a query
 
 ```bash
-    const polygon = <GeoJson>{
-      type: 'Polygon',
-      coordinates: [
-        [
-          [-4.86423, 63.59999],
-          [19.86423, 63.59999],
-          [19.86423, 52.59999],
-          [-4.86423, 52.59999],
-          [-4.86423, 63.59999],
-        ],
-      ],
-    };
+const polygon = <GeoJson>{
+    type: 'Polygon',
+    coordinates: [
+    [
+        [-4.86423, 63.59999],
+        [19.86423, 63.59999],
+        [19.86423, 52.59999],
+        [-4.86423, 52.59999],
+        [-4.86423, 63.59999],
+    ],
+    ],
+};
 
-    const response = await client.wells.getWellsByPolygon({geometry: polygon});
+const response = await client.wells.getWellsByPolygon({geometry: polygon});
 ```
 
 ![query](figures/wells-sdk.png)
--- a/packages/wells/rollup.config.js
+++ b/packages/wells/rollup.config.js
@@ -9,7 +9,7 @@ export default {
       dir: 'dist',
       format: 'cjs',
       sourcemap: true,
-      externalLiveBindings: true,
+      externalLiveBindings: false,
     },
   ],
   external: [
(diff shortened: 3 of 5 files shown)


packages/wells/src/index.ts
```diff
@@ -2,3 +2,9 @@
 
 export * from '@cognite/sdk';
 export { default as CogniteClient } from './client/cogniteClient';
+export { Wells } from './client/api/wells';
+
+// dataclasses and types
+export * from './client/model/Well';
+export * from './client/model/GeoJson';
+export * from './client/model/WellHeadLocation';
```
Sollimann, line 10: Here are the new exports

polomani:
@mrtraser haven't you fixed a similar issue back then?

polomani:
@Sollimann you need to set `externalLiveBindings=true`, like here https://github.com/cognitedata/cognite-sdk-js/commit/9dbec813197e450c9655d2e6df36e3b4d0cf39b2

Sollimann:
@polomani @mrtraser does my `index.ts` and `rollup.config.js` look good now?

polomani:
looks good to me, yes.
but you can always build locally and test using symlinks

Sollimann:
@polomani So after making the changes setting external live bindings to true, I ran locally inside the wells package `yarn link` and got the message: 
image
So that seems to work, and I can also find the package locally:
image

When I now link to this package locally, I still get the same issue:
image

I checked the the `rollup.config.js` file and the `externalLiveBindings: true` changes are also reflected there: 
image

Any idea why it still not works?

Sollimann:
FYI, I also made sure that it was actually the local package that was linked: 
image

Sollimann:
Any idea here @mrtraser @polomani ? :)

Sollimann:
So, I have now managed to actually import the package and build it locally using symlinks. It turns out that in `rollup.config.js` the **externalLiveBindings** should be set to **false** and not **true**. I am now able to import the library, however, I get a `403 Forbidden` when using any of the methods. The target project and my API key is valid, and it the methods executes as intended both running locally and in the integration tests in the CI/CD pipeline. This is my `App.js` for testing the package using symlinks: 

image

and this is the error message in the failureCallback with the response printing *undefined* : 
image

CogniteClient seems to acknowledge my api-key and project (or I would get a 401), but prevents the method from executing. Experienced this before and know what causes it? @polomani @mrtraser

polomani:
> externalLiveBindings should be set to false and not true

correct! sorry, mistyped my suggestion earlier.

Regarding `setupLoggedInClient`, plz revert that logic back.
Derived packages should have pretty much the same api as the official sdk, this way we would mitigate problems later when something becomes official.
`setupLoggedInClient` is only needed for tests and must not be exposed


403 means that your api key doesn't have capability enabled for this specific resource type.

Sollimann:
Ok, will remove setupLoggedInClient :+1: But how come my API key work running everything locally, and even the integration tests in the CI/CD pipeline, but when I package and export the wells-sdk, I suddently get a 403? @polomani Are you able to extend the access rights for my CDF API key?

polomani:
> but when I package and export the wells-sdk, I suddently get a 403?

you need to check which project you connect to, and what api key you are using.
I suspect it to be different from `COGNITE_WELLS_CREDENTIALS`

Sollimann:
@polomani I have double checked, and I am using the exact same credential, project, base-url and app-id

Sollimann:
@polomani you are right, I do have issues with my access rights. I get the same in Python now. I will sort it out. Thank you for your patience :)

packages/wells/src/client/cogniteClient.ts
```diff
@@ -35,3 +26,19 @@ export default class CogniteClient extends CogniteClientStable {
...
 }
+
+export function setupClient(baseUrl: string = Constants.BASE_URL) {
+  return new CogniteClient({
+    appId: `${name}`,
+    baseUrl,
+  });
+}
+
+export function setupLoggedInClient() {
```
drsphelps: Did this want to be removed, and the imports in the tests reverted back to `./testUtils`?

drsphelps (changes requested):
Looks good, I just think that the `SetupLoggedInClient` needs to be reverted. Correct me if I understood Ivan's comment wrong

packages/wells/src/client/cogniteClient.ts
Sollimann: Yea, I think they want want to create the `CogniteClient` manually. It was only for the test setup

drsphelps (approved):
LGTM 👍

PR #439, opened 2020-10-22
feat: breaking changes in geospatial-sdk [release]

Sollimann:
[SBS-2782] - Update JS SDK after breaking changes in the Geospatial SDK

On another note, Travis seems to hang a lot today. The jobs remain queued for a really long time.

[SBS-2782]: https://cognitedata.atlassian.net/browse/SBS-2782

PR #440, opened 2020-10-23
feat: wells sdk higher order func [release]

Sollimann:
[SBS-2781] - add optional argument to fetch wells by customized functions

Some tenants might structure their data differently, and therefore the current setup for fetching Assets/wells in CDF might not apply for all. This can be solved using higher-order functions that allow their logic to be overridden

[SBS-2781]: https://cognitedata.atlassian.net/browse/SBS-2781

### DIFF
--- a/packages/wells/src/__tests__/integration/wells.int.spec.ts
+++ b/packages/wells/src/__tests__/integration/wells.int.spec.ts
@@ -19,7 +19,7 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
   });
 
   test('standard filter - get well by asset name', async () => {
-    const response = await client.wells.getWellByName('Well A');
+    const response = await client.wells.listByName('Well A');
 
     expect(response.length).toBe(1);
   });
@@ -27,13 +27,13 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
   test('custom filter - get well by asset name', async () => {
     const name = 'Well A';
     const fn: SearchWell = async (args: string) =>
-      await client.wells.getWellByName(args);
-    const response = await client.wells.getWellByName(name, fn);
+      await client.wells.listByName(args);
+    const response = await client.wells.listByName(name, fn);
     expect(response.length).toBe(1);
   });
 
   test('standard filter - get well by asset name prefix', async () => {
-    const response = await client.wells.getWellsByNamePrefix('Well');
+    const response = await client.wells.listByNamePrefix('Well');
 
     response.forEach(function(well) {
       expect(well.name.startsWith('Well')).toBe(true);
@@ -45,9 +45,9 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
     const namePrefix = 'Well';
 
     const fn: SearchWell = async (args: string) =>
-      await client.wells.getWellsByNamePrefix(args);
+      await client.wells.listByNamePrefix(args);
 
-    const response = await client.wells.getWellsByNamePrefix(namePrefix, fn);
+    const response = await client.wells.listByNamePrefix(namePrefix, fn);
     response.forEach(function(well) {
       expect(well.name.startsWith('Well')).toBe(true);
     });
@@ -55,14 +55,14 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
   });
 
   test('standard filter - get empty wells array with name suffix', async () => {
-    const response = await client.wells.getWellsByNamePrefix('A');
+    const response = await client.wells.listByNamePrefix('A');
 
     expect(response.length).toBe(0);
   });
 
   test('standard filter - get well by asset id', async () => {
     const id = 2278618537691581;
-    const response = await client.wells.getWellById(id);
+    const response = await client.wells.getById(id);
 
     expect(response[0]);
     expect(response.length).toBe(1);
@@ -72,9 +72,9 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
     const id = 2278618537691581;
 
     const fn: SearchWell = async (args: number) =>
-      await client.wells.getWellsByIds([{ id: args }]);
+      await client.wells.getByIds([{ id: args }]);
 
-    const response = await client.wells.getWellById(id, fn);
+    const response = await client.wells.getById(id, fn);
 
     expect(response[0]);
     expect(response.length).toBe(1);
@@ -84,7 +84,7 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
     const polygon =
       'POLYGON ((-4.86423 63.59999, 19.86423 63.59999, 19.86423 52.59999, -4.86423 52.59999, -4.86423 63.59999))';
 
-    const response = await client.wells.getWellsByPolygon({
+    const response = await client.wells.searchByPolygon({
       geometry: polygon,
       limit: 1,
       offset: 0,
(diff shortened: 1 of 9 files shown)


packages/wells/src/client/model/Well.ts
```diff
@@ -1,5 +1,8 @@
 import { WellHeadLocation } from './WellHeadLocation';
 
+// Customizable function that takes in CogniteClient and args, and return a promise of a well
+export type SearchWell = (args: any) => Promise<Well[]>;
+
```
Sollimann: All custom methods will take in `any` as argument and have to return a `Promise<Well[]>`

packages/wells/src/client/api/wells.ts
```diff
@@ -202,6 +233,11 @@ export class Wells extends AssetsAPI {
       return { id: x.assetIds[0] };
     });
 
+    // only the calls for CDF are cutomizable, not the once for geoSpatial
+    if (customFilter) {
+      return await customFilter(assetIds);
+    }
+
```
Sollimann: When fetching wells by polygons, you still have to input a polygon as **wkt** or **GeoJson**, so that the geoSpatial method can get retrieve the asset ids. However, the you can customize the way Assets are being called in CDF and translated/wrapped to wells.

Sollimann:
@rek Are these the kinds of filters you want?

packages/wells/src/__tests__/integration/wells.int.spec.ts
```diff
@@ -31,21 +41,46 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
...
+
+    const fn: SearchWell = async (args: string) =>
+      await client.wells.getWellsByNamePrefix(args);
+
+    const response = await client.wells.getWellsByNamePrefix(namePrefix, fn);
+    response.forEach(function(well) {
+      expect(well.name.startsWith('Well')).toBe(true);
+    });
+    expect(response.length).toBe(2);
+  });
```
Sollimann: @rek example of how you can override with  a custom function `fn: SearchWell = (args) => someCustomMethod(args)`

packages/wells/src/client/api/wells.ts
```diff
@@ -81,8 +80,16 @@ export class Wells extends AssetsAPI {
...
+   * @param customFilter a custom filter you can apply, input: any -> output: Promise<Well[]>
    */
-  public getWellByName = async (wellName: string): Promise<Well[]> => {
+  public getWellByName = async (
+    wellName: string,
+    customFilter?: SearchWell
+  ): Promise<Well[]> => {
+    if (customFilter) {
+      return await customFilter(wellName);
+    }
```
drsphelps, line 91: Maybe I'm missing some context here, but why would you want to do this? If you are passing a function and argument to `getWellByName` why not just call the function with the argument yourself?

drsphelps: Would it not be better to get the assets by name first and then apply the custom filter? Like in the `getWellsByPolygon` method

Sollimann: It was requested to have a method/filter that would totally override the logic of the function as long as it returns `Promise<Well[]>`. AkerBP tenants and BP tentants structure their data differently, so the queries will be different. And yes @drsphelps , I agree you might as well just call the function with the argument yourself. But this is the requested functionally as far as I am concerned.

drsphelps: I think we could maybe be a bit more selective about which methods get these custom filters? For example, all the wells no matter how they were ingested (bp vs aker bp) will have the CDF id and external id within the asset, so we don't need a custom filter when getting by id?

jucc: @Sollimann and I had a chat about this earlier today. 

This may not seem not super useful now that we're dealing with wells/wellbores, which are pretty much a given on how to structure. But they are not even the tip of the iceberg. 
When we get to more nuanced datatypes, for a very simple example, well tops (the intersection points of wells with horizons, which are geospatial surfaces) it could be that we model it as, say, as geospatial points in the well data layer, and then some customer decides to model it as events, or as sequences (which is what we currently have in AkerBP, as geospatial was not available in the last hackathon), or even a timeseries if they have time-based data. Then we would want them to be able to override completely, even the type of resource they are looking for, and these custom methods make a lot more sense.

So, for the sake of consistency, if we're going to offer this for the more complex datatypes, we should also offer it to the simpler ones. It doesn't look to me like it's a big deal to add the option to all of the methods, and it reduces the mental overload of trying to decide (and of the user having to check if it is available when they use) which ones have it.
 
Maybe the real issue is if we have too many methods? Should we start by just implementing the ones Discover really uses?

packages/wells/src/client/api/wells.ts
jucc: Shouldn't the user still have the option of providing any method that receives a polygon (wkt or geojson) and returns a list of wells, regardless of how they find it (they may not use geospatial if they don't want to)

In truth, I think this whole method should be encapsulated in a higher level of abstraction.That is:
one method like this: 
```
public getByPolygon = async ({
    geometry,	   
    crs = 'epsg:4326',
    outputCrs = 'EPSG:4326',
    limit = 1000,
    offset = 0,
    customFilter = SearchWell,
  }
```
and, internally, this method would either return the custom filter (like the other ones you wrote) OR do what this current method is doing, calling geospatial, but defining the default arguments to find well heads in geospatial, like ` source = 'wellmodel, layer = 'point',	  attributes = ['geometry']`. 

All of those parameters would be fixed for us, and defined inside of the well data layer. If the user does not want to use geospatial, or wants to use it with their own custom version of these parameters, they provide the custom method.

packages/wells/src/client/model/Well.ts
jucc: I don't know JS, so I'm way out of my comfort zone here and may be saying something very wrong, but maybe we should have one that returns multiple wells (`Promise<Well[]>)` and another one that returns a single well (`Promise<Well>`)?

jucc (commented):
Nit, but wouldn't it make sense to rename the methods in general so that they don't have to be called wells.getWellByName or wells.getWellByPolygon, but instead just wells.getByName or wells.getByPolygon? Since there always be the api prefix, it already explains what it will get :) 

Also, we should follow the convention used in the Cognite SDKs (and general good practice, I guess) to help the user so that methods that return one well would be named `get` and the ones that return a list of wells (like the search by polygon) would be named list, or, in this specific case, maybe search

packages/wells/src/client/api/wells.ts
jucc: Also, @drsphelps , sorry we didn't give you the context before. This is not just related to how data is structured itself, but also to how Discover fetches it. so, the plan was to have this possibility of overriding any of the wells sdk methods (at least the JS one) with another custom method. 
In this way, Discover can always just call the same thing (for example, wells.getByName or, in the future wellbore.listTops(), but have an internal config file that selects which methods to override based on the CDF project/tenant. So BP would have its own list of overriden methods, and other customers could have their own in the future. For any methods that are not overriden, the default well data layer implementation would be called.

hhenriques (approved):
I didn't review the code of the wells package. We should look into making your team it's code owner

jucc (approved):
Great work! We'll probably have to dig into more on how we can combine these methods to have, e.g., one method that lists wells and the polygon, name, and all the others that are incoming as filters in that big method. But we'll get into it once we have the basics sorted out ;)

jucc:
@hhenriques I had only reviewed the wells part last week, because I don't understand the release mechanism and interaction with the rest of the SDK, so that should be a perfect complement! :joy:

Sollimann:
@hhenriques @polomani In my two last PRs ([this one for.ex](https://github.com/cognitedata/cognite-sdk-js/pull/439)) the wells-sdk is no longer automatically bumping the version and releasing the new version of the SDK to NPM. Why is that? I have both **feat** and **[release]** in the PR name/description. As you can see on [github](https://github.com/cognitedata/cognite-sdk-js/releases/tag/%40cognite%2Fsdk-wells%400.1.0) the version is still 0.1.0 like the first time I released the wells-sdk. Same on [npmjs](https://www.npmjs.com/package/@cognite/sdk-wells).

polomani:
@Sollimann if you look into the git history, the last pr on master doesn't have `[release]` keyword, probably got lost when you squashed

PR #441, opened 2020-10-30
feat: add wellbores search to wells sdk [release]

Sollimann:
[SBS-2734] - List method which can return all wellbores from CDF
[SBS-2735] - Extend list method which can return all immediate children
[SBS-2736] - Extend list method to return all children of a given well

also: added more examples to readme and added wellbores to subsurface-test tentant for running integration tests.

[SBS-2734]: https://cognitedata.atlassian.net/browse/SBS-2734
[SBS-2735]: https://cognitedata.atlassian.net/browse/SBS-2735
[SBS-2736]: https://cognitedata.atlassian.net/browse/SBS-2736

### DIFF
--- a/packages/wells/README.md
+++ b/packages/wells/README.md
@@ -22,7 +22,7 @@ yarn install
 yarn build
 ```
 
-### consuming
+## consuming
 
 ##### set your env variables (must be valid for both cdf and geospatial API)
 
@@ -31,14 +31,14 @@ COGNITE_WELLS_PROJECT=<project-tenant>
 COGNITE_WELLS_CREDENTIALS=<your-api-key>
 ```
 
-#### setup client
+### setup client
 
-```bash
-const { CogniteClient } = require("@cognite/sdk-wells");
+```js
+const { CogniteClient } = require('@cognite/sdk-wells');
 
 let client = new CogniteClient({
   appId: `JS SDK test (${name})`,
-  baseUrl: "https://api.cognitedata.com",
+  baseUrl: 'https://api.cognitedata.com',
 });
 
 client.loginWithApiKey({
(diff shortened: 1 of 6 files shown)


packages/wells/README.md
```diff
@@ -43,42 +47,84 @@ client.loginWithApiKey({
...
+const wellId = 2278618537691581;
+const response = await client.wellbores.listChildren(wellId);
 ```
 
-![query](figures/wells-sdk.png)
+#### List all wellbores (all levels of subtree) from wellId
+
+```js
+const rootId = 4438800495523058;
+const response = await client.wellbores.listByWellI(rootId);
```
drsphelps: typo: Id

packages/wells/src/client/api/wellbores.ts
```diff
@@ -0,0 +1,118 @@
+//import { SpatialRel, GeometryRel } from '@cognite/geospatial-sdk-js';
+import { Asset, AssetsAPI } from '@cognite/sdk';
+import { SearchWellbores, Wellbore } from '../model/Wellbore';
+//import { WellHeadLocation } from '../model/WellHeadLocation';
+//import { geospatialClient } from './utils';
+//import { GeoJson } from '../model/GeoJson';
```
drsphelps: Comments can be removed 😄

packages/wells/src/client/api/wellbores.ts
```diff
@@ -0,0 +1,118 @@
...
+  ): Promise<Asset[]> => {
+    const body = {
+      filter: {
+        metadata: {
+          type: 'Wellbore',
+        },
+        ...exactSearch,
+      },
+      search: {
+        ...fuzzySearch,
```
drsphelps: I don't think the fuzzy search is being used in this class? I guess it might be used in the future, but personally I would remove it for now and then `this.search` could then be exchanged for `this.list`.

drsphelps (approved):
Couple of nitpicks, but looks good on the whole

packages/wells/src/client/api/wellbores.ts
Sollimann: Removed the fussySearch part. `this.list` return `CursorAndAsyncIterator<Asset>`, so i'd rather use `this.search` which returns an `Promise<Asset[]>`.

PR #442, opened 2020-11-03
feat: add search and filter method to wells sdk [release]

drsphelps:
[SDL-28] - Fetch and filter wells in an area 

[SDL-28]: https://cognitedata.atlassian.net/browse/SDL-28

### DIFF
--- a/packages/wells/src/client/api/wells.ts
+++ b/packages/wells/src/client/api/wells.ts
@@ -1,6 +1,13 @@
 import { SpatialRel, GeometryRel } from '@cognite/geospatial-sdk-js';
 import { Asset, AssetsAPI, IdEither, Label } from '@cognite/sdk';
-import { Well, SearchWells, SearchWell } from '../model/Well';
+import {
+  Well,
+  SearchWells,
+  SearchWell,
+  WellFilter,
+  WellGeometry,
+} from '../model/Well';
+import { OPERATOR, DATA_SOURCE, FIELD, BLOCK } from '../../constants';
 import { WellHeadLocation } from '../model/WellHeadLocation';
 import { geospatialClient } from './utils';
 import { GeoJson } from '../model/GeoJson';
@@ -40,22 +47,24 @@ export class Wells extends AssetsAPI {
         id: asset.id,
         name: asset.name,
         wellHeadLocation: Wells.mapToWellHeadLocation(asset),
-        dataSource: Wells.getLabel('DataSource', asset.labels),
-        block: Wells.getLabel('Block', asset.labels),
-        field: Wells.getLabel('Field', asset.labels),
-        operator: Wells.getLabel('Operator', asset.labels),
+        dataSource: Wells.getLabel(DATA_SOURCE, asset.labels),
+        block: Wells.getLabel(BLOCK, asset.labels),
+        field: Wells.getLabel(FIELD, asset.labels),
+        operator: Wells.getLabel(OPERATOR, asset.labels),
       };
     });
   };
 
-  static getLabel(type: string, labels: Label[] | undefined): string | void {
+  static getLabel(type: string, labels: Label[] | undefined): string[] {
+    const values = [];
     if (labels) {
       for (const element of labels) {
         if (element.externalId.startsWith(type)) {
-          return element.externalId.split(':')[1];
+          values.push(element.externalId.split(':')[1]);
         }
       }
     }
+    return values;
   }
 
   /**
@@ -251,67 +260,55 @@ export class Wells extends AssetsAPI {
     return await this.getByIds(assetIds);
   };
 
-  public filterPolygonSearch = async ({
-    geometry,
-    crs = 'epsg:4326',
-    outputCrs = 'EPSG:4326',
-    limit = 1000,
-    name,
-    id,
-    data_source,
-    operator,
-    field,
-    block,
+  public listWells = async ({
+    wellGeometry,
+    filter = {},
     customFilter = undefined,
+    limit = 1000,
   }: {
-    geometry: string | GeoJson;
-    crs?: string;
-    outputCrs?: string;
-    limit?: number;
-    name?: string;
-    id?: number;
-    data_source?: string[];
-    operator?: string[];
-    field?: string[];
-    block?: string[];
+    wellGeometry: WellGeometry;
+    filter: WellFilter;
     customFilter?: SearchWells;
+    limit?: number;
   }): Promise<Well[]> => {
     if (customFilter) {
-      return await customFilter(geometry);
+      return await customFilter(wellGeometry.geometry);
     }
 
     let polygonWells = await this.searchByPolygon({
-      geometry,
-      crs,
-      outputCrs,
-      limit,
-      offset: 0,
+      geometry: wellGeometry.geometry,
+      crs: wellGeometry.crs,
+      outputCrs: wellGeometry.outputCrs,
+      limit: limit,
     });
 
-    if (name) {
-      polygonWells = polygonWells.filter(x => x.name === name);
+    if (filter.name) {
+      polygonWells = polygonWells.filter(x => filter.name!.includes(x.name));
     }
-    if (id) {
-      polygonWells = polygonWells.filter(x => x.id === id);
-    }
-    if (data_source) {
+    if (filter.dataSource) {
       polygonWells = polygonWells.filter(
-        x => x.dataSource && data_source.includes(x.dataSource)
+        x =>
+          x.dataSource &&
+          x.dataSource.map(y => filter.dataSource!.includes(y)).includes(true)
       );
     }
-    if (operator) {
+    if (filter.operator) {
       polygonWells = polygonWells.filter(
-        x => x.operator && operator.includes(x.operator)
+        x =>
+          x.operator &&
+          x.operator.map(y => filter.operator!.includes(y)).includes(true)
       );
     }
-    if (field) {
+    if (filter.field) {
       polygonWells = polygonWells.filter(
-        x => x.field && field.includes(x.field)
+        x =>
+          x.field && x.field.map(y => filter.field!.includes(y)).includes(true)
       );
     }
-    if (block) {
+    if (filter.block) {
       polygonWells = polygonWells.filter(
-        x => x.block && block.includes(x.block)
+        x =>
+          x.block && x.block.map(y => filter.block!.includes(y)).includes(true)
       );
     }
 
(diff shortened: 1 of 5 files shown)


packages/wells/src/client/api/wells.ts
```diff
@@ -236,4 +250,71 @@ export class Wells extends AssetsAPI {
 
     return await this.getByIds(assetIds);
   };
+
+  public filterPolygonSearch = async ({
```
rek: can we hide what is happening behind the sdk a bit more, so instead of `filterPolygonSearch` i would do just a `list` and internally the SDK would take care of doing my filters in whatever way is reasonable.

eg here: https://github.com/cognitedata/cognite-sdk-js/blob/master/samples/typescript/quickstart.ts we only `list`, we don't have a `filterByRoots`

jucc: Precisely what II was about to say! So, the idea would be something like wells.list(polygon=[polygon object containing points and crs], blocks=[a list of block names], fields=[a list of field names], ...)

packages/wells/src/client/api/wells.ts
```diff
@@ -236,4 +250,71 @@ export class Wells extends AssetsAPI {
...
+    geometry: string | GeoJson;
+    crs?: string;
+    outputCrs?: string;
+    limit?: number;
+    name?: string;
+    id?: number;
+    data_source?: string[];
+    operator?: string[];
+    field?: string[];
+    block?: string[];
```
rek: how do you propose adding my next filters, for example i need:



- GammaRay
- Caliper
- Resistivity 
- Density
- Neutron
- PPFG
- Geomechanics
- Core
- NDS - (Non Drilling Surprise)
  - Risk Type
  - Severity
  - Probability
- NPT - (Non Productive Time)
  - Duration
  - Detail
  - Code
- KB elevation
- TVD
- Water Depth
- Spud date
- Deviation Cutoff Inclination Angle
- Dog Leg Severity

perhaps a FilterQuery that we can extend from?

eg: https://github.com/cognitedata/cognite-sdk-js/blob/2c0356bccef4f38b0365448f7eb66230ef6c9e53/packages/stable/src/types.ts#L1288

packages/wells/src/client/api/wells.ts
```diff
@@ -40,10 +40,24 @@ export class Wells extends AssetsAPI {
...
+        operator: Wells.getLabel('Operator', asset.labels),
       };
     });
   };
 
+  static getLabel(type: string, labels: Label[] | undefined): string | void {
+    if (labels) {
+      for (const element of labels) {
+        if (element.externalId.startsWith(type)) {
+          return element.externalId.split(':')[1];
```
jucc: I thought I had explained this so poorly, but then you understood it so perfectly! You rock, Dylan!! :rocket:

packages/wells/src/client/api/wells.ts
```diff
@@ -40,10 +40,24 @@ export class Wells extends AssetsAPI {
         id: asset.id,
         name: asset.name,
         wellHeadLocation: Wells.mapToWellHeadLocation(asset),
+        dataSource: Wells.getLabel('DataSource', asset.labels),
```
jucc: Maybe it would be worth having an enum somwhere with the possible values of this before it explodes (at some point we will also have to add the relationship types and maybe some other "constants" that may be better organized somewhere together)

packages/wells/src/client/api/wells.ts
jucc: So, I guess the method already works as expected, it maybe just a matter of renaming to list and creating a "Geometry" class to input to the polygon filter, where you would have polygon string + crs input + crs output, with both crses defaulting to 4326?

packages/wells/src/client/api/wells.ts
```diff
@@ -236,4 +250,71 @@ export class Wells extends AssetsAPI {
...
+    field,
+    block,
+    customFilter = undefined,
+  }: {
+    geometry: string | GeoJson;
+    crs?: string;
+    outputCrs?: string;
+    limit?: number;
+    name?: string;
+    id?: number;
```
jucc: I know we've done that in the Cognuit api, so can't blame, but is it kind of weird to have id in a list filter? It may confuse the user. If they have the id, they could just call a "retrieve" method

packages/wells/src/client/api/wells.ts
drsphelps: So you mean the input to the method would just be a single object that has attributes which contain all the filters you want? I don't think we can use an extension on FilterQuery as we aren't doing the filtering in the call to CDF, but we could create a similar type which contains all the filters you need (which might be what you meant in the first place)

packages/wells/src/client/api/wells.ts
drsphelps: No your explanation was great, made perfect sense!

packages/wells/src/client/api/wells.ts
Sollimann: @drsphelps mabye something like [this](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/constants.ts) ?

packages/wells/src/client/api/wells.ts
jucc: Most of these other filters fall into the "measurement" category, which imho should be treated separately. 

So, while the ones implemented in this PR (and what we plan to treat as simple labels and/or metadata) are reasonably basic string attribute filters that seem like a good choice to live as first-class citizens of the list method, I guess these other ones you mentioned would implement special filters. 

So we planned to have a hasMeasurements(measurements=[some list of measurement names])  - implemented by relationships - and a measurementValues, and, for this second  case, indeed we would need a MeasurementFilter class that has to be specialized or extended by each special class of measurement where we want to filter values from (because some of them will be special little snowflakes that will need some sort of rules to filter)

packages/wells/src/client/api/wells.ts
jucc: Oh, one thing though - I may be misunderstanding the JS here, but it seems you are allowing each well to have only one operator, one field, one of all things (by returning only the first one found)? One of the reasons we went with labels instead of just adding this to metadata was to allow them to have multiple values (in particular, operator is sometimes overriden by a bunch of sub-values of the operator, and dataSource can be multiple sources that contributed to assemblng this well data). 
So you may need to collect all labels that fall into this here and have the attributes in the well being an array of values

packages/wells/src/client/api/wells.ts
drsphelps: The only thing is that every time someone on the BP project decides to add a new label, we would have to come back here and extend the enum? Or do you just mean data source, and that data source is well defined?

packages/wells/src/client/api/wells.ts
jucc: But, again, now I look at it and think this would be the most idiomatic Python way of doing stuff, and I'm out of my league trying to use it in the JS SDK :) - so, if the idiomatic JS way of doing it would be having an interface FilterQuery for allfilters, by all means, let's do it!

packages/wells/src/client/api/wells.ts
```diff
@@ -236,4 +250,71 @@ export class Wells extends AssetsAPI {
...
+      return await customFilter(geometry);
+    }
+
+    let polygonWells = await this.searchByPolygon({
+      geometry,
+      crs,
+      outputCrs,
+      limit,
+      offset: 0,
+    });
```
Sollimann: This is super-duper nitpick, but I think offset is already defaulted to be 0.

Sollimann (commented):
Apart from the things pointed out by Adam and Juliana, I think it looks good. Nice and clean implementation :100: :+1: Could you add some examples of use in the readme, like I have already done with wells and wellbores?

Ping me again when you have edited, and I'll review again

packages/wells/src/client/api/wells.ts
drsphelps: So we can't call the method list as the AssetAPI class we are extending already has a method called list. `listWells` would be choice, but that is a bit of duplication of wells when doing `wells.listWells`. And suggestions?

packages/wells/src/client/api/wells.ts
drsphelps: What should the behaviour be here when filtering? For example, if a well has 2 operators, do both need to be in the filter for the well to be include or just one?

packages/wells/src/client/api/wells.ts
Sollimann: I guess we should avoid having constants/enums that are specific to either AkerBP or BP. I'm thinking just adding regular constants and well-defined terms like i.e 'DataSource'

drsphelps: Ah I see what you mean now

packages/wells/src/client/api/wells.ts
drsphelps: The newest commit should address these concerns @rek I've created a WellFilter interface which can be extended in the future to include all of these attributes, without breaking any existing code, let me know what you think

rek: excellent. was mainly showing that list to point out that fact we needed the interface more generic, and now i see you have done exactly that! 💯 🥇

packages/wells/src/client/api/wells.ts
jucc: So, you would have a filter somewhat like labels=[X1, X2, X3] and would return an "OR" of them. So if the well had 
 labels [X2, X4], it would return true because ther is one in common. 
But now that I think of it, we can in the first version at least suppose only one of each field for simplicity. As long as the query can filter on multiple. Let's leave it for now and ask Javier and the domain people if they feel it would make sense to have both multiple values in the query and multiple values in the well. For the field and block, it doesn't make sense to have more than one in the well. But for some other things that we may want to represent as labels, it could.

drsphelps: In the latest commit the functionality is as you have described it here. Multiple values can be in the well and if one of them is in the query it is included

packages/wells/src/client/api/wells.ts
```diff
@@ -40,10 +47,26 @@ export class Wells extends AssetsAPI {
         id: asset.id,
         name: asset.name,
         wellHeadLocation: Wells.mapToWellHeadLocation(asset),
+        dataSource: Wells.getLabel(DATA_SOURCE, asset.labels),
```
jucc, line 50: Just one more question - does getLabel return an optional? We need to consider the possibhility that people may start with missing data (not having some of the info filled yet). If this returns an optional label and then the attribute can be set to null, should be ok, but just wanted to check ;)

jucc (approved):

PR #443, opened 2020-11-03
feat: added wellbore trajectories [release]

Sollimann:
[SDL-12] - List wellbore trajectories with optional filter on actual in JS SDK

* Trajectories are now expressed as **Survey[]** and are found on the wellbore object using a lazy getter _wellbore.trajectory()_. All trajectory rows are found on the trajectory object (**Survey** dataclass) and can be acessed using a lazy getter _trajectory.rows(limit=1000)_.

* Added endpoints for only listing trajectories. Found by accessing client.surveys.listTrajectories(wellboreId), which filters on DefiniteSurvey and Survey. Should this endpoint only be client.surveys.list(wellboreId)?

 * Ingested wellbore trajectories and populated the rows in subsurface-test to be able to run the ingration tests.

[SDL-12]: https://cognitedata.atlassian.net/browse/SDL-12


packages/wells/src/client/cogniteClient.ts
```diff
@@ -4,17 +4,30 @@ import { version } from '../../package.json';
...
+    // Turns into $BASE_URL/api/$API_VERSION/projects/$PROJECT/sequences
+    this.surveysSDK = this.apiFactory(Surveys, 'sequences');
+
+    /* We want to hide the dependency injection from the users
+     of the SDK. This allows surveys endpoints being accessed on the
+     wellbores object using 'this'. This also ensures
+     a Singleton pattern with only one CogniteClient being 
+     instanciated and we reuse that connection even when
+     surveys endpoints are called within the wellbores class */
+    this.wellboresSDK.surveysSdk = this.surveysSDK;
```
Sollimann, line 30: We want to hide the dependency injection from the users of the SDK. The code snippets below allows surveys endpoints being accessed on the wellbores object using 'this'. This also ensures a Singleton pattern with only one CogniteClient being instanciated and we reuse that connection even when surveys endpoints are called within the wellbores class.

added a public setter and private getter inside the wellbores class: 
image

the lazy getter can now access the surveys endpoints of 'this' with the same client instance like this:
image

I took some time to figure this one out, because I really didn't want to ingest the CogniteClient into each of the wellbores endpoints to be able to use surveys endpoints within the wellbores class.

packages/wells/src/client/api/surveys.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    const surveys = await this.search({
+      filter: {
+        assetIds: [assetId],
+        metadata: { type: 'Survey', PHASE: 'ACTUAL' },
+      },
+    });
+
+    const sequences = surveys.concat(definiteSurveys);
+    return this.mapToSurvey(sequences);
```
Sollimann: It would be really nice if the core JS SDK would allow filtering on several types in a single request. Now I split the queries in two async requests, and then concat them.

rek: you are right. it would be amazing if we could filter metadata by an array! @hhenriques

rek: also, this scales a bit better:
```suggestion
    return this.mapToSurvey([
       ...surveys,
       ...definiteSurveys,
    ]);
```

packages/wells/README.md
```diff
@@ -126,33 +152,57 @@ const rootId = 4438800495523058;
...
-      operator: ['A', 'B'],
-      field: ['A', 'B'],
-      block: ['A', 'B'],
-    },
-    limit: 1000,
-  }
+```js
+const wellId = 2278618537691581;
+const wellbores = await client.wellbores.listChildren(wellId);
+wellbores.forEach(async wellbore => {
```
rek: FYI async forEach is sometimes confusing, do we want to recommend using it?

often when we write code like this we actually meant for a more synchronous for-of type thing

https://advancedweb.hu/how-to-use-async-functions-with-array-foreach-in-javascript/

packages/wells/src/client/model/Survey.ts
```diff
@@ -0,0 +1,42 @@
+import { SequenceRow } from 'stable/dist/src/api/sequences/sequenceRow';
+
+export type SearchSurveys = (args?: any) => Promise<Survey[]>;
+
+export interface Survey {
```
jucc, line 5: Things like assetId or externalId are characteristics of the internal representation (Sequence) and should not be on the Survey data class. 

The survey should instead contain a direct reference to its Wellbore (as in the Wellbore data class). I expect we will add more attributes to the Survey data class once Javeir and Julie finish the mapping of data, but the wellbore itself contains a lot of the info that helps people working with the survey, for example, for visualization. 

So, with a direct reference, we would have something like thisSurvey.wellbore.well.waterDepth. We can add aliases as we proposed between wells and wellbores, so if the user calls, for example,  thisSurvey.getWaterDepth(), we would just return the attribute from thisSurvey.wellbore.well.waterDepth. 
In any case, having the reference to the wellbore directly fetched when we fetch the survey would allow us to do that without having to make another call to CDF to find the asset with that assetId and then do the same repeating CDF calls up through the hierarchy until we find the well

packages/wells/src/client/api/surveys.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+  };
+
+  /**
+   * List all trajectories for a particular assetId (wellbore)
+   * and returns a list of surveys
+   *
+   * @param assetId
+   */
+  listTrajectories = async (
```
jucc, line 35: This looks good! Could we also add a method in the Wellbore class so if the user does wellbore.listSurveys(), we would internally call this method with the wellbore id as a parameter?

I don't know how it's implemented in the JS SDK (I am more familiar with the Python one) but it should be sort of similar to what happens when we call asset.children() in the assets API -  There probably is an internal reference to the CDF client in the asset class (like asset._cdfClient) and then the asset itself has a `children` attribute which is lazy, meaning it is not actually filled until you actually call the `children()` method, then it does asset._cdfClient.list to find other assets that have this asset id as parentId.

Also, from what I understood from Julie's presentation, I believe we can have the trajectory metadata already embedded in the wellbore class, so you could do wellbore.trajectory and get that one trajectory. Then the wellbore.listSurveys() method would list all surveys, regardless of being the real trajectory or planned or anything else

packages/wells/src/client/api/surveys.ts
```diff
@@ -0,0 +1,62 @@
...
+  private mapToSurvey = (sequences: Sequence[]): Survey[] => {
+    return sequences.map(sequence => {
+      return <Survey>{
+        id: sequence.id,
+        name: sequence.name,
+        description: sequence.description,
+        assetId: sequence.assetId,
+        externalId: sequence.externalId,
+        metadata: sequence.metadata,
+        rows: async (limit: number = 100): Promise<SequenceRow[]> => {
```
jucc, line 19: Will this promise be lazy or eager? I.e. when you create a Survey object, will it always fetch all its rows by default, or only when the user tries to access survey.rows?
Since the data can be quite heavy, especially if you have a method that returns several surveys, we probably want this to be lazy, so if the user is only interested in metadata, they don't have to wait for all the data to be loaded

Again, I don't know a lot of JS, but from quickly reading this link - https://tusharsharma.dev/posts/js-promises-eager-not-lazy - it seems you may want to have the rows as a method in the Survey data class instead of an attribute?

packages/wells/src/client/model/Survey.ts
```diff
@@ -0,0 +1,42 @@
...
+   * Custom, application specific metadata. String key -> String value.
+   * @type {{ [key: string]: string; }}
+   * @memberof Survey
+   */
+  metadata?: { [key: string]: string };
+  /**
+   * @type {Promise<SequenceRow[]>}
+   * @memberof Survey
+   */
+  rows(limit?: number): Promise<SequenceRow[]>;
```
jucc, line 31: See comment above (below?) - I believe this should probably be a method instead of an attribute, but I may be wrong due to not knowing the language enough :) 

In this case, ust like mentioned in the other comment about the wellbore, we would need the cdf client as an internal private attribute of the survey object to execute the method when called

packages/wells/src/client/api/surveys.ts
Sollimann: rows is implemented as a method, could have written `async function rows(limit: number = 100) : Promise<SequenceRow>` to get the same. It is a lazy implementation and I also double-checked it when running the tests, so rows() will not be called when the a survey/trajectory is instantiated, it will only be called when you actually run the rows method: `const rows = await trajectories[someIndex].rows(100)`.

Sollimann: same with wellbore.trajectories(), it is also function when the object is instantiated: 
image

packages/wells/src/client/model/Survey.ts
Sollimann: it is declared as a method :+1: : 
image

packages/wells/src/client/api/surveys.ts
Sollimann: Do you want an extra method called `wellbore.listSurveys()` or to rename the current method from `wellbore.listTrajectories()` to `wellbore.listSurveys()`. And should `wellbore.trajectory` list the trajectory that has actually been drilled?

The listTrajectories already calls the for the trajectories internally on the wellbore object based on the wellbore id: 
image

in the case of wellbores, the `asset.id` is the wellbore id and `asset.parentId` is reference to the well Id for that wellbore

packages/wells/src/client/model/Survey.ts
Sollimann: Should we do this in a separate PR mabye?

Sollimann: I can remove assetId and externalId from the Survey data class in this PR

packages/wells/src/client/api/surveys.ts
Sollimann: This is how you would find your wellbores, trajectories and rows based on a wellId: 
image

packages/wells/README.md
```diff
@@ -181,8 +181,8 @@ wellbores.forEach(async wellbore => {
 ```js
 const wellboreId = 4618298167286402;
 const trajectories = await client.surveys.listTrajectories(wellboreId);
-trajectories.forEach(async trajectory => {
-  const rows = await element.rows();
+for (const trajectory of trajectories) {
+  const rows = await trajectory.rows();
   console.log(rows);
 });
```
sighol: Since we are doing async calls, maybe instead of awaiting inside the for loop, maybe it's possible to create a promise for each item and await them all at once? I think there is something called `Promise.all()` that takes a list of futures/promises and waits until the last one is done. If the `trajectory.rows()` function call is going to do any network calls, all network calls will be done sequentially now.

Or maybe it doesn't matter. Could be an idea if we want to improve performance.

sighol (commented):
I don't have enough experience with js or the well-model to approve this, but it looks good :+1:

packages/wells/src/client/api/surveys.ts
jucc: Ok, so the surveys class is the one that actually holds the reference to the client and then you have to have this to call mapToSurvey? So the wellbore has to ive inside this class so it can access the `this` property?

drsphelps (approved):
LGTM 👍  nice work to stop the need for the client instance to be passed to each Survey

PR #444, opened 2020-11-12
feat: retrieve datum and location from parent well in wellbore [release]

drsphelps:
[SDL-15] Return datum object in body of wellbore
[SDL-31] Add well attributes aliases to all wellbores

[SDL-15]: https://cognitedata.atlassian.net/browse/SDL-15
[SDL-31]: https://cognitedata.atlassian.net/browse/SDL-31

### DIFF
--- a/packages/wells/src/client/api/wellbores.ts
+++ b/packages/wells/src/client/api/wellbores.ts
@@ -3,6 +3,9 @@ import { SearchWellbores, Wellbore } from '../model/Wellbore';
 import { Survey } from '../model/Survey';
 import { Surveys } from './surveys';
 import { accessApi } from '@cognite/sdk-core';
+import { Well, WellDatum } from '../model/Well';
+import { Wells } from './wells';
+import { WellHeadLocation } from '../model/WellHeadLocation';
 
 export class Wellbores extends AssetsAPI {
   /**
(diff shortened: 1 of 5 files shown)


packages/wells/src/client/api/wellbores.ts
```diff
@@ -32,18 +35,56 @@ export class Wellbores extends AssetsAPI {
...
+          return well.datum;
+        }
+      };
+      wellbore.getWellHeadLocation = async (): Promise<
+        WellHeadLocation | undefined
+      > => {
+        const well = await wellbore.parentWell();
+        if (well) {
+          return well.wellHeadLocation;
+        }
```
Sollimann: Why is the `getDatum` and `getWellHeadLocation` declaration outside of the `const wellbore = <Wellbore>{ }`? Is it because if implementing inside you would have a lazy implementation, but implementing outside of `const wellbore` doing `wellbore.getDatum ` is an eager implementation? I guess what we want is the eager implementation for `getDatum` and `getWellHeadLocation`.

Sollimann: f.ex `trajectories: async() ...` is a lazy implementation, and will only execute when called like `wellbore.trajectories()`

packages/wells/src/client/api/wellbores.ts
```diff
@@ -32,18 +35,56 @@ export class Wellbores extends AssetsAPI {
...
+   * @param asset
+   */
+  private getParentWell = async (asset: Asset): Promise<Well | undefined> => {
+    if (asset.parentId) {
+      let parentWell = await this.retrieve([{ id: asset.id }]);
+      while (parentWell[0].metadata && parentWell[0].metadata.type !== 'Well') {
+        if (!asset.parentId) {
+          return undefined;
+        }
+        parentWell = await this.retrieve([{ id: asset.parentId }]);
```
jucc: nit, but would it be possible to retrieve the first item of the result here so parentWell is am actual well instead of a list of wells, and then not have to do parentWell[0] in the while and return statements? It seems a bit clearer

jucc (approved):
lgtm!

packages/wells/src/client/api/wellbores.ts
```diff
@@ -32,18 +35,56 @@ export class Wellbores extends AssetsAPI {
...
+      }
+      return Wells.mapToWell(parentWell)[0];
```
Sollimann: So, my understanding is that you search upwards in the asset hierarchy until you find a parent-asset that is type `'Well` and return that well. I guess a wellbore cannot keep a parent reference to another wellbore? so it has to have a reference to parent that is a type 'Well'. If there the user insert an asset that already has type 'well', you should just return that asset as a Well.

If a well cannot keep a parent-reference to another well, I guess that if the argument asset is already type 'Well', you could just check that on the first line and return `undefined `straight away as `if (asset.metadata.type == 'Well') return undefined `

To be able to hit the condition on line 79: `if (!asset.parentId)` the condition on line 76 `if (asset.parentId)` would have to be true. So, I guess the condition on line 79 will always be false?

on line 77, if you want to get the `parentWell`, shouldn't it be: `let parentWell = await this.retrieve([{ id: asset.parentId }]);`? If not, you could just do `let parentWell = asset` 

On line 82, let's say something has gone wrong in the ingestion and the type of the parent asset (parentWell[0].metadata.type = 'Wel') or something else. I guess the program will get stuck in the while-loop infinitely?

Sollimann (changes requested):

packages/wells/src/client/api/wellbores.ts
drsphelps: They are defined outside of the body as they need to be able to reference the wellbore.parentWell which is not possible inside the body as the object has not been created yet. They are still lazy implementations

packages/wells/src/client/api/wellbores.ts
drsphelps: A well can't have another well as a parent, so yeah I think if it has type well I can just return the well, and any other type than wellbore I will just return undefined.

If something has gone wrong in the ingestion then the loop will continue to run past the 'Wel' asset and get to the top of the asset heirarchy where the parentId will be undefined, and therefore the loop will return undefined.

The queries you had about line 76 and 79, and line 77 are typos, which I can fix

Sollimann: with regards to the 'Wel' example, my understanding is that the results from line 82 `parentWell = await this.retrieve([{ id: asset.parentId }]);` will never change, since the id in `asset.parentId` will always be the same. It would run past 'Wel' if it was `parentWell = await this.retrieve([{ id: parentWell.parentId }]);` on line 82?

drsphelps: Oh that's another typo! I guess it's not caught in the tests because the wellbore tested on is a direct descendant of the well

drsphelps: Nice catch!

Sollimann (approved):
Looks good now! :100: :1st_place_medal:

PR #445, opened 2020-11-16
feat(files): add geoLocation property

mrtraser:


packages/stable/src/__tests__/api/files.int.spec.ts
```diff
@@ -98,6 +113,23 @@ describe('Files integration test', () => {
...
+          relation: 'intersects',
+          shape: {
+            type: 'Point',
+            coordinates: [10, 20],
+          },
+        },
+      },
+    });
+
+    expect(aggregates.length).toBe(1);
```
polomani:
```suggestion
    expect(aggregates.length).toBeGreaterThan(0);
```
I believe is better idea

polomani (approved):
Looks good to me, but I am not a codeowner))

hhenriques (approved):

PR #447, opened 2020-11-19
fix: datasets ID scope in datasets ACLs

frxstrem:
The name of the datasets ID scope should be `idScope`, `datasetScope` is not a valid name.

### DIFF
--- a/packages/stable/src/types.ts
+++ b/packages/stable/src/types.ts
@@ -93,6 +93,12 @@ export interface AclScopeAll {
   all: {};
 }
 
+export interface AclScopeIds {
+  idScope: {
+    ids: CogniteInternalId[];
+  };
+}
+
 export type AclScopeAnalytics = AclScopeAll;
 
 export type AclScopeApiKeys = AclScopeAll | AclScopeCurrentUser;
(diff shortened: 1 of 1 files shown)


polomani (changes requested):
`AclScopeDatasetsIds` is used in many places, for resource types ALCs, like here.

```
export type AclScopeEvents = AclScopeAll | AclScopeDatasetsIds;
```
and it is actually valid there. The problem is that it was mistakenly used in this specific case.

so we need to create a separate type for this exception

frxstrem:
@polomani Nice catch, I wasn't aware that it was used for multiple ACLs in this way. I'll fix the PR to work around that.

Though I think that means that this problem is just a symptom of a different problem: that the "scope" values of different ACLs aren't really comparable types (e.g. `DatasetsAcl.Scope` has absolutely no relation to `EventsAcl.Scope` in the actual protobuf schema, even if they both may refer to a set of dataset IDs), so a union of types in this way isn't really the "correct" way to model it.

polomani:
need to fix this as well I believe?
```
export type AclScopeFiles = AclScopeAll | AclScopeDatasets;
```
to
```
export type AclScopeFiles = AclScopeAll | AclScopeDatasetsIds;
```

and squash & rename commit name to be the same format as pr title :)

polomani (approved):

PR #452, opened 2020-11-27
fix: incorrect rotation types for 3d revision

maksnester:
There were few places where I had type errors that came from sdk so I fixed them. 

* consistent Tuple usage 
* `CreateRevision3D` had incorrect rotation type (booleans instead of numbers)
* ~correct type alias for Tuple3(array must have fixed length for Tuple3, just as we have for Tuple2)~ gave up on that one because of failing tests which use empty arrays, IDK if it's intentional so left as it was


polomani (approved):

maksnester:
@mrtraser hi, may I get some approval on that? Doesn't look like approval from Ivan allows it to be merged.

mrtraser (approved):

PR #455, opened 2020-11-30
fix: restructure sdk to use well service api instead of cdf api

Sollimann:
[SDL-144]

I have restructured the project for use with the well service API rather than the CDF API. I have configured a new http-client, deleted the tests and deprecated the methods used with CDF. The new http-client have yet to be tested with any of the endpoints in the well-service, so I'll leave this as a draft for now.

[SDL-144]: https://cognitedata.atlassian.net/browse/SDL-144


Sollimann:
@drsphelps the dataclasses seems right :+1: I think we should branch out from this PR now, so that it doesn't get to big when adding our endpoints. I'll wait with merging this one, cause we're not totally sure the refactoring works before we have actually tested the httpclient for our endpoints. If they work, we can merge this PR, and then merge our branches with endpoint-implementation right after.

drsphelps (approved):
LGTM, and is working with the service!

PR #456, opened 2020-12-07
feat: add method for fetching wellbore trajectory

Sollimann:
[SDL-127] - Fetch trajectory for wellbore in JS SDK

[SDL-127]: https://cognitedata.atlassian.net/browse/SDL-127


packages/wells/src/client/api/wellboresAPI.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.client = httpClient;
+  }
+
+  public set setProject(project: String) {
+    this.project = project;
+  }
+
+  /* eslint-disable */
+  public getTrajectory = async (wellboreId: number): Promise<Survey | undefined> => {
+      const path: string = `/${this.project}/wellbores/${wellboreId}/trajectory`
```
drsphelps: should an error be returned if the project is not defined? I think the API will just return a 404, which wouldn't very helpful?

packages/wells/src/client/api/wellboresAPI.ts
```diff
@@ -0,0 +1,30 @@
...
+      let trajectory: Survey | undefined;
+      await this.client?.get<Survey>(path)
+      .then(response => {
+        if (response.status == 200){
+          trajectory = response.data;
+        } else {
+          trajectory = undefined;
```
drsphelps: here we will be hiding the error message received from the API, perhaps we could have some logging or maybe throw an error to help the user debug?

drsphelps (commented):
A couple of comments, although you may have good reasons for doing it the way you are

packages/wells/src/client/api/wellboresAPI.ts
Sollimann: I guess we can set a 400 Bad Request if the project is not set in the client?

drsphelps: Sounds good!

packages/wells/src/client/api/wellboresAPI.ts
```diff
@@ -0,0 +1,31 @@
...
+
+      if (this.project == undefined){
+        throw new HttpError(400, "The client project has not been set.", {})
+      }
+
+      const path: string = `/${this.project}/wellbores/${wellboreId}/trajectory`
+      
+      return await this.client?.get<Survey>(path)
+      .then(response => response.data)
+      .catch(err => {
```
drsphelps, line 27: This is super clean!

drsphelps (approved):
LGTM!

PR #457, opened 2020-12-08
feat: add endpoint for well filter

drsphelps:


packages/wells/src/__tests__/integration/wells.int.spec.ts
```diff
@@ -15,7 +17,63 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
...
+  });
+
+  test('get by id - 404 if well does not exist', async () => {
+    expect(client).not.toBeUndefined();
+    const wellId: number = 99999999999999;
+
+    await client.wells.getId(wellId)
+      .then(response => response)
+      .catch(err => {
+        console.log(err)
```
Sollimann: this logging can be removed

packages/wells/src/client/api/wellsApi.ts
```diff
@@ -1,17 +1,45 @@
...
+  };
+
+  public getId = async (id: number): Promise<Well | undefined> => {
+    const path: string = `/${this.project}/wells/${id}`
+    return await this.client?.get<Well>(path)
+      .then(response => response.data)
+      .catch(err => {
+        throw new HttpError(err.status, err.errorMessage, {})
+    });
+  };
```
Sollimann: Nit: Should check if project exist
```suggestion
  public getId = async (id: number): Promise<Well | undefined> => {
    if (this.project == undefined){
        throw new HttpError(400, "The client project has not been set.", {})
    }
    const path: string = `/${this.project}/wells/${id}`
    return await this.client?.get<Well>(path)
      .then(response => response.data)
      .catch(err => {
        throw new HttpError(err.status, err.errorMessage, {})
    });
  };
```

packages/wells/src/client/api/wellsApi.ts
```diff
@@ -1,17 +1,45 @@
...
+      .catch(err => {
+        throw new HttpError(err.status, err.errorMessage, {})
+    });
   };
+  /* eslint-enable */
 
-}
+  public filter = async (filter: WellFilter): Promise<Well[] | undefined> => {
+    const path: string = `/${this.project}/wells/list`
+    return await this.client?.post<Well[]>(path, {'data': filter})
```
Sollimann: Check if project exist, if not throw a 400

packages/wells/src/client/api/wellsApi.ts
```diff
@@ -1,17 +1,45 @@
...
 
+  public set setProject(project: String) {
+    this.project = project;
+  }
+
   /* eslint-disable */
   public list = async (): Promise<Well[] | undefined> => {
-    const response = await this.client?.post<Well[]>("/wells")
-    return response?.data
+    const path: string = `/${this.project}/wells`
```
Sollimann: check if project exist here as well

Sollimann (approved):
Added some nit comments, other than that it looks great! :+1: :100:

Sollimann:
I changed the name of the PR. The linter does not allow capital letters in the PR name, the master CI/CD pipeline would have failed. Also I removed the [release] tag, I think we should wait with release a bit more until we are sure it is working and we have added proper documentation with examples in the README.

We should also clean up all the deprecated code before releasing

PR #461, opened 2020-12-10
feat: fetching rows from a survey

Sollimann:
[SDL-176] - Add endpoint for fetching rows from a survey in the JS SDK

[SDL-176]: https://cognitedata.atlassian.net/browse/SDL-176


packages/wells/src/client/api/surveysApi.ts
```diff
@@ -0,0 +1,40 @@
...
+  public set setHttpClient(httpClient: CDFHttpClient) {
+    this.client = httpClient;
+  }
+
+  public set setProject(project: String) {
+    this.project = project;
+  }
+
+  /* eslint-disable */
+  public getData = async (surveyId: number, start?: number, end?: number, limit?: number, cursor?: string, columns?: string[]): Promise<SurveyData | undefined> => {
```
drsphelps: So in the wells endpoints we're making the user pass the filter as an object, whereas here they pass individually, it may be worth deciding on a standard to be used across the endpoints

drsphelps (approved):
Bit of a nit about method signatures, but otherwise LGTM

packages/wells/src/client/api/surveysApi.ts
Sollimann: Yea, I wasn't quite sure what to decide on. Should we pass an object, or as parameters. If we pass as parameters, then it will probably be less code implementation for the once using the SDKs in the sense that they don't have to import data classes and instantiate data classes in their code to be able to use our methods. On the other hand, it might be a lot of pro's in passing data classes instead. What's your thought? Which one should we go for? @drsphelps

PR #465, opened 2020-12-14
feat: add method for fetching wellbore by Id

Sollimann:
[SDL-190] - add method for fetching wellbore by Id in JS SDK

[SDL-190]: https://cognitedata.atlassian.net/browse/SDL-190


packages/wells/src/client/model/MeasurementType.ts
```diff
@@ -1,4 +1,4 @@
-export const enum MeasurementType {
+export enum MeasurementType {
   GammaRay = 'GammaRay',
   Caliper = 'Caliper',
   Resistivity = 'Resistivity',
```
Sollimann, line 4: if we use `export const enum MeasurementType`, we cannot import the enum in a js project. If we use `export enum MeasurementType` we can import in a js project, but we cannot import internally from another module. I solved this by creating an internal `MeasurementType` for the tests, since import failed. Let me know if you have a good workaround.

packages/wells/src/__tests__/integration/wellbores.int.spec.ts
```diff
@@ -3,8 +3,19 @@
...
+enum MeasurementType {
+  GammaRay = 'GammaRay',
+  Caliper = 'Caliper',
+  Resistivity = 'Resistivity',
+  Density = 'Density',
+  Neutron = 'Neutron',
+  PPFG = 'PPFG',
+  Geomechanics = 'Geomechanics',
+  Core = 'Core',
+}
```
Sollimann, line 18: See comment above

packages/wells/src/index.ts
```diff
@@ -1,19 +1,18 @@
...
 export * from './client/model/Wellbore';
 export * from './client/model/WellHead';
 export * from './client/model/WellFilter';
 export * from './client/model/WellType';
 export * from './client/model/Survey';
-export * from './client/model/MeasurementType';
-export * from './client/model/Measurement';
+export { MeasurementType } from './client/model/MeasurementType';
+export { Measurement, Measurements } from './client/model/Measurement';
 export * from './constants';
```
Sollimann, line 17: Javascript does not like when we export using an `*`, so I wrote them out explicitly. Seems to work in typescript though

drsphelps (approved):
LGTM

PR #467, opened 2020-12-15
feat: add geojson support to wells sdk

drsphelps:
[SDL-180] - Add support for GeoJson in the JS SDK

[SDL-180]: https://cognitedata.atlassian.net/browse/SDL-180

### DIFF
--- a/packages/wells/package.json
+++ b/packages/wells/package.json
@@ -9,7 +9,7 @@
   "version": "0.7.0",
   "scripts": {
     "clean": "rm -rf dist/ docs/ codeSnippets/",
-    "test": "jest --config=../../jest.config.js --testPathPattern=/wells/",
+    "test": "jest --config=../../jest.config.js --testPathPattern=/wells/src",
     "lint": "eslint 'src/**/*.{js,ts}'",
     "lint:fix": "yarn lint --fix",
     "prepublishOnly": "yarn build",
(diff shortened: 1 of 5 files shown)


packages/wells/package.json
```diff
@@ -9,7 +9,7 @@
   "version": "0.8.0",
   "scripts": {
     "clean": "rm -rf dist/ docs/ codeSnippets/",
-    "test": "jest --config=../../jest.config.js --testPathPattern=/wells/",
+    "test": "jest --config=../../jest.config.js --testPathPattern=/wells/src",
```
drsphelps, line 12: when I used yalc to install the package locally, it generated its own test files, and they were being run when I was using `yarn test` so have scoped the test path a little closer to stop this

packages/wells/src/client/api/wellsApi.ts
```diff
@@ -37,6 +38,33 @@ export class WellsAPI {
...
+
+  private convertToApiWellFilter(filter: WellFilter): WellFilterAPI {
+    let polygon: PolygonFilterAPI | undefined;
+    if (filter.polygon) {
+      polygon = {
+        geometry: filter.polygon.wktGeometry!, crs: filter.polygon.crs, geometryType: "wkt"
+      }
+    } else {
+      polygon = undefined;
+    }
```
Sollimann: nit: I prefer this, but it's up to you :) 
```suggestion
    let polygon: PolygonFilterAPI | undefined = undefined;
    if (filter.polygon) {
      polygon = {
        geometry: filter.polygon.wktGeometry!, crs: filter.polygon.crs, geometryType: "wkt"
    }
```

Sollimann (approved):
Looks great :100: :+1:  Could you add an example in the README as well? :)

PR #470, opened 2021-01-07
feat: accessor for filtering 3D Asset Mappings [BND3D-676]

larsmoa:
This enables accessing endpoint https://docs.cognite.com/api/v1/#operation/filter3DAssetMappings from the SDK.

### DIFF
--- a/packages/stable/src/api/3d/assetMappings3DApi.ts
+++ b/packages/stable/src/api/3d/assetMappings3DApi.ts
@@ -13,13 +13,6 @@ import {
 
 export class AssetMappings3DAPI extends BaseResourceAPI<AssetMapping3D> {
   /**
-   * @override
-   */
-  protected get searchUrl() {
-    return this.url('filter');
-  }
-
-  /**
    * [List 3D asset mappings](https://doc.cognitedata.com/api/v1/#operation/get3DMappings)
    *
    * ```js
--- a/packages/stable/src/api/3d/assetMappings3DApi.ts
+++ b/packages/stable/src/api/3d/assetMappings3DApi.ts
@@ -8,9 +8,17 @@ import {
   CreateAssetMapping3D,
   CursorResponse,
   DeleteAssetMapping3D,
+  Filter3DAssetMappingsQuery,
 } from '../../types';
 
 export class AssetMappings3DAPI extends BaseResourceAPI<AssetMapping3D> {
+  /**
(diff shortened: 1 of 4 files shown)


polomani:
@larsmoa wouldn't it be more efficient to run tests locally first?

larsmoa:
@polomani: local-testservice broken for me and not something I wanted to dive into. In retrospect; maybe.

haakonflatval-cognite:
@polomani Are there any instructions anywhere as to how to run the tests locally?

packages/stable/src/api/3d/assetMappings3DApi.ts
```diff
@@ -8,9 +8,17 @@ import {
...
   DeleteAssetMapping3D,
+  Filter3DAssetMappingsQuery,
 } from '../../types';
 
 export class AssetMappings3DAPI extends BaseResourceAPI<AssetMapping3D> {
+  /**
+   * @override
+   */
+  protected get searchUrl() {
+    return this.url('filter');
```
polomani: this shouldn't be needed

packages/stable/src/api/3d/assetMappings3DApi.ts
```diff
@@ -30,6 +38,30 @@ export class AssetMappings3DAPI extends BaseResourceAPI<AssetMapping3D> {
...
+    modelId: CogniteInternalId,
+    revisionId: CogniteInternalId,
+    query: Filter3DAssetMappingsQuery
+  ): CursorAndAsyncIterator<AssetMapping3D> => {
+    const path = `${this.encodeUrl(modelId, revisionId)}/list`;
+    return super.listEndpoint(
+      params =>
+        this.post<CursorResponse<AssetMapping3D[]>>(path, { data: params }),
+      query
+    );
```
polomani, line 55: lines 58-62 above should be possible to replace with
```
super.listEndpoint(this.callListEndpointWithPost, query);
```

haakonflatval-cognite: Would that call use the right URL? The URL must contain `modelId` and `revisionId`, which must be communicated somehow.

polomani: yes, you're right. then disregard my previous comment

vegardok (approved):

PR #472, opened 2021-01-12
feat: add method to fetch wellbores for a well [release]

drsphelps:
[SDL-220]

[SDL-220]: https://cognitedata.atlassian.net/browse/SDL-220


Sollimann (approved):
Looks great! :100: Could you add an example to the readme as well?

PR #477, opened 2021-01-26
feat(beta): add Template Groups API

Tapped:


packages/beta/src/api/templateGroups/templateGroupsApi.ts
```diff
@@ -0,0 +1,96 @@
...
+   * [Retrieve Template Groups](https://pr-1202.specs.preview.cogniteapp.com/v1.json.html#operation/postApiV1ProjectsProjectTemplategroupsByids)
+   * <!-- or [similar](https://doc.cognitedata.com/api/v1/#operation/getEventByInternalId) -->
+   *
+   * Retrieve Template Groups by externalIds.
+   *
+   * ```js
+   * const templateGroups = await client.templateGroups.retrieve([{externalId: 'abc'}]);
+   * ```
+   */
+  public retrieve = (ids: ExternalId[], ignoreUnknownIds: Boolean = false) => {
```
polomani:
```suggestion
  public retrieve = (ids: ExternalId[], ignoreUnknownIds: boolean = false) => {
```
and better to wrap this aaditional parameter into the object like it's done in other apis

packages/beta/src/api/templateGroups/templateGroupsApi.ts
```diff
@@ -0,0 +1,96 @@
...
+   *
+   * Delete Template Groups by externalIds.
+   *
+   * ```js
+   * await client.templateGroups.delete([{ externalId: "Wells" }]);
+   * ```
+   */
+  public delete = (ids: ExternalId[]) => {
+    return super.deleteEndpoint(ids, {
+      ignoreUnknownIds: true,
```
polomani: for some reason it's by default only for assets api? need to recheck this, maybe it worth being consistently bad :)

Tapped: Hmm, so I should rather expose that to the user then?
For events we are not exposing it at all, so it'll be inconsistent anyways.

packages/beta/src/api/templateGroups/templateGroupsApi.ts
Tapped: I see thanks for the suggestion

Tapped:
bilde
@mrtraser Do you know why this is happening?

mrtraser:
Guess you shouldn't add a description to typedocs, it breaks code snippets generation.

packages/beta/src/api/templateGroups/templateGraphQlApi.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright 2020 Cognite AS
+
+import { BaseResourceAPI } from '@cognite/sdk-core';
+import { GraphQlResponse } from '../../types';
+
+export class TemplateGraphQlApi extends BaseResourceAPI<any> {
+  /**
+   * [Run a GraphQL query](https://pr-1202.specs.preview.cogniteapp.com/v1.json.html#operation/postApiV1ProjectsProjectTemplategroupsExternalidVersionsVersionGraphql)
+   *
+   * See https://graphql.org/learn/ for more info on how to construct a GraphQL query.
```
mrtraser: Guess such comments breaks code snippets generation

Tapped: Hmm, I'm not sure if I follow. What do you mean breaks code snippets generation?

mrtraser: If you'll look through the other comments to endpoints (`eventsApi.ts` for instance), all of them have typedoc comments explained how to use it (under ```js ... ``` section). All of them have a quite strict structure, cause we generate code-snippets for documentation purposes. Basically, you need to have a link to the API docs followed by js example of how to use the endpoint.

mrtraser: You can take a look at how code snippets generate in the `scripts` folder of the project

Tapped: Wut, so you can't have description together with a title and JS snippet as this comment?

mrtraser: Guess the idea behind this is to have all required information by link. But I also think that you can have it right above link sections without adding empty space between sections. I'll provide an example.

packages/beta/src/api/templateGroups/templateGroupVersionsApi.ts
```diff
@@ -0,0 +1,69 @@
...
+   * [Upsert a Template Group version](https://pr-1202.specs.preview.cogniteapp.com/v1.json.html#operation/postApiV1ProjectsProjectTemplategroupsExternalidVersionsUpsert)
+Create or update a Template Group version.
+   *
+   * A Template Group update supports specifying different conflict modes, which is used when an existing schema already exists.
+   *
+   * Patch -> It diffs the new schema with the old schema and fails if there are breaking changes.
+   * Update -> It sets the new schema as schema of a new version.
+   * Force -> It ignores breaking changes and replaces the old schema with the new schema.
+   *
+   * The default mode is "patch".
```
mrtraser:
```suggestion
```

packages/beta/src/api/templateGroups/templateGroupVersionsApi.ts
```diff
@@ -0,0 +1,69 @@
...
+import {
+  ExternalTemplateGroupVersion,
+  TemplateGroupVersion,
+  TemplateGroupVersionFilterQuery,
+} from '../../types';
+
+export class TemplateGroupVersionsApi extends BaseResourceAPI<
+  TemplateGroupVersion
+> {
+  /**
```
mrtraser, line 13:
```suggestion
  /**
  * A Template Group update supports specifying different conflict modes, which is used when an existing schema already exists.
   * Patch -> It diffs the new schema with the old schema and fails if there are breaking changes.
   * Update -> It sets the new schema as schema of a new version.
   * Force -> It ignores breaking changes and replaces the old schema with the new schema.
   * The default mode is "patch".
```

packages/beta/src/api/templateGroups/templateGraphQlApi.ts
mrtraser: As a workaround, you can specify the same comments in `js` section as well, like in `logoutApi.ts`

Tapped:
@mrtraser This PR seems ready for review, but failed randomly in asset integration test.

Tapped:
@mrtraser Now it's at least green. What do you think about the introduced changes?

mrtraser (approved):
Sorry for the late response, missed that thing that it should be approved after all changes

Tapped:
No problem! Thanks a lot for the review 💯

PR #480, opened 2021-02-01
chore: address breaking changes in well service API [release]

Sollimann:
[SDL-268] - added fuzzy search by name and description
[SDL-246] - Add option to filter on AND and OR in /wells/list in js SDK

[SDL-268]: https://cognitedata.atlassian.net/browse/SDL-268

[SDL-246]: https://cognitedata.atlassian.net/browse/SDL-246

### DIFF
--- a/packages/wells/src/client/model/WellFilter.ts
+++ b/packages/wells/src/client/model/WellFilter.ts
@@ -12,7 +12,7 @@ export interface WellFilter {
    * @type {string}
    * @memberof WellFilter
    */
-  description?: string;
+  stringMatching?: string;
   /**
    * @type {string[]}
    * @memberof WellFilter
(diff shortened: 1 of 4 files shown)


packages/wells/src/client/model/WellFilter.ts
```diff
@@ -12,7 +12,7 @@ export interface WellFilter {
    * @type {string}
    * @memberof WellFilter
    */
-  description?: string;
+  stringMatching?: string;
```
sighol, line 15: Is it obvious to everyone that `stringMatching` is `name` and `description` fuzzy search?

sighol (approved):
LGTM

packages/wells/src/client/model/WellFilter.ts
Sollimann: It was suggested by Juliana we'd use `stringMatching` instead of `query` which is in CDF: 
image

We could change it to something like `matchOnNameOrDescription`, but CDF might expand the search into more fields from what I understand in the description of `query` above.

PR #482, opened 2021-02-08
chore: update wellbore data class [release]

Sollimann:
[SDL-284] - Add `wellId` to the wellbores datatype in the well-service in the JS SDK

[SDL-284]: https://cognitedata.atlassian.net/browse/SDL-284

### DIFF
--- a/packages/wells/src/client/model/Wellbore.ts
+++ b/packages/wells/src/client/model/Wellbore.ts
@@ -33,7 +33,7 @@ export interface Wellbore {
    */
   externalId?: string;
   /**
-   * Parent asset id in CDF
+   * CDF id of the well that contains this wellbore
    * @type {string}}
    * @memberof Wellbore
    */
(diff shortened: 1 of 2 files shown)


sighol (approved):

packages/wells/src/client/model/Wellbore.ts
```diff
@@ -15,21 +15,29 @@ export type SearchWellbore = (args?: any) => Promise<Wellbore>;
...
+   * External Id for the wellbore
+   * @type {string}}
    * @memberof Wellbore
    */
-  name: string;
+  externalId?: string;
   /**
-   * Custom, application specific metadata. String key -> String value.
-   * @type {{ [key: string]: string; }}
+   * Parent asset id in CDF
```
jucc: The well is not necessarily the parent of the wellbore, it could be higher up above the hierarchy. This doc string should probably state something more like "CDF id of the well that contains this wellbore"

PR #486, opened 2021-02-15
fix: setting a field resolver to undefined [release]

Tapped:
We should not auto wrap values that are undefined since a Constant Resolver with undefined value is not supported by the API. Instead we should just ignore undefined field resolvers.

### DIFF
--- a/packages/beta/src/api/templateGroups/templateInstancesApi.ts
+++ b/packages/beta/src/api/templateGroups/templateInstancesApi.ts
@@ -138,7 +138,10 @@ class TemplateInstanceCodec {
   ): { [K in string]: FieldResolver } {
     const mappedResolvers = toPairs(fieldResolvers).map(
       ([name, fieldResolver]) => {
-        if (!this.isFieldResolver(fieldResolver)) {
+        if (
+          fieldResolver !== undefined &&
+          !this.isFieldResolver(fieldResolver)
+        ) {
           // Auto-wrap objects, numbers etc. as ConstantResolver.
           return [name, new ConstantResolver(fieldResolver)];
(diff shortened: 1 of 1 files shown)


packages/beta/src/api/templateGroups/templateInstancesApi.ts
```diff
@@ -138,7 +138,9 @@ class TemplateInstanceCodec {
   ): { [K in string]: FieldResolver } {
     const mappedResolvers = toPairs(fieldResolvers).map(
       ([name, fieldResolver]) => {
-        if (!this.isFieldResolver(fieldResolver)) {
+        if (fieldResolver === undefined) {
+          return [name, undefined];
+        } else if (!this.isFieldResolver(fieldResolver)) {
```
mrtraser:
```suggestion
        if (fieldResolver !==  undefined && !this.isFieldResolver(fieldResolver)) {
```

packages/beta/src/api/templateGroups/templateInstancesApi.ts
```diff
@@ -138,7 +138,9 @@ class TemplateInstanceCodec {
   ): { [K in string]: FieldResolver } {
     const mappedResolvers = toPairs(fieldResolvers).map(
       ([name, fieldResolver]) => {
-        if (!this.isFieldResolver(fieldResolver)) {
+        if (fieldResolver === undefined) {
+          return [name, undefined];
```
mrtraser:
```suggestion
```

mrtraser (changes requested):

peternic (approved):

PR #487, opened 2021-02-16
feat(aad-auth): add ability to login using AzureAD auth flow

mrtraser:
This PR introduces Azure AD auth flow support (aka native tokens). PR is a draft status, things to be added:
- [x] test coverage of introduced auth flow
- [x] sample how to use new auth flow followed by the process description itself
- [x] add documentation for Azure AD OAuth flow


packages/core/src/aad.ts
```diff
@@ -0,0 +1,253 @@
...
+  }
+
+  /**
+   * Calls getAllAccounts and determines the correct account to sign into,
+   * currently defaults to first account found in cache.
+   *
+   * https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/docs/Accounts.md
+   */
+  getAccount(): AccountInfo | null {
+    const currentAccounts = this.msalApplication.getAllAccounts();
```
vegardok: this won't work, the SDK have to store which user was chosen and only use that

packages/core/src/aad.ts
```diff
@@ -0,0 +1,253 @@
...
+    const msalDefaultConfig = AzureAD.getDefaultMSALConfig(debug);
+    this.msalApplication = new PublicClientApplication({
+      ...msalDefaultConfig,
+      ...config,
+    });
+    this.setCluster(cluster);
+  }
+
+  setCluster(clusterName: string): void {
+    this.cluster = clusterName;
```
vegardok: if this changes it'll invalidate the token

packages/core/src/aad.ts
```diff
@@ -0,0 +1,253 @@
...
+      return currentAccounts[0];
+    }
+
+    return null;
+  }
+
+  /**
+   * Gets the token to read user profile data silently, or falls back to interactive redirect.
+   */
+  async getProfileTokenRedirect(): Promise<void | AuthenticationResult> {
```
vegardok: This is useless, you either want to get a token silently and just fail if it didn't work, or explicitly got for an interactive login

packages/core/src/aad.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    await this.msalApplication.logout(logOutRequest);
+
+    this.account = undefined;
+  }
+
+  /**
+   * Returns already acquired access token
+   */
+  async getCDFToken(): Promise<string | null> {
```
vegardok, line 142: Can you add `getAccountToken` as well? It'll be needed to access user details through microsofts graph API

packages/core/src/aad.ts
```diff
@@ -0,0 +1,253 @@
...
+      this.account = account;
+
+      return account;
+    }
+  }
+
+  private async handleAuthRedirect(): Promise<AccountInfo | null> {
+    let redirectResult;
+    try {
+      redirectResult = await this.msalApplication.handleRedirectPromise();
```
vegardok, line 222: if the result contains `account` this is the chosen account. this have to be persisted to LS or similar and used in `getAccount` to get the right token

packages/core/src/baseCogniteClient.ts
```diff
@@ -61,6 +71,18 @@ export interface OAuthLoginOptions extends Project {
   accessToken?: string;
 }
 
+export interface OAuthLoginForAADOptions {
+  cluster: string;
+  clientId: string;
+  tenantId: string;
```
vegardok: This is optional. If it's not defined you should use the `/common` multi tenant endpoint

https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-convert-app-to-be-multi-tenant

packages/core/src/baseCogniteClient.ts
```diff
@@ -188,12 +215,21 @@ export default class BaseCogniteClient {
...
+   *
+   * // or you can login using AzureAD authentication flow (in case your projects supports it)
+   * client.loginWithOAuth({
+   *   cluster: '[CLUSTER]',
+   *   clientId: '[CLIENT_ID]', // client id of your AzureAD application
+   *   tenantId: '[TENANT_ID]', // tenant id of your AzureAD tenant
+   * });
    * // after login you can do calls with the client
    * (async () => {
+   *   await client.authenticate();
```
vegardok, line 236: If you use AAD you'll have to set the project after authenticating

packages/core/src/baseCogniteClient.ts
```diff
@@ -265,6 +294,23 @@ export default class BaseCogniteClient {
...
+  /**
+   * Returns access token in case of AzureAD authentication flow usage
+   *
+   * ```js
+   * client.loginWithOAuth({cluster: 'bluefield', ...});
+   * await client.authenticate();
+   * const accessToken = await client.getAzureADAccessToken();
+   * ```
+   */
+  public getAzureADAccessToken(): Promise<string | null> {
```
vegardok, line 342: I would call this the CDF token, not access token. Can you add the actual access token as well? The one that allows calling the MS graph API

packages/core/src/baseCogniteClient.ts
```diff
@@ -401,6 +447,70 @@ export default class BaseCogniteClient {
...
+    clientId,
+    tenantId,
+    signInType,
+    debug,
+  }: OAuthLoginForAADOptions) => {
+    const config = {
+      auth: {
+        clientId,
+        authority: `https://login.microsoftonline.com/${tenantId}`,
+        redirectUri: `${window.location.origin}`,
```
vegardok, line 500: add `navigateToLoginRequestUrl: true` as well. it enabled restoring the url state after redirecting back to the app

packages/core/src/baseCogniteClient.ts
```diff
@@ -401,6 +447,70 @@ export default class BaseCogniteClient {
...
+      },
+    };
+
+    const azureAdClient = new AzureAD({ config, cluster, debug });
+
+    this.azureAdClient = azureAdClient;
+
+    return async () => {
+      const account = await azureAdClient.initAuth();
+      const authResult = await azureAdClient.getProfileTokenRedirect();
```
vegardok: This doesn't look right, you are getting a token via redirect and doing a login (new redirect) below?

packages/core/src/aad.ts
```diff
@@ -0,0 +1,253 @@
...
+      scopes: this.userScopes,
+      extraScopesToConsent: this.getCDFScopes(),
+    };
+  }
+
+  private get loginRedirectRequest(): RedirectRequest {
+    return {
+      scopes: this.userScopes,
+      extraScopesToConsent: this.getCDFScopes(),
+      redirectStartPage: window.location.origin,
```
vegardok: Set this to `window.location.href`. This won't do anything without setting `navigateToLoginRequestUrl` to true in auth options.

packages/core/src/aad.ts
mrtraser: I got your point, make sense to me, going to change

packages/core/src/aad.ts
mrtraser: That's true, with the next attempt to get a token or request api you'll pass through the `authentication()` step to get a valid token first, which you might expect to get when changing cluster. In case you find this redundant, we can remove it until somebody requests it. Then we might ask about expected flow in case of changing cluster

packages/core/src/baseCogniteClient.ts
mrtraser: yeah, make sense, completely redundant here

packages/core/src/aad.ts
```diff
@@ -0,0 +1,244 @@
...
+    });
+    this.cluster = cluster;
+  }
+
+  getCluster(): string {
+    return this.cluster;
+  }
+
+  /**
+   * Calls getAllAccounts and determines the correct account to sign into,
```
vegardok: comment out of date

packages/core/src/aad.ts
```diff
@@ -233,21 +230,15 @@ export class AzureAD {
...
-      if (e instanceof InteractionRequiredAuthError) {
-        return this.msalApplication
-          .acquireTokenRedirect(interactiveRequest)
-          .catch(noop);
-      }
+  private setLocalAccountIdToLocalStorage(localId?: string): void {
+    if (!localId) {
+      localStorage.removeItem(accountLocalStorageKey);
+    } else {
+      localStorage.setItem(accountLocalStorageKey, localId);
```
polomani, line 236: sdk used to work in nodejs previously (not the authentication part, but it worked if token is provided from outside),
do we want to make aad-auth work in nodejs as well? or is it out of the current scope?

mrtraser: you still can provide `accessToken` via `loginWithOAuth`, as it was before. The only question here that you can't refresh this token in any way in the node environment

packages/core/src/baseCogniteClient.ts
```diff
@@ -265,6 +295,34 @@ export default class BaseCogniteClient {
...
+   * await client.authenticate();
+   * const accessToken = await client.getCDFToken();
+   * ```
+   */
+  public getCDFToken(): Promise<string | null> {
+    if (!this.azureAdClient) {
+      throw Error('CDF token can be acquired only using AzureAD auth flow');
+    }
+
+    return this.azureAdClient.getCDFToken();
```
vegardok, line 329: is it possible to expose the token access via legacy auth as well? this would be handy for apps that roll their own http client or APIs but use the same auth token.

packages/core/src/baseCogniteClient.ts
```diff
@@ -23,12 +24,19 @@ import {
   OnTokens,
 } from './login';
 import { MetadataMap } from './metadata';
-import { getBaseUrl, isUsingSSL, projectUrl } from './utils';
+import {
+  getBaseUrl,
+  isOAuthWithAADOptions,
```
vegardok, line 29: Can you expose which flow is in use to the outside? It would be convenient to expose this state our of the SDK since some APIs work with legacy auth and not native tokens, and vica verca.

vegardok (commented):
A couple of more comments but all in all it looks really good. I want to check out the branch and set it up in fusion, but my schedule is pretty packed so I haven't gotten the chance yet

packages/core/src/baseCogniteClient.ts
mrtraser: Yes, but the old one can be outdated, cause we don't have refresh logic around it. We can have it as a separate improvement if requested.

packages/core/src/baseCogniteClient.ts
mrtraser: yep, definitely make sense

mrtraser: `getOAuthFlowType` has been added

vegardok (approved):

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
-## More
+### More
 
 Here is an article describing more in detail about the authentication process:
 https://doc.cognitedata.com/dev/guides/iam/external-application.html#tokens
+
+## Azure AD auth flow
+
+Azure AD auth flow can be used via the same `loginWithOAuth` method of `CogniteClient`,
+but different `options` structure has to be passed. You'll find what options structure has to be provided in examples below.
```
polomani: some people might be not familiar with cdf auth flow anyway, I don't think there is a point of saying this

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+Besides, here list of pre-requirements, that need to be in place to use Azure AD authentication flow:
+
+ - CDF project, which you're trying to log in, has to be configured to accept tokens issued by Azure AD (to be related to the cluster, that accepts Azure AD tokens, `bluefield` for instance)
+ - You should have the `clientId` of your Azure AD application token will be issued for
+ - Cluster your CDF project related to (`bluefield` for instance)
```
polomani: maybe make a section "requirements"?

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
 Here is an article describing more in detail about the authentication process:
 https://doc.cognitedata.com/dev/guides/iam/external-application.html#tokens
+
+## Azure AD auth flow
+
+Azure AD auth flow can be used via the same `loginWithOAuth` method of `CogniteClient`,
+but different `options` structure has to be passed. You'll find what options structure has to be provided in examples below.
+Besides, here list of pre-requirements, that need to be in place to use Azure AD authentication flow:
+
+ - CDF project, which you're trying to log in, has to be configured to accept tokens issued by Azure AD (to be related to the cluster, that accepts Azure AD tokens, `bluefield` for instance)
```
polomani: I believe default cluster also accepts them. Only difference is that you have to ask auth team to enable it for the project

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+ - Cluster your CDF project related to (`bluefield` for instance)
+
+Azure AD authentication flow supports two login types:
+
+ - [With redirect](#authentication-via-redirect)
+ - [With popup](#authentication-via-popup)
+
+### Authentication via redirect
+
+When you authenticate using redirect, auth flow remains quite similar that we have in CDF auth flow.
```
polomani: again, user might be not familiar with the cdf flow, better to duplicate the explanation

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+But the IdP in this is Azure where user has to log in to prove identity.
+
+> You might find useful example application using redirect Azure AD auth flow [here](../samples/react/authentication-aad/src/App.js).
+> And don't forget to provde required environment variables via `.env` file.
```
polomani:
```suggestion
> And don't forget to provide required environment variables via `.env` file.
```

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+// set project name to make requsts to
+client.setProject('cdf-project-name');
+
+// then use the SDK:
+const assets = await client.assets.retrive({ id: 23232789217132 });
+```
+
+After `client.authenticate()` call application waits for the message from the just popped up window about
+successful or unsuccessful log in inside it. In case of `client.authenticate()` returns `true`
+you good to set cdf project name and make requests.
```
polomani:
```suggestion
you good to set CDF project name and make requests.
```

guides/authentication.md
vegardok: That's correct, the interesting bit would be that some clusters _only_ support this flow (all new ones), otherwise it depends on the projects you have access to in the cluster

guides/authentication.md
```diff
@@ -48,14 +48,21 @@ The access token will expire after some time and result in the user get `401` fr
 
 ## How to authenticate with the SDK?
 
-There are two ways to authenticate using the SDK:
+SDK supports few authentication flows:
```
Anette-Cognite: The SDK supports the following authentication flows:

guides/authentication.md
```diff
@@ -48,14 +48,21 @@ The access token will expire after some time and result in the user get `401` fr
...
+
+There are two ways to authenticate using the CDF auth flow:
 
 - [With redirect](#authentication-with-redirects)
 - [With popups](#authentication-with-popups)
 
 ### Authentication with redirects
 
-When doing authentication with redirects the currect browser window will be redirected to the identity provider for the user to log in. After login the same browser window will be redirect back to your application.
+When doing authentication with redirects the current browser window will be redirected to the identity provider for the user to log in.
```
Anette-Cognite: When doing authentication with redirects, the browser window will be redirected to the identity provider for the user to sign in.

guides/authentication.md
```diff
@@ -48,14 +48,21 @@ The access token will expire after some time and result in the user get `401` fr
...
+After the login, the same browser window will be redirected back to your application.
```
Anette-Cognite: After signing in, the browser window will be redirected back to your application.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
 });
 ```
 
-## More
+### More
 
 Here is an article describing more in detail about the authentication process:
```
Anette-Cognite: Read more about the authentication process:

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
 https://doc.cognitedata.com/dev/guides/iam/external-application.html#tokens
+
+## Azure AD auth flow
+
+Azure AD auth flow can be used via the same `loginWithOAuth` method of `CogniteClient`,
+but different `options` structure has to be passed. You'll find what options structure has to be provided in examples below.
+Besides, here list of pre-requirements, that need to be in place to use Azure AD authentication flow:
```
Anette-Cognite: The following prerequisites must be in place to use Azure AD authentication flow:

guides/authentication.md
Anette-Cognite: The CDF project you're trying to sign in to must be configured to accept tokens issued by Azure AD.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+
+ - CDF project, which you're trying to log in, has to be configured to accept tokens issued by Azure AD (to be related to the cluster, that accepts Azure AD tokens, `bluefield` for instance)
+ - You should have the `clientId` of your Azure AD application token will be issued for
```
Anette-Cognite: Rewrite? Not sure I understand these 2 sentences?

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+ - Cluster your CDF project related to (`bluefield` for instance)
+
+Azure AD authentication flow supports two login types:
```
Anette-Cognite: Azure AD authentication flow supports these sign in methods:

guides/authentication.md
Anette-Cognite: Agree. Duplicate.

guides/authentication.md
Anette-Cognite: Remember to provide the required environment variables in the `.env` file.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+
+ - [With redirect](#authentication-via-redirect)
+ - [With popup](#authentication-via-popup)
+
+### Authentication via redirect
+
+When you authenticate using redirect, auth flow remains quite similar that we have in CDF auth flow.
+But the IdP in this is Azure where user has to log in to prove identity.
```
Anette-Cognite: How about:  The IdP in this example is Azure where the user has to sign in to prove the identity.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+await client.authenticate();
+
+// set project name to make requsts to
+client.setProject('cdf-project-name');
+
+// then use the SDK:
+const assets = await client.assets.retrive({ id: 23232789217132 });
+```
+
+With the call `await client.authenticate()` you'll be redirected to the IdP for log in.
```
Anette-Cognite: With the call `await client.authenticate()` you'll be redirected to the IdP to sign in.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+After a successful login, you'll be redirected back and `await client.authenticate()` call
+will return you `true` as a result of the successful login. An important thing here is
```
Anette-Cognite: will return you `true` as a result. It is important

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+to set project for the `CogniteClient` instance via `client.setProject('project-name')` 
+
+### Authentication via popup
```
Anette-Cognite: pop-up

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+
+Popup can be also used to provide user ability to log in Azure.
```
Anette-Cognite: You can also provide a pop-up window for the user to sign in to Azure.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+await client.authenticate();
+
+// set project name to make requsts to
+client.setProject('cdf-project-name');
+
+// then use the SDK:
+const assets = await client.assets.retrive({ id: 23232789217132 });
+```
+
+After `client.authenticate()` call application waits for the message from the just popped up window about
```
Anette-Cognite: After `client.authenticate()` call, the application waits for the message from the pop-up window about

guides/authentication.md
mrtraser: @Anette-Cognite, can we use `popup` to be consistent with the rest text?

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+successful or unsuccessful log in inside it. In case of `client.authenticate()` returns `true`
```
Anette-Cognite: successful or unsuccessful sig in. In case of `client.authenticate()` returns `true`

Anette-Cognite:
Hi
We try to follow Microsoft:
https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/p/pop-up

We use pop-up window (or menu) as a noun. And also sing in (not log in)

*Anette K. Jacobsen*
Technical Writer
+47 926 02 784 |anette.jacobsen@cognite.com
www.cognite.com | REVEALING INDUSTRIAL REALITY™


See the latest news <https://www.cognite.com/newsroom> from Cognite.
Subscribe  <https://www.cognite.com/newslettersignups>to our newsletter.

guides/authentication.md
mrtraser: Rewritten

guides/authentication.md
mrtraser: This section has been rewritten

guides/authentication.md
Anette-Cognite: you can set the CDF project name and make requests.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+you good to set cdf project name and make requests.
+
+### Get cdf token
+
+After user has been successfully authenticated, CDF token can be acquired via `await client.getCDFToken()`.
```
Anette-Cognite: After the user has been authenticated, you can acquire the CDF token via `await client.getCDFToken()`.

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+
+// set project name to make requsts to
+client.setProject('cdf-project-name');
+
+// then use the SDK:
+const assets = await client.assets.retrive({ id: 23232789217132 });
+```
+
+With the call `await client.authenticate()` you'll be redirected to the IdP for log in.
+After a successful login, you'll be redirected back and `await client.authenticate()` call
```
Anette-Cognite: After you have signed in, you'll be redirected back and `await client.authenticate()` call

guides/authentication.md
```diff
@@ -212,7 +219,113 @@ client.loginWithOAuth({
...
+```
+
+After `client.authenticate()` call application waits for the message from the just popped up window about
+successful or unsuccessful log in inside it. In case of `client.authenticate()` returns `true`
+you good to set cdf project name and make requests.
+
+### Get cdf token
+
+After user has been successfully authenticated, CDF token can be acquired via `await client.getCDFToken()`.
+This method works only for Azure AD authentication flow. Here is also ability to check which flow has been used with `loginWithOAuth`:
```
Anette-Cognite: You can also check which flow has been used with `loginWithOAuth`:

packages/core/src/baseCogniteClient.ts
```diff
@@ -188,12 +219,22 @@ export default class BaseCogniteClient {
    *
    * const client = new CogniteClient({ appId: '[YOUR APP NAME]' });
    *
+   * // using Cognite authentication flow
    * client.loginWithOAuth({
    *   project: '[PROJECT]',
    *   onAuthenticate: REDIRECT // optional, REDIRECT is by default
    * });
+   *
+   * // or you can login using AzureAD authentication flow (in case your projects supports it)
```
Anette-Cognite: sign in

mrtraser:
@Anette-Cognite, thank you for a super-wide review of the authentication guide part!

Anette-Cognite (approved):
LGTM

PR #488, opened 2021-02-16
fix(typo): in authentication documentation [release]

tgib23:
Just documentation typo fix

### DIFF
--- a/guides/authentication.md
+++ b/guides/authentication.md
@@ -72,7 +72,7 @@ client.loginWithOAuth({
 });
 
 // then use the SDK:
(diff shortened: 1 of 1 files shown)


ultrathlete (approved):
Thanks for the fix! This won't be released until you put `[release]` in the PR headline.

ultrathlete:
so if you want to release it now, you should do that otherwise it will be released on the next release, whenever that will be.

tgib23:
@me-ydv-5 thanks for your notice. even though i'm not in a hurry at all, just fixed the PR headline.

mrtraser (approved):

PR #489, opened 2021-02-25
fix: allows CDF authentication from localhost via http

torbjornopheim:
I was developing on localhost over HTTP, which is supported by our api, and it didn't say anywhere in the documentation that it didn't for the sdk, just that using https was in the guide.

It seems that the sdk didn't support non-ssl connections as it had a isUsingSsl conditional around the login, so the user(me) was hitting redirect loops. This in turn increases the payloads in the /login call, until it hit 431( Response header fields too large).

As I see it, the quick fix would be to remove the conditional, which is what I've suggested for now.

If someone with more knowledge about this codebase can confirm that this will work, that would be great, seeing as it has only been handling https for a while.

### DIFF
--- a/packages/core/src/login.ts
+++ b/packages/core/src/login.ts
@@ -19,6 +19,7 @@ import {
   getBaseUrl,
   isSameProject,
   isUsingSSL,
+  isLocalhost,
   promiseCache,
   removeQueryParameterFromUrl,
 } from './utils';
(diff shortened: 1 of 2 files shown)


polomani:
@torbjornopheim could you install this locally and test it out? I believe it's the easiest way to know if it works

packages/core/src/login.ts
```diff
@@ -313,7 +314,7 @@ export function createAuthenticateFunction(options: CreateAuthFunctionOptions) {
         }
       };
 
-      if (isUsingSSL()) {
+      if (isUsingSSL() || isLocalhost()) {
```
valgussev, line 317: Do we have a way to implement an integration test to cover this behavior?

mrtraser: We do have some units that cover OAuth, but don't have any integration to cover the whole process

mrtraser (approved):

PR #493, opened 2021-03-05
feat(aad): use LS state as source of truth

vegardok:


packages/core/src/aad.ts
```diff
@@ -156,7 +153,7 @@ export class AzureAD {
    * Can be used for getting user details via Microsoft Graph API
    */
   async getAccountToken(): Promise<string | null> {
-    if (this.account) return null;
+    if (!this.getAccount()) return null;
```
mrtraser, line 156: Nice catch here, my false

mrtraser (approved):
LGTM

PR #501, opened 2021-03-17
chore: add casing support [release]

Sollimann:
[SDL-113] - implement casing support in js sdk

I still think it's a bad idea that we are not consistent about using well-data-layer resource types (Well, Wellbore, Survey, Trajectory, Casing). Casing f.ex is now just a Sequence, compared to Survey which is in fact a Survey data class. I think it can cause confusion and collisions when applications that depend on resource types with the same name, but one from the core SDK and one from ours.

[SDL-113]: https://cognitedata.atlassian.net/browse/SDL-113

### DIFF
--- a/packages/wells/src/__tests__/integration/casings.int.spec.ts
+++ b/packages/wells/src/__tests__/integration/casings.int.spec.ts
@@ -1,6 +1,7 @@
 import { setupLoggedInClient } from '../testUtils';
 import CogniteWellsClient from 'wells/src/client/cogniteWellsClient';
 import { Sequence, SequenceData } from 'wells/src/client/model/Sequence';
+import { Wellbore } from 'wells/src/client/model/Wellbore';
 //import { Well } from 'wells/src/client/model/Well';
 
 // suggested solution/hack for conditional tests: https://github.com/facebook/jest/issues/3652#issuecomment-385262455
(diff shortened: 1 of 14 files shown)


packages/wells/src/client/api/wellboresApi.ts
```diff
@@ -145,4 +146,40 @@ export class WellboresAPI {
 
     return measurements?.items.map((measurement: Measurement) => this.addLazyMethodsForMeasurement(measurement)) 
   }
+
+  public getCasings = async (wellOrWellboreId: number): Promise<Sequence[] | undefined> => {
```
drsphelps, line 176: Could we not add a casing object in the same way we have for trajectories, which then allows the getCasingData to be a method on that object?

drsphelps: Could we also add this into the `addLazyMethodForWellbore` method so we can directly get casings from the wellbore object

Sollimann: I'll add it :+1:

drsphelps (approved):
lgtm

packages/wells/src/__tests__/integration/casings.int.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      const casings: Sequence[] | undefined = await wellbore?.casings();
+      expect(casings?.length).toBeGreaterThan(0);
+      casings?.forEach(async casing => {
+      expect(casing.columns?.length).toBe(7)
+    })
+    })
+  });
+
+
+  test('Get casing data on casing object', async () => {
```
drsphelps, line 91: This is slightly nit: but could we have a test here that passes some options through to the casing.data(). Just to test that the start/end/columns are being handled correctly?

PR #506, opened 2021-03-25
chore: support for bluefield, azure dev, and bp northeurope [release]

Sollimann:
[SDL-489] 

Also removed boilerplate code

[SDL-489]: https://cognitedata.atlassian.net/browse/SDL-489


sighol (commented):
In our SDK's we have a `Cluster` enum that we can select from. Should we maybe instead just let the user set a baseUrl directly? Like in the cognite sdk.

So we can have two settings. One for baseUrl. And another one for the env keyword. But I'm not sure how to make that obvious to the user. If he select cogdev, if he want's to connect to `bp`, he has to set `env` to `bp`. Hmm

Sollimann:
@sighol Not sure which makes the most sense at this point. We have already had our first user (Geir), mistaking base url for being the CDF base url and not the well-service base url. Right now in the Python SDK, if you pass a base url in the constructor, that will override the base url selected from the `Cluster` enum.

image

In the js sdk,  you will pass the `Cluster` and it will set the baseUrl and `env` for you. So if you pass `Cluster.BP` it will use the cogdev base url and the `?env=bp` param.

packages/wells/src/client/baseWellsClient.ts
```diff
@@ -148,3 +162,39 @@ export default class BaseWellsClient {
...
+    if (this.project == undefined) {
+      throw new HttpError(400, 'The client project has not been set.', {});
+    }
+    if (this.cluster == undefined) {
+      throw new HttpError(400, 'No cluster has been set.', {});
+    }
+    let path =
+      this.cluster == Cluster.BLUEFIELD
+        ? `/${this.project}${baseUrl}`
+        : `/${this.project}${baseUrl}?env=${this.cluster}`;
```
sighol: It is only the cogdev environment that needs an `?env` variable. And only if you use something other than `api`.
You could create a function that takes in a cluster and returns a pair of `baseUrl`, and `env`. And if `env` is `!= null`, then use that.

packages/wells/src/client/baseWellsClient.ts
```diff
@@ -148,3 +162,39 @@ export default class BaseWellsClient {
...
+
+    if (cursor) {
+      path += `&cursor=${cursor}`;
+    }
```
sighol: Does this work if we don't have an env keyword?
I think this URL would be illegal: `/wells&cursor=abc123`.

drsphelps (approved):
lgtm, only thing is that we might need to document people can use the cluster values

PR #511, opened 2021-03-31
chore: spuddate to wells [release]

Sollimann:
add spuddate to wells

### DIFF
--- a/packages/wells/src/client/api/wellsApi.ts
+++ b/packages/wells/src/client/api/wellsApi.ts
@@ -56,6 +56,7 @@ export class WellsAPI extends ConfigureAPI {
       block: well.block,
       field: well.field,
       operator: well.operator,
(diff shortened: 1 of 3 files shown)


packages/wells/src/client/api/wellsApi.ts
```diff
@@ -56,6 +56,7 @@ export class WellsAPI extends ConfigureAPI {
       block: well.block,
       field: well.field,
       operator: well.operator,
+      spudDate: well.spudDate != undefined ? new Date(well.spudDate) : null,
```
Sollimann, line 59: This is a hack I had to do to be able to call methods on the date object i.e 'toLocalDateString()'. Wasn't able to do that without explicitly creating a object using `new`

caelyn-cognite (approved):

PR #513, opened 2021-04-06
chore(release): lock version to a specific date

Tapped:


vegardok:
what is the background for this?

Tapped:
It's part of the release system that was designed last year. It will allow us to do gracefully breaking changes if required during a beta period.

vegardok (approved):

PR #518, opened 2021-04-12
feat: support for Filter3DAssetMapping endpoint

haakonflatval-cognite:


haakonflatval-cognite:
Superceded by #470 (which I found out about a tad bit too late)

PR #520, opened 2021-04-13
chore(adfs): update adfs resource assumption

vegardok:


begozcan (approved):
We should probably fix these unit tests some time 👀

PR #521, opened 2021-04-13
chore: trigger [release] 

vegardok:


begozcan (approved):
Wait..
Name it -> chore: trigger [release] instead of ~~chore: [trigger] release~~ 😅

PR #524, opened 2021-04-21
feat(core): don't check if its https if its not a browser

jarlah:
In air-api in application-services I have thousand of thousands of requests that push "You should use SSL (https) when you login with OAuth since CDF only allows redirecting back to an HTTPS site" to the error logs in Stackdriver in production. It's pretty much a useless warning (that sadly is picked up with severity ERROR by Stackdriver). But the problem is, why should it be there? When there is no browser? So, the isBrowser was added to avoid "null pointer exceptions". This change is making sure its never called unless its a browser. specifically where it tries to log the statement above. 

I assume the function isUsingSSL is used other places, so I don't touch it.

### DIFF
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -26,6 +26,7 @@ import {
   isOAuthWithCogniteOptions,
   isUsingSSL,
   projectUrl,
+  isBrowser,
 } from './utils';
 import { version } from '../package.json';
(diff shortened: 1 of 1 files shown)


polomani (approved):
LGTM

vegardok:
you have to add `[release]` to the PR for it to be released to the current major version

vegardok (approved):

PR #527, opened 2021-04-26
chore: added filter slow method and test [release]

Sollimann:
[SDL-590] - Add 'slow' option to the javascript sdk for filtering wells

Turned out our cursoring of wells have never actually worked. The `cursor` field of `WellItems` should have been `nextCursor`, so it always ended up being undefined/null.

[SDL-590]: https://cognitedata.atlassian.net/browse/SDL-590

### DIFF
--- a/packages/wells/README.md
+++ b/packages/wells/README.md
@@ -109,6 +109,8 @@ const wells = await client.wells.filter(filter);
 
 #### _Filter wells without cursor (get ALL results sequentially):_
 
+Get wells from the well-service until `nextCursor` is empty. Due to performance issue in the well-service, this can take a long time. We don't recommend using this in a user interface where the user expects to see the results as fast as possible. Consider implementation your own pagination so that you can show part of the response as it arrives.
+
 ```ts
 import { WellFilter } from '@cognite/sdk-wells';
 
--- a/packages/wells/README.md
+++ b/packages/wells/README.md
@@ -107,6 +107,19 @@ const filter: WellFilter = {
 const wells = await client.wells.filter(filter);
 ```
 
+#### _Filter wells without cursor (get ALL results sequentially):_
+
+```ts
+import { WellFilter } from '@cognite/sdk-wells';
+
+const polygon = 'POLYGON ((0.0 0.0, 0.0 80.0, 80.0 80.0, 80.0 0.0, 0.0 0.0))';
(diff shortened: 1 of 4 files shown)


packages/wells/src/client/api/wellsApi.ts
```diff
@@ -131,12 +142,33 @@ export class WellsAPI extends ConfigureAPI {
...
-    });
+    return filter
+  }
+
+  public filterSlow = async (userFilter: WellFilter): Promise<Well[] | undefined> => {
+    let wells: Well[] = []
+    let cursor = undefined
+    const filter: WellFilterAPI = this.wrapFilter(userFilter);
+    do {
+        const wellsChunk: WellItems | undefined = await this.sendFilterRequest(filter, cursor);
```
sighol, line 153: Can sendFilterRequest return undefined? :open_mouth:

packages/wells/src/client/api/wellsApi.ts
```diff
@@ -131,12 +142,33 @@ export class WellsAPI extends ConfigureAPI {
...
+        if (wellsChunk == undefined) {
+          break
+        }
+
+        if (wellsChunk.items.length > 0) {
+          wells = wells.concat(wellsChunk.items)
+        }
```
sighol: why check the length? Can't you just write `wells = wells.concat(wellsChunk.item)`?

packages/wells/README.md
```diff
@@ -107,6 +107,19 @@ const filter: WellFilter = {
 const wells = await client.wells.filter(filter);
 ```
 
+#### _Filter wells without cursor (get ALL results sequentially):_
+
```
sighol, line 111: I think we should add something about performance here. 

> Get wells from the well-service until `nextCursor` is empty. Due to performance issue in the well-service, this can take a long time. We don't recommend using this in a user interface where the user expects to see the results as fast as possible. Consider implementation your own pagination so that you can show part of the response as it arrives.

Or something.

sighol (approved):

packages/wells/src/client/api/wellsApi.ts
Sollimann: yes, since the `Promise` can throw an error, it can return an `undefined`. Typescript complains if it is not added.

PR #534, opened 2021-04-29
fix: typo [release]

Sollimann:
fix typo

### DIFF
--- a/packages/wells/src/client/api/wellsApi.ts
+++ b/packages/wells/src/client/api/wellsApi.ts
@@ -32,7 +32,7 @@ export class WellsAPI extends ConfigureAPI {
       }
     }
     return {
-      wellType: filter.wellType,
+      wellType: filter.wellTypes,
       stringMatching: filter.stringMatching,
       quadrants: filter.quadrants,
       blocks: filter.blocks,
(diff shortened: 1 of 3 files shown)


packages/wells/src/client/model/WellFilter.ts
```diff
@@ -8,7 +8,7 @@ export interface WellFilter {
    * @type {string[]}
    * @memberof WellFilter
    */
-  wellType?: string[];
+  wellTypes?: string[];
```
drsphelps, line 11: I think this needs to be changed in the `WellFilterAPI` type as well

packages/wells/src/client/api/wellsApi.ts
```diff
@@ -32,7 +32,7 @@ export class WellsAPI extends ConfigureAPI {
       }
     }
     return {
-      wellType: filter.wellType,
+      wellType: filter.wellTypes,
```
sighol: Should it be `wellTypes: ...` here?

sighol (approved):

packages/wells/src/client/api/wellsApi.ts
Sollimann: updated in `WellTypesApi` as well now

drsphelps (approved):
lgtm

PR #535, opened 2021-04-29
chore: uncomment asset type import

Sollimann:
Tried to create a separate nodejs typescript project, but wouldn't compile due to us having `export * @cognite/sdk` in the `/src/types.ts` file. Compiles now after deleting the file and importing `Asset` directly from the SDK. Also removed geospatial as dependecy.

### DIFF
--- a/packages/wells/src/client/model/Well.ts
+++ b/packages/wells/src/client/model/Well.ts
@@ -1,6 +1,5 @@
 import { Wellhead } from './wellhead';
 import { Wellbore } from './Wellbore';
-// import { Asset } from 'wells/src/types';
 import { Asset } from '@cognite/sdk';
 import { DoubleWithUnit } from './DoubleWithUnit';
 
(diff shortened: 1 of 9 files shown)


packages/wells/src/client/model/Well.ts
```diff
@@ -1,6 +1,7 @@
 import { Wellhead } from './wellhead';
 import { Wellbore } from './Wellbore';
-import { Asset } from 'wells/src/types';
+// import { Asset } from 'wells/src/types';
```
sighol: why comment it out and not just remove it?

packages/wells/src/client/model/Wellbore.ts
```diff
@@ -1,4 +1,5 @@
-import { Asset } from 'wells/src/types';
+// import { Asset } from 'wells/src/types';
```
sighol: same here

packages/wells/src/client/model/Well.ts
Sollimann: ups, missed this one

sighol (approved):
lgtm

PR #538, opened 2021-05-03
feat(diagram): stable diagram api

chingyulin:
This pr introduces a new diagram API. The service contract can be found here: https://github.com/cognitedata/service-contracts/pull/1273


packages/stable/src/types.ts
```diff
@@ -2763,3 +2766,137 @@ export interface EntityMatchingFilter {
...
+   * Allow partial (fuzzy) matching of entities in the engineering diagrams. Creates a match only when it is possible to do so unambiguously.
+   */
+  partialMatch?: string;
+  /**
+   * Each detected item must match the detected entity on at least this number of tokens. A token is a substring of consecutive letters or digits.
+   */
+  minTokens?: number;
+}
+
+type OneOfFiledId =
```
issaafalkattan:
```suggestion
type FileIdEither =
```
SDK uses IdEither, so probably good to keep a similar naming convention

packages/stable/src/types.ts
```diff
@@ -2763,3 +2766,137 @@ export interface EntityMatchingFilter {
...
+type AllOfFileId = {
+  /**
+   * The ID of a file in CDF.
+   */
+  fileId: CogniteInternalId;
+} & {
+  /**
+   * The external ID of a file in CDF.
+   */
+  fileExternalId: CogniteExternalId;
```
issaafalkattan: Aren't file external Ids optional?

issaafalkattan (approved):
Looks great 🎉! 
Couple suggestions but non-blocking.

packages/stable/src/types.ts
chingyulin: I thought the pipeline symbol "|" makes it "either-or" already?

packages/stable/src/types.ts
```diff
@@ -2763,3 +2766,167 @@ export interface EntityMatchingFilter {
...
+       */
+      fileExternalId: CogniteExternalId;
+    };
+
+type AllOfFileId = {
+  /**
+   * The ID of a file in CDF.
+   */
+  fileId: CogniteInternalId;
+} & {
```
```suggestion
```

packages/stable/src/types.ts
polomani: this type (`AllOfFileId`) is a part of response type `DiagramDetectResultItem`, not a request.
Is it by design?  I don't see it in the docs

packages/stable/src/types.ts
```diff
@@ -2763,3 +2766,167 @@ export interface EntityMatchingFilter {
...
+   * Allow partial (fuzzy) matching of entities in the engineering diagrams. Creates a match only when it is possible to do so unambiguously.
+   */
+  partialMatch?: string;
+  /**
+   * Each detected item must match the detected entity on at least this number of tokens. A token is a substring of consecutive letters or digits.
+   */
+  minTokens?: number;
+}
+
+type OneOfFiledId =
```
```suggestion
type FileIdEither =
```
better to use postfix instead of prefix, also change it so it's more consistent with `IdEither` type

packages/stable/src/types.ts
```diff
@@ -2763,3 +2766,167 @@ export interface EntityMatchingFilter {
...
+      fileId: CogniteInternalId;
+    }
+  | {
+      /**
+       * The external ID of a file in CDF.
+       */
+      fileExternalId: CogniteExternalId;
+    };
+
+type AllOfFileId = {
```
```suggestion
type FileIdAndExternalId = {
```
so it is more self-explanatory

packages/stable/src/types.ts
chingyulin: Yes, the request can only reference a file with either fileId or fileExternalId, not the both, while the result could return fileId and maybe fileExternalId

polomani (approved):
I looked briefly, LGTM

remember to squash :)

PR #542, opened 2021-05-05
feat(wells): add sourcewellbores to to wellbore type

sighol:
Add `sourceWellbores[]` to the `Wellbore` type.

### DIFF
--- a/packages/wells/src/__tests__/integration/wellbores.int.spec.ts
+++ b/packages/wells/src/__tests__/integration/wellbores.int.spec.ts
@@ -22,6 +22,8 @@ enum MeasurementType {
   LOT = 'LOT',
 }
 
+const casingWearWellboreId = 8269456345006483;
+
 // suggested solution/hack for conditional tests: https://github.com/facebook/jest/issues/3652#issuecomment-385262455
 const describeIfCondition =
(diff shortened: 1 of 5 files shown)


packages/wells/src/__tests__/integration/wellbores.int.spec.ts
```diff
@@ -22,6 +22,8 @@ enum MeasurementType {
   LOT = 'LOT',
 }
 
+const casingWearWellboreId = 8269456345006483;
```
Sollimann, line 25: We might want to do the same thing here as we've done in the python sdk that we first call CDF using the regular sdk to get the id based on an external id so that the tests don't fail if we re-ingest data

Sollimann: This is not specific to this PR though. We should do it throughout all the tests

Sollimann: created a card for it: https://cognitedata.atlassian.net/secure/RapidBoard.jspa?rapidView=391&modal=detail&selectedIssue=SDL-631&assignee=5d1c705543d1510d3accfb4a

Sollimann (approved):
lgtm :+1:

PR #543, opened 2021-05-05
fix: readme and remove dead code [release]

Sollimann:
typo in docs + removing duplicated method


packages/wells/src/client/api/wellboresApi.ts
```diff
@@ -118,17 +118,6 @@ export class WellboresAPI extends ConfigureAPI {
...
-    const path: string = this.getPath(`/wellbores/${wellboreId}/trajectory`);
-
-    return await this.client
-      .asyncGet<Survey>(path)
-      .then(response => response.data)
-      .catch(err => {
-        throw new HttpError(err.status, err.data.error.message, {});
-      });
-  };
-
```
Sollimann, line 126: we already have this method in the `SurveysAPI` class

drsphelps (approved):
lgtm

PR #556, opened 2021-05-31
fix: wellbore class [release]

Sollimann:


packages/wells/src/client/model/Wellbore.ts
```diff
@@ -48,12 +48,6 @@ export interface Wellbore {
    * @memberof Wellbore
    */
   id: number;
-  /**
-   * Wellbore description
-   * @type {string}
-   * @memberof Wellbore
-   */
-  description?: string;
```
sighol: Why have you removed the wellbore description?

drsphelps, line 56: Didn't someone add the description for wellbores last week on request from discover?

Sollimann: Couldn't find it in the data class: 
image

drsphelps: Make sure you have the latest version of the service, it was only added recently

drsphelps (approved):

PR #558, opened 2021-06-04
feat: add filters for nds events [release]

drsphelps:


packages/wells/src/__tests__/integration/wells.int.spec.ts
```diff
@@ -655,4 +655,49 @@ describeIfCondition('CogniteClient setup in wells - integration test', () => {
     const wells = await client.wells.filter(filter);
     expect(wells).not.toBeUndefined();
   });
+
+  test('filter well on nds probability and severity', async () => {
+    const filter: WellFilter = {
+      nds: {
+        severities: ["1"],
+        probabilities: ["1"]
```
Sollimann: In the well-service. Would it make more sense to have an `Array<Int>` instead of `Array<String>`?

Sollimann (approved):
Looks good :+1:

PR #564, opened 2021-06-07
chore: add github action to run test suite

vegardok:


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,48 @@
+name: Node.js CI lint/codecov/commit check
+
+on:
```
andersfylling: If you're moving away from jenkins, you might want to run this on pull requests as well:
```yaml
on:
  push:
    branches:
    - master
  pull_request:
    branches:
    - master
```

vegardok: @andersfylling  The idea was that this should _only_ run on pull requests, I have a 3rd workflow that should run on master pushes. Fixed. (and it's moving away from Travis, not jenkins)

andersfylling (approved):

andersfylling:
LGTM! I see some things are failing, but as it isn't immediately obvious to me I wonder if it's tooling related(?). When it comes to the script, I think it all seems good.

vegardok:
> LGTM! I see some things are failing, but as it isn't immediately obvious to me I wonder if it's tooling related(?). When it comes to the script, I think it all seems good.

I would guess secrets, but that doesn't explain why it's going on and on without failing. I'll get it running before merging 👍

PR #565, opened 2021-06-07
feat(document preview): [SBS-3717] Endpoints for document preview

andersfylling:
Endpoints for the document preview service found under /documents:
 - https://cognitedata.atlassian.net/browse/SBS-3717


packages/beta/src/api/documents/previewApi.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020 Cognite AS
+
+import { BaseResourceAPI } from '@cognite/sdk-core';
+
+import { DocumentId, DocumentPreviewTemporaryLink } from '../../types';
+import Any = jasmine.Any;
+
+export class PreviewAPI extends BaseResourceAPI<Any> {
+  public document = (
```
andersfylling, line 10: How exactly can I go about supporting file download?

and also how can I remove the "accept: application/json" header that is being injected?

andersfylling:
@vegardok do you know why that CI/lint is failing? I updated the PR title and re-ran all jobs, but it seems to still fail on an outdated title.

andersfylling:
@vegardok these tests should fail as the unit test references a method that does not exist.

PR #571, opened 2021-06-08
feat(project): partial project update

lassemand:


packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,124 @@ export interface ProjectUpdate {
...
+*/
+export interface SinglePatchLong {
+    set?: number;
+    setNull?: boolean;
+}
+
+export interface SinglePatchOAuth2Configuration {
+    set?: OAuth2Configuration;
+    setNull?: boolean;
+}
```
polomani: Try to use/reuse generic union types, something like `SinglePatchString` but maybe with a generic type?

polomani: Using optional fields in this case is less preferred, as `set` and `setNull` exclude each other

packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,124 @@ export interface ProjectUpdate {
...
+    defaultGroupId?: SinglePatchLong;
+    validDomains?: ArrayPatchString;
+    applicationDomains?: ArrayPatchString;
+    authenticationProtocol?: SinglePatchRequiredString;
+    azureADConfiguration?: SinglePatchAzureADConfiguration;
+    oAuth2Configuration?: SinglePatchOAuth2Configuration;
+    oidcConfiguration?: OidcConfigurationUpdate;
+}
+
+export interface OidcConfigurationDTO {
```
```suggestion
export interface OidcConfiguration {
```

packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,124 @@ export interface ProjectUpdate {
...
+}
+
+export interface ProjectUpdateDTO {
+    update: ProjectUpdateObjectDTO;
+}
+
+/**
+* Contains the instructions on how to update the project. Note: azureADConfiguration, oidcConfiguration and oAuth2Configuration are mutually exclusive
+*/
+export interface ProjectUpdateObjectDTO {
```
```suggestion
export interface ProjectUpdateObject {
```
We don't use `DTO` postfix

packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,124 @@ export interface ProjectUpdate {
...
+    logClaims?: ModifyPatchClaimNames;
+}
+
+export interface ModifyPatchClaimNames {
+    set?: Array<ClaimName>;
+    add?: Array<ClaimName>;
+    remove?: Array<ClaimName>;
+}
+
+export interface ProjectUpdateDTO {
```
```suggestion
export interface ProjectUpdate {
```

packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,124 @@ export interface ProjectUpdate {
...
+export interface ClaimName {
+    claimName?: string;
+}
+
+export interface ModifyPatchInteger {
+    set?: number;
+    setNull?: boolean;
+}
+
+export interface OidcConfigurationUpdateDTO {
```
```suggestion
export interface OidcConfigurationUpdate {
```

packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,124 @@ export interface ProjectUpdate {
...
+    tokenUrl?: SinglePatchString;
+    issuer?: SetField<string>;
+    audience?: SetField<string>;
+    skewMs?: ModifyPatchInteger;
+    accessClaims?: ModifyPatchClaimNames;
+    scopeClaims?: ModifyPatchClaimNames;
+    logClaims?: ModifyPatchClaimNames;
+}
+
+export interface ModifyPatchClaimNames {
```
polomani: could you make a generic type for this?
like `ArrayPatchLong`

packages/stable/src/types.ts
polomani: We try to not use DTO postfix

polomani:
you need to run
```
yarn build && yarn lint:fix
```
locally.

polomani:
Also, read https://github.com/cognitedata/cognite-sdk-js/blob/master/CONTRIBUTING.md
There are some guidelines about commit messages

packages/stable/src/types.ts
lassemand: That was annoying. I just added a script that copied the model of the service contract into types.ts to avoid manual typing errors.

polomani: yes, but service contracts identifiers are not written in the way to be used as interface names)
we could try to enforce it at some point, but it's a bigger initiative.

lassemand:
> Also, read https://github.com/cognitedata/cognite-sdk-js/blob/master/CONTRIBUTING.md
> There are some guidelines about commit messages

Sure will squash all the commits

PR #573, opened 2021-06-09
feat(project): partial project update

lassemand:


lassemand:
With the current setup partial project update can not be tested.

My suggestion is that a separate instance of api-auth is spinned up behind ambassador prior to the tests on project update is running. That would no matter what be dealt with in another PR

packages/stable/src/types.ts
```diff
@@ -209,9 +209,17 @@ export interface ApiKeyRequest {
   serviceAccountId: CogniteInternalId;
 }
 
+export type ArrayPatchString =
+  | { set: Set<string> }
+  | { add?: Set<string>; remove?: Set<string> };
```
polomani: Shouldn't it be just an array here?

packages/stable/src/types.ts
```diff
@@ -209,9 +209,17 @@ export interface ApiKeyRequest {
...
+
 export type ArrayPatchLong =
-  | { set: number[] }
-  | { add?: number[]; remove?: number[] };
+  | { set: Array<number> }
+  | { add?: Array<number>; remove?: Array<number> };
+
+export type ArrayPatchClaimNames =
+  | { set: Array<ClaimName> }
+  | { add?: Array<ClaimName>; remove?: Array<ClaimName> };
```
```suggestion
  | { add?: ClaimName[]; remove?: ClaimName[] };
```
we use short syntax in other places

packages/stable/src/types.ts
lassemand: Nope. They need to be unique.

polomani: have you checked that this actually works?
```
JSON.stringify(new Set("a", "b"))
```
prints  `{}`

polomani: tbh, that's up to the api to perform such validations.
sets are not necessarily easy to work with

lassemand: That seems like an unfortunate behaviour.

polomani (approved):
LGTM

lassemand:
@vegardok

begozcan (approved):

PR #576, opened 2021-06-11
feat: appended oidc configuration to project response

lassemand:


packages/stable/src/types.ts
```diff
@@ -1656,6 +1656,7 @@ export interface ProjectResponse {
   urlName: UrlName;
   defaultGroupId?: DefaultGroupId;
   authentication?: OutputProjectAuthentication;
+  oidcConfiguration?: OidcConfiguration;
```
lassemand, line 1659: https://github.com/cognitedata/service-contracts/blob/796d064d4830ab1ea603788783681aaa7a442b4a/versions/v1/projects.yml#L440

vegardok (approved):

PR #578, opened 2021-06-15
chore(tests): remove bash retry

vegardok:
There are some retry-setup in the test suite. That should be used/changed of the current retry isn't
good enough.


andersfylling (approved):

andersfylling:
I figured yarn returned exit code 0 as it successfully called the jest/lint tools. But this might be the actual fix!

PR #579, opened 2021-06-15
feat: document preview endpoints [release]

andersfylling:
Endpoints for the document preview service found under /documents:
 - https://cognitedata.atlassian.net/browse/SBS-3717

### DIFF
--- a/packages/beta/src/__tests__/api/documentsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/documentsApi.int.spec.ts
@@ -43,7 +43,6 @@ describe.skip('documents api', () => {
     expect(documents.items.length).toEqual(1);
     const document = documents.items[0];
 
-    await client.documents.preview.document(document.id, mediaTypePDF);
     await client.documents.preview.documentAsPdf(document.id);
     await client.documents.preview.documentAsImage(document.id, 0);
   });
--- a/packages/beta/src/__tests__/api/documentsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/documentsApi.int.spec.ts
@@ -34,4 +34,17 @@ describe.skip('documents api', () => {
     const response = await client.documents.pipelines.list();
     expect(response.items.length).toBeGreaterThan(0);
   });
+  test('fetch preview', async () => {
+    const mediaTypePDF = 'application/pdf';
+    const documents = await client.documents.list({
+      limit: 1,
+      filter: { mimeType: { equals: mediaTypePDF } },
+    });
+    expect(documents.items.length).toEqual(1);
+    const document = documents.items[0];
+
+    await client.documents.preview.document(document.id, mediaTypePDF);
+    await client.documents.preview.documentAsPdf(document.id);
+    await client.documents.preview.documentAsImage(document.id, 0);
+  });
 });
--- a/packages/beta/src/__tests__/api/documentsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/documentsApi.int.spec.ts
@@ -21,7 +21,7 @@ describe.skip('documents api', () => {
   });
   test('list documents, by size', async () => {
     const response = await client.documents.list({
-      filter: { size: { min: 10, max: 900533317 } },
+      filter: { size: { min: 10, max: 9005383317 } },
       limit: 1,
     });
     expect(response.items.length).toEqual(1);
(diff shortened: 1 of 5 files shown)


packages/beta/src/api/documents/previewApi.ts
```diff
@@ -0,0 +1,68 @@
...
+    const response = await this.get<ResponseType>(this.url('temporaryLink'), {
+      params: {
+        documentId: documentId,
+      },
+    });
+    return response.data;
+  }
+}
+
+export type acceptsType = 'application/pdf' | 'image/png';
```
maaland: nit: capitalize the first letter: `AcceptsType`

packages/beta/src/__tests__/api/documentsApi.unit.spec.ts
```diff
@@ -237,4 +237,53 @@ describe('Documents unit test', () => {
       .reply(200, { items: [] });
     await client.documents.feedback.reject([{ id: 1 }]);
   });
+
+  test('document preview uri', async () => {
+    const base = (path: string): string => {
+      return `/api/v1/projects/${project}/documents${path}`;
+    };
+
+    const prev = client.documents.preview;
```
maaland: nit: `prev` usually reads `previous`. I'd prefer if you rename this to `preview`

packages/beta/src/__tests__/api/documentsApi.unit.spec.ts
```diff
@@ -237,4 +237,53 @@ describe('Documents unit test', () => {
...
+    );
+  });
+
+  test('document preview pdf', async () => {
+    nock(mockBaseUrl)
+      .get(new RegExp('/documents/preview'))
+      .query({ documentId: 1 })
+      .once()
+      .reply(200);
+    await client.documents.preview.documentAsPdf(1);
```
maaland, line 267: Should we assert that the response is as expected here? Not just the status code, but the contents.

packages/beta/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -34,4 +34,16 @@ describe.skip('documents api', () => {
...
+    const mediaTypePDF = 'application/pdf';
+    const documents = await client.documents.list({
+      limit: 1,
+      filter: { mimeType: { equals: mediaTypePDF } },
+    });
+    expect(documents.items.length).toEqual(1);
+    const document = documents.items[0];
+
+    await client.documents.preview.documentAsPdf(document.id);
+    await client.documents.preview.documentAsImage(document.id, 0);
```
maaland, line 47: Should we assert that the response is as expected here? Not just the status code, but the contents.

packages/beta/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -34,4 +34,16 @@ describe.skip('documents api', () => {
     const response = await client.documents.pipelines.list();
     expect(response.items.length).toBeGreaterThan(0);
   });
+  test('fetch preview', async () => {
```
maaland, line 37: Can you also make an integration test of fetching a temporary link?

maaland (changes requested):
Some comments and nitpicks, but this looks nice!

packages/beta/src/__tests__/api/documentsApi.unit.spec.ts
andersfylling: I have wondered about uploading a tiny tiny pdf and compare against that ID specifically. But if that file is accidentally deleted, the integration tests for the sdk here is going to halt every new PR.

Uploading a new document for every run and then checking against that could be done. But the added time concerned me. What do you think?

maaland: I agree, we wouldn't want a brittle situation. But in this unit test you're mocking the response, right? My `ts` is rusty. Could you add a hardcoded arraybuffer there? It doesn't need to make sense other than having whatever headers are needed to be a pdf. I think we do something similar in the tests for the document preview service itself?

maaland: https://github.com/cognitedata/document-preview/blob/master/src/test/kotlin/com/cognite/documentpreview/controller/PreviewControllerTest.kt#L74

maaland: Might be more hassle than it's worth tho, as you'll kind of just end up testing the mock

packages/beta/src/__tests__/api/documentsApi.int.spec.ts
maaland: When will this test fail?

andersfylling: it's just to verify against our api that the request sent is correct (like typo in the mime value) or incorrect headers

packages/beta/src/__tests__/api/documentsApi.unit.spec.ts
andersfylling: Added a test for this in the integration test when a pdf is fetched

maaland (approved):

packages/beta/src/api/documents/previewApi.ts
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  public documentAsImage = (
+    id: DocumentId,
+    page: number
+  ): Promise<ArrayBuffer> => {
+    return this.previewEndpoint<ArrayBuffer>(id, 'image/png', page);
+  };
+
+  public temporaryLink = (
```
polomani, line 21: is there any documentation planned? code snippets?

andersfylling: yes, the documentation can be found here under /documents: https://docs.cognite.com/api/playground/#operation/getIcon

But did not plan to copy the documentation into the sdk in case the openapi documentation changes. Otherwise we will just bring up code snippets in our migration guide https://cognitedata.atlassian.net/wiki/spaces/search/pages/2923987180/Migration+guide+files+unstructured+documents

andersfylling: Was thinking about adding the openapi link to each endpoint when we move to v1, if this is what you were thinking about?

polomani: Yes, so you could do that later :)

tommythorsen (approved):

begozcan (approved):

PR #585, opened 2021-06-18
chore: fix github release action [release]

polomani:
Have found this example on the internet. We might be accessing the wrong variable

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,7 +27,7 @@ jobs:
         run: yarn build
 
   publish:
-    if: ${{ contains(github.event.push.head_commit.message, '[release]') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
         run: yarn build
 
   publish:
-    if: ${{ contains(github.event.push.head_commit.message, '[release]') }}
+    if: ${{ contains(github.event.head_commit.message, '[release]') }}
```
vegardok:
```suggestion
    if: ${{ contains(github.event.commits[0].message, '[release]') }}
```

vegardok (approved):

PR #587, opened 2021-06-22
feat(core): add oidc auth code flow [release]

vegardok:
BREAKING CHANGE: loginWithOAuth signature is changed


packages/core/src/baseCogniteClient.ts
```diff
@@ -234,46 +267,51 @@ export default class BaseCogniteClient {
    * const client = new CogniteClient({ appId: '[YOUR APP NAME]' });
    *
    * // using Cognite authentication flow
-   * client.loginWithOAuth({
-   *   project: '[PROJECT]',
-   *   onAuthenticate: REDIRECT // optional, REDIRECT is by default
+   * client.loginWithOAuth('CDF_OAuth', {
```
jarlah: TODO: update these docs

packages/core/src/baseCogniteClient.ts
```diff
@@ -589,7 +631,7 @@ export default class BaseCogniteClient {
         user = null;
       }
 
-      if (!user || !user.access_token) {
+      if (!user || user.access_token) {
```
jarlah, line 626: is this intentional leaving out the `!`in front of `user.access_token`? did it came from my branch 🙈

samples/react/authentication-adfs/src/App.js
```diff
@@ -57,7 +57,7 @@ function App() {
 
   useEffect(() => {
     const login = async (client) => {
-      const result = await client.loginWithOAuth({
+      const result = await client.loginWithOAuth('ADFS_OAUTH', {
```
jarlah: needs to be updated to send in flow object

samples/react/authentication-aad/src/App.js
```diff
@@ -61,7 +61,7 @@ function App() {
 
   useEffect(() => {
     const login = async (client) => {
-      const result = await client.loginWithOAuth({
+      const result = await client.loginWithOAuth('AAD_OAUTH', {
```
jarlah: needs to be updated to send in flow object

packages/core/src/baseCogniteClient.ts
```diff
@@ -235,46 +263,66 @@ export default class BaseCogniteClient {
...
+   * options: {
+   * authority: https://example.com/adfs/oauth2/authorize,
+   * requestParams: {
+   * cluster: 'cluster-name',
+   * clientId: 'adfs-client-id',
+   * },
+   * }
+   * });
+   *
+   * // or with client credentials
```
jarlah: this is not a part of this PR but will be a part of my PR. possibly remove.

packages/core/src/baseCogniteClient.ts
```diff
@@ -58,14 +64,32 @@ export interface ApiKeyLoginOptions extends Project {
...
+export type ADFS_OAUTH = {
+  type: 'ADFS_OAUTH';
+  options: OAuthLoginForADFSOptions;
+};
+export type OIDC_AUTHORIZATION_CODE_FLOW = {
+  type: 'OIDC_AUTHORIZATION_CODE_FLOW';
+  options: OAuthLoginForOIDCAuthFlowOptions;
+};
 
+export type AuthFlowType =
```
jarlah, line 85: so glad we did this. 🙌 much more readable and only by using a basic union with common fields

packages/core/src/baseCogniteClient.ts
jarlah: so if we don't have a user or a user WITH access token .... I think this is wrong

samples/react/authentication-auth-code-flow/src/App.js
```diff
@@ -0,0 +1,100 @@
...
+  const [token, setToken] = useState();
+  const [inspect, setInspect] = useState();
+  const [isSignedIn, setIsSignedIn] = useState(false);
+  const [error, setError] = useState(undefined);
+
+
+  const client = useMemo(() => new CogniteClient({ appId: 'sample-app-id' }), []);
+
+  useEffect(() => {
+    client.loginWithOAuth('OIDC_AUTHORIZATION_CODE_FLOW', {
```
jarlah: needs to be updated to send in flow object

jarlah (changes requested):
looks good. Have some change requests. and/or comments

packages/core/src/baseCogniteClient.ts
kimjoar: Agree — this approach looks great

jarlah:
I have been doing a lot of thinking on the flow manager approach in the code. So, we have a flow manager and a method inside the base client that handles the flow. So the logic is spread into two places, the flow manager and the `loginWith[FlowType]` method. I wonder, is it possible to merge the contents of the `loginWith[FlowType]` method into the flow manager? We could still continue to "manually" resolve the flow manager in the switch, but we wouldn't need to define a method in the base client. One way to avoid creating this method is to pass functions for setting bearer token and cluster.

This could be solved in the following manner:

```typescript
      case 'OIDC_VENDOR_GENERIC_CLIENT_CREDENTIALS_FLOW': {
        this.vendorGenericFlowManager = new OidcVendorGeneric(flow.options, this.setCluster, this.setBearerToken);
        [authenticate, token] = await this.vendorGenericFlowManager.init();
        break;
      }
```

where `this.setCluster` and `this.setBearerToken` is lambda functions on the class (possible to pass around without lossing this context). Or even just pass them as `(cluster: string) => this.httpClient.setCluster(cluster)`or `this.httpClient.setCluster`(the latter if the setCluster in httpClient is a lambda.

the `init` method name might need to renamed to something else because some flow managers already have an init method. But the point remains.

packages/core/src/baseCogniteClient.ts
```diff
@@ -58,14 +64,32 @@ export interface ApiKeyLoginOptions extends Project {
...
 
-export const AAD_OAUTH = 'AAD_OAUTH';
-export const CDF_OAUTH = 'CDF_OAUTH';
-export const ADFS_OAUTH = 'ADFS_OAUTH';
-export type AuthFlowType =
-  | typeof AAD_OAUTH
-  | typeof CDF_OAUTH
-  | typeof ADFS_OAUTH;
+export type AAD_OAUTH = {
+  type: 'AAD_OAUTH';
```
polomani, line 69: I really like the discriminated union approach.
However, could we still have type strings as constants? `const AAD_OAUTH = 'AAD_OAUTH'`

packages/core/src/baseCogniteClient.ts
```diff
@@ -235,46 +263,66 @@ export default class BaseCogniteClient {
...
    *
    * // or you can sign in using AzureAD authentication flow (in case your projects supports it)
    * client.loginWithOAuth({
-   *   cluster: '[CLUSTER]',
-   *   clientId: '[CLIENT_ID]', // client id of your AzureAD application
-   *   tenantId: '[TENANT_ID]', // tenant id of your AzureAD tenant. Will be set to 'common' if not provided
+   * type: 'AAD_OAUTH',
+   * options: {
+   * cluster: '[CLUSTER]',
+   * clientId: '[CLIENT_ID]', // client id of your AzureAD application
```
polomani: could we fix the indentation in here?

packages/core/src/baseCogniteClient.ts
jarlah: I will make the change in my PR for now which is mostly identical except for missing the tests in this one

jarlah: but why do you want it btw .. the union will take care of it never matching anything else

vegardok: like this?

```
export const AAD_OAUTH_TYPE: 'AAD_OAUTH' = 'AAD_OAUTH';
export type AAD_OAUTH = {
  type: typeof AAD_OAUTH_TYPE;
  options: OAuthLoginForAADOptions;
};
```

not sure if I see the value, is it for better vanilla JS interop?

packages/core/src/baseCogniteClient.ts
vegardok: yea, looks like a typo to me, good catch

samples/react/authentication-adfs/src/App.js
vegardok: fixed

samples/react/authentication-aad/src/App.js
vegardok: fixed

packages/core/src/baseCogniteClient.ts
vegardok: done

samples/react/authentication-auth-code-flow/src/App.js
vegardok: fixed

packages/core/src/baseCogniteClient.ts
vegardok: fixed

polomani (commented):
As I mentioned before, my biggest concern is about a breaking change in general.
We have working adfs/aad flow, and breaking the users just for a cleaner code's sake is not really a good excuse. I doubt we have resources to backport bugfixes/features anyway, so by releasing this we just make external developer's life harder.

I would keep the same method we had before, and just introduce an additional one.
Something like `setLoginConfig(...)` which wouldn't support API keys/legacy OAuth.
As I remember that was what we agreed on earlier.

vegardok:
> We have working adfs/aad flow, and breaking the users just for a cleaner code's sake is not really a good excuse.

The main purpose of the PR is to add OIDC support for Aize/Auth0. The API change is not motivated by clean code, but less brittle code. [this](https://github.com/cognitedata/cognite-sdk-js/pull/587/files#diff-6552eacdc0f4378c00e24a96877df8352eeb0db006852dfdcf511f59d9b781aeL231) is error prone, and the more flows you add, the more error prone it will be.

polomani:
> and the more flows you add, the more error prone it will be

I completely agree, therefore I suggested adding a separate function to handle all these different flows (all of them except the legacy one)

`CDF_OAUTH` will be deprecated in a couple of months, another major?

vegardok:
> > and the more flows you add, the more error prone it will be
> 
> I completely agree, therefore I suggested adding a separate function to handle all these different flows (all of them except the legacy one)

That would still be a breaking change for AAD/ADFS, so I don't see that as an alternative to the objection against frequent major version updates. I think the current setup is a good fit for changing flows in the future, e.g adding client credentials will be a minor change.

The way I see it, this PR will be breaking and is an improvement. I don't see why adding a new method instead of changing the API is better? 
> 
> `CDF_OAUTH` will be deprecated in a couple of months, another major?

Q1 22 at the earliest, but that is _optimistic_. Or, it has been deprecated since Q1 21, but I don't think it will be shut down before the last customer has migrated. So who knows. 

> Something like setLoginConfig(...) which wouldn't support API keys/legacy OAuth.

This is also not correct, you are doing more than just setting the config (checking old tokens, trying to silently refresh them if possible). So the semantics of the existing method is a better fit.

PR #589, opened 2021-06-23
feat!: add client credentials flow

jarlah:
ready for review


packages/core/package.json
```diff
@@ -10,10 +10,12 @@
   "scripts": {
     "clean": "rm -rf dist/ docs/",
     "test": "jest --config=../../jest.config.js --testPathPattern=/core/",
+    "test:debug": "node --inspect-brk ../../node_modules/.bin/jest --config=../../jest.config.js --testPathPattern=/core/ --runInBand",
```
bdimitrijoski: Just out of curiosity, why do you need this here? Can't you do the same from root package.json or maybe create a script that will receive some args?

packages/core/src/authFlows/oidc_auth_code_flow.ts
```diff
@@ -0,0 +1,117 @@
...
+      scope: extraScope,
+      extraQueryParams: authenticateParams,
+      userStore: new WebStorageStateStore({ store: window.localStorage }),
+      loadUserInfo: false,
+
+      redirect_uri: `${window.location.protocol}//${window.location.host}`,
+    });
+  }
+
+  public async init() {
```
bdimitrijoski: Maybe remove `public async`

packages/core/src/authFlows/oidc_auth_code_flow.ts
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+              return Promise.resolve(null);
+            }
+          });
+        } else {
+          throw e;
+        }
+      })
+      .finally(() => {
+        const redirectStateStr = window.localStorage.getItem(localStorageKey);
```
bdimitrijoski: Just a suggestion here. Maybe it is good to write a wrapper class like storage service that will behave like a wrapper and just choose whatever storage type you need inside. Also this can do all the json parsing stuff.

Example:
```JS
export class StorageProvider {
private driver;

constructor(){ this.driver = window.localStorage; }

getItem() {
// check if it exists, do the json parsing and validation here

}
}

```

bdimitrijoski (dismissed):
Nice! 
I wrote some comments, but consider them just as suggestions :)

packages/core/package.json
jarlah: from @vegardok PR. I can remove everything I don't need for this PR

packages/core/src/authFlows/oidc_auth_code_flow.ts
jarlah: please revisit the code. I removed the code from @vegardok s PR

packages/core/src/authFlows/oidc_auth_code_flow.ts
jarlah: I could but its async ;) I can remove public without any side effect, but its good I think to be clear about what methods can be called and what is helper methods. In addition this file is now deleted.

jarlah:
@bdimitrijoski  I removed the stuff from @vegardok s PR and rewrote my own client credentials logic to not have a method within baseCogniteClient and rather put the code inside the OidcClientCredentials class.

packages/core/package.json
jarlah: @bdimitrijoski I removed the test:debug

packages/core/src/baseCogniteClient.ts
```diff
@@ -194,8 +216,8 @@ export default class BaseCogniteClient {
    *
    * const client = new CogniteClient({ appId: '[YOUR APP NAME]' });
    * client.loginWithApiKey({
-   *   apiKey: '[API KEY]',
-   *   project: '[PROJECT]',
+   * apiKey: '[API KEY]',
```
peternic: Nitpick, maybe not remove the indent

peternic (approved):
Not an OIDC expert, but lgtm 👍🏻

packages/core/src/baseCogniteClient.ts
jarlah: fixed 👍

bdimitrijoski (approved):
Looks good!

PR #590, opened 2021-06-23
feat: directoryPrefix filter for files [release]

andersfylling:
https://cognitedata.atlassian.net/browse/CDF-8667

### DIFF
--- a/packages/stable/src/__tests__/api/files.int.spec.ts
+++ b/packages/stable/src/__tests__/api/files.int.spec.ts
@@ -89,6 +89,7 @@ describe.skip('Files integration test', () => {
     expect(retrievedFile.mimeType).toBe(localFileMeta.mimeType);
     expect(retrievedFile.uploaded).toBeTruthy();
     expect(retrievedFile.sourceCreatedTime).toEqual(sourceCreatedTime);
+    expect(retrievedFile.directory).toEqual('/test/testing');
   });
 
   test('retrieve with non-existent id', async () => {
--- a/packages/stable/src/__tests__/api/files.int.spec.ts
+++ b/packages/stable/src/__tests__/api/files.int.spec.ts
@@ -76,6 +76,12 @@ describe.skip('Files integration test', () => {
     expect(file.geoLocation).toEqual(localFileMeta.geoLocation);
   });
 
+  test('validate directoryPrefix format against CDF', async () => {
+    await client.files.list({ filter: {
+      directoryPrefix: "/test" },
+    });
+  });
+
   test('retrieve', async () => {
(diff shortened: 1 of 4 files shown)


maaland (changes requested):
Nice! Maybe a test of upload and list with the `directory` field as well, to be safe?

andersfylling:
> Nice! Maybe a test of upload and list with the `directory` field as well, to be safe?

I'm trying to avoid doing any business logic tests in general. Since the sdk should just have a valid skeleton of the datastructures, and it's up to the user to define the arguments, I prefer just validating the outgoing datastructures/format. Otherwise the sdk can break if there are any changes to behaviour in /files, blocking all other PR's. But maybe this does not matter much if it's in stable?

maaland:
For sure, I agree, not talking about the business logic. It would be good to test that the `directory` field in the file datastructure works, not just the `directoryPrefix` filed in the file filter. 

Looks like the [file create test](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/__tests__/api/files.int.spec.ts#L63) is skipped for some reason, but maybe you can add `directory` to the [update test](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/__tests__/api/files.int.spec.ts#L144) ? Like what was done there when `geoLocation` was added.

maaland:
Actually, are the tests in this file even run? I see that the top-level describe is [skipped](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/__tests__/api/files.int.spec.ts#L17) 🤔

andersfylling:
> Actually, are the tests in this file even run? I see that the top-level describe is [skipped](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/__tests__/api/files.int.spec.ts?rgh-link-date=2021-06-23T15%3A19%3A24Z#L17) thinking

They might be setup tests that they ran "once" when they opened a PR.

maaland:
> > Actually, are the tests in this file even run? I see that the top-level describe is [skipped](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/__tests__/api/files.int.spec.ts?rgh-link-date=2021-06-23T15%3A19%3A24Z#L17) thinking
> 
> They might be setup tests that they ran "once" when they opened a PR.

Doubt it, as the entire outer `describe` containing the tests is skipped (`describe.skip`), meaning it should not run. Confirm by looking at the `test` section of the  [checks](https://github.com/cognitedata/cognite-sdk-js/runs/2895123071?check_suite_focus=true) of the PR. Does not look like this test suite is run, unless I'm missing something.

maaland (approved):

bdimitrijoski (approved):

PR #591, opened 2021-06-24
feat: added support for variables and operationName for templates

peternic:
Added support for variables and operation name to the templates graphql. This will make it possible to use codegen to generate and compiletime check queries.

So need to be able to pass in the variables and the operationName through the sdk.

### DIFF
--- a/packages/beta/src/__tests__/api/templateGraphQlApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/templateGraphQlApi.int.spec.ts
@@ -57,14 +57,18 @@ describe.skip('template group test', () => {
   });
 
   it('should run a GraphQL query', async () => {
-    const result = await client.templates.group(externalId).version(1)
-      .runQuery(`
+    const result = await client.templates
+      .group(externalId)
+      .version(1)
+      .runQuery({
+        query: `
       {
         testList {
           a
           b
         }
-      }`);
+      }`,
+      });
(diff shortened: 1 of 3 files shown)


stignite (approved):
🚀

peternic:
After approval, is it correct that I shall merge this as blow to get it released?
```
feat(templates): added support for variables and operationName for templates [release]

BREAKING CHANGE: The templates runQuery function in beta has changed signature, now accepts an object with query, variables and operationName
```

jarlah:
You have degraded test coverage or test coverage is wrong. I don't see any specifically wrong with PR itself ;) Ive seen `feat!(module): message` floating about but I'm not sure what the `!` means yet.

peternic:
> You have degraded test coverage or test coverage is wrong. I don't see any specifically wrong with PR itself ;) Ive seen `feat!(module): message` floating about but I'm not sure what the `!` means yet.

Yes, seems like it is checking against a very old commit for coverage. Not sure how to fix, any ideas?

The ! Is for breaking changes I think...

polomani:
@peternic `BREAKING CHANGE: ` should work according to https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular#angular-convention
You could also try to run lerna locally and see if it bumps the number.

We had issues with major releases previously, please reach out if you face it :)

PS. ignore codecov

danlevings (approved):

begozcan (approved):

PR #597, opened 2021-06-29
chore: use GH_LERNA_TOKEN for lerna version [release]

peternic:
Lerna version is updating the master branch on release, and the master branch is protected so it fails making the release commit. So adding a temporary admin token to make Lerna able to push to master


bdimitrijoski (approved):

polomani:
you need this as well
```
with:
        persist-credentials: false
        
```

PR #601, opened 2021-06-29
chore: add git username and email for github action [release]

peternic:
Hope this does the trick, adding git username and email for github actions


polomani:
We already doing it earlier in the file, instead we should try to use the correct token
https://github.com/cognitedata/cognite-sdk-js/pull/603

PR #607, opened 2021-06-30
feat: client credentials flow [release]

jarlah:
using two oidc libraries, but there will only be one library in use at a specific time anyway, unless you have two clients in the same context. I would guess this doesn't affect each other. But, should it be possible to use client credentials from browser? If we make it like its here now in the PR, its possible but utterly stupid to use client id and client secret in a browser.


vegardok:
The two libraries should not be an issue since one of them _only_ supports the browser env and the other _only_ supports node. if `openid-client` had browser support we could only use that, but, alas.

vegardok (approved):

PR #609, opened 2021-07-07
feat(revisions3d): parameter to update metadata property [release]

haakonflatval-cognite:


packages/stable/src/types.ts
```diff
@@ -2005,6 +2005,25 @@ export interface RevisionCameraProperties {
...
+
+export interface RevisionMetadataUpdateSet {
+  set: { [key: string]: string };
+}
+
+export interface RevisionMetadataUpdateAddRemove {
+  /**
+   * Key/value pairs to add
+   */
+  add?: { [key: string]: string };
```
bdimitrijoski: @haakonflatval-cognite Maybe it is better if you create new type for key value pairs and reuse it. 

Ex:
```JS
export interface KeyValuePair {
[key: string]: string;
}

export interface RevisionMetadataUpdateAddRemove {
add: KeyValuePair;
}
```

Also I see that the metadata is optional in `SecurityCategory`. Do you need  `add` and `remove` here to be optional also?

polomani (approved):

packages/stable/src/types.ts
haakonflatval-cognite: I've added RevisionMetadata as its own type

PR #610, opened 2021-07-12
feat: add api version parameter to BaseCogniteClient [release]

andersfylling:
I'm interested in making a playground package, but I cannot do so as the `v1` version is hardcoded. This change allows package devs to specify api version as a constructor parameter for BaseCogniteClient. This change is backward compatible.

See example for playground implementation here: https://github.com/cognitedata/cognite-sdk-js/pull/611


bdimitrijoski (approved):
Nice!

PR #611, opened 2021-07-13
feat: add playground package [release]

andersfylling:
Playground package for cdf. I added two common operations for /documents to validate that it's working.

I am unfamiliar with how the documentation test works and I would highly appreciate some feedback on that.

Edit: Note I removed test-snippets and extract-snippets as I'm not certain how this will function with the playground. Right now the playground client is named differently, to make the package separation obvious.


packages/playground/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "build": "yarn clean && rollup -c && yarn esCheck",
+    "watch": "rollup -cw",
+    "esCheck": "es-check es5 './dist/index.js'",
+    "docs": "typedoc --options typedoc.js --tsconfig tsconfig.json src/index.ts",
+    "docs:bundle": "yarn docs && mkdir -p ../../docs/playground && cp -r docs/* ../../docs/playground/",
+    "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js --package-name @cognite/sdk-playground",
+    "test-snippets": "yarn extract-snippets && yarn tsc -p codeSnippets/tsconfig.build.json"
+  },
+  "dependencies": {
+    "@cognite/sdk": "^4.3.0",
```
polomani: you are not actually using it? it seems like you only need sdk-core

andersfylling: I use it in the types.ts file

polomani:
you have a dead link (as the page doesn't exist yet)
```
ERROR: 1 dead links found!
[✖] https://cognitedata.github.io/cognite-sdk-js/playground/classes/_beta_src_cogniteclient_.cogniteclient.html 
```

packages/playground/src/cogniteClientPlayground.ts
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Cognite AS
+
+import { accessApi, ClientOptions, BaseCogniteClient } from '@cognite/sdk-core';
+import { version } from '../package.json';
+import { DocumentsAPI } from './api/documents/documentsApi';
+
+export default class CogniteClientPlayground extends BaseCogniteClient {
```
bdimitrijoski, line 7: @andersfylling Why not just `export class CogniteClientPlayground`? You are exporting it anyway with that name in index.ts.

packages/playground/src/cogniteClientPlayground.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  protected get version() {
+    return `${version}-playground`;
+  }
+
+  protected initAPIs() {
+    super.initAPIs();
+
+    // Lock version to the following date
+    this.httpClient.setDefaultHeader('cdf-version', 'V20210406');
```
bdimitrijoski: @andersfylling Just out of curiosity, why do you set this to hardcoded value? Also maybe this can be exported as constant and then use it here.

packages/playground/src/cogniteClientPlayground.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.documentsApi = this.apiFactory(DocumentsAPI, 'documents');
+  }
+  static urlEncodeExternalId(externalId: string): string {
```
bdimitrijoski: @andersfylling Not sure why do you need this and where do you use it.

bdimitrijoski (approved):
LGTM

packages/playground/src/cogniteClientPlayground.ts
andersfylling: I'm unaware why, but this is the same setup as the beta and stable package.

packages/playground/src/cogniteClientPlayground.ts
andersfylling: Well spotted! I missed this unused method.

packages/playground/src/cogniteClientPlayground.ts
andersfylling: This seems to be related to https://cognitedata.atlassian.net/wiki/spaces/CDF/pages/2979397841/Experimental+feature+access+in+a+tenant and in this context should be removed.

PR #613, opened 2021-07-14
feat(playground): add document preview [release]

andersfylling:
~blocked by: https://github.com/cognitedata/cognite-sdk-js/pull/611~


packages/playground/src/api/documents/documentsApi.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface DocumentsAggregatesResponse<T> extends ItemsWrapper<T> {
+  aggregates?: DocumentsAggregate[];
+}
+
+export class DocumentsAPI extends BaseResourceAPI<Document> {
+  private readonly previewAPI: PreviewAPI;
+
+  constructor(...args: [string, CDFHttpClient, MetadataMap]) {
+    super(...args);
+    this.previewAPI = new PreviewAPI(args[0] + '/preview', args[1], args[2]);
```
bdimitrijoski: Maybe you can destructure the array here. It will be more readable when working with in future.
```JS
const [baseUrl, arg1Name, arg2Name] = args;
```

packages/playground/src/api/documents/documentsApi.ts
```diff
@@ -0,0 +1,56 @@
...
+  public search = (
+    query: ExternalDocumentsSearch
+  ): Promise<DocumentsAggregatesResponse<Document>> => {
+    return this.searchDocuments<DocumentsAggregatesResponse<Document>>(query);
+  };
+
+  public list = (
+    scope?: DocumentsRequestFilter
+  ): CursorAndAsyncIterator<Document> => {
+    return super.listEndpoint(this.callListEndpointWithPost, scope);
```
bdimitrijoski: Maybe you should use `this` here instead of `super`?
```JS
return super.listEndpoint(this.callListEndpointWithPost, scope);
``

File: packages/playground/src/api/documents/previewApi.ts [hunk 3]
```diff
@@ -0,0 +1,68 @@
+// Copyright 2020 Cognite AS
+
+import { BaseResourceAPI } from '@cognite/sdk-core';
+import { HttpResponseType } from '@cognite/sdk-core';
+
+import { DocumentId, DocumentPreviewTemporaryLink } from '../../types';
+import Any = jasmine.Any;
```
bdimitrijoski, line 7: @andersfylling I guess you don't need this here :)

packages/playground/src/api/documents/previewApi.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export class PreviewAPI extends BaseResourceAPI<Any> {
+  public documentAsPdf = (id: DocumentId): Promise<ArrayBuffer> => {
```
bdimitrijoski, line 10: @andersfylling I am not familiar with the coding conventions used in the sdk, but IMO we shoudn't use `public` since by default class functions are public anyway.
Also I would change functions to just plain functions instead of property = function.

```JS
documentAsPdf(id: DocumentId): Promise<ArrayBuffer>  {
//...code here
}
```

polomani: @bdimitrijoski regarding arrow functions: we have it that way in other APIs, makes sense to keep consistency

packages/playground/src/api/documents/documentsApi.ts
andersfylling: nice one, updated the code!

packages/playground/src/api/documents/previewApi.ts
andersfylling: I think this is required for the parent class decl: `PreviewAPI extends BaseResourceAPI<Any>`

packages/playground/README.md
```diff
@@ -13,13 +13,20 @@ or with npm
...
+>>>>>>> master
 ```js
 import { CogniteClientPlayground } from '@congite/sdk-playground';
 ```
 
+<<<<<<< HEAD
+=======
 The CogniteClientPlayground can be initialized/configured in the same manner as the other packages (eg. stable, beta, etc.).
 
+>>>>>>> master
```
maaland: merge conflict annotations :p

packages/playground/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -27,6 +29,53 @@ describe('documents api', () => {
       },
       limit: 1,
     });
-    expect(response.items.length).toEqual(1);
+    expect(response.items).toHaveLength(1);
+  });
+
+  describe('documents preview', () => {
```
maaland: nit:

```suggestion
  describe('document preview', () => {
```

packages/playground/src/api/documents/documentsApi.ts
```diff
@@ -36,6 +42,10 @@ export class DocumentsAPI extends BaseResourceAPI<Document> {
     return super.listEndpoint(this.callListEndpointWithPost, scope);
```
maaland: Agree with @bdimitrijoski that you should use `this` here instead of `super`

maaland (changes requested):
Some minor comments. Also please take a second pass at the other comments made to see if you want to address any of them.

packages/playground/README.md
andersfylling: oh wow

maaland (approved):
Looks good! :)

PR #616, opened 2021-07-19
feat: expose id token of AAD/ADFS/OIDC in one method [release]

vegardok:
Create a method equal to `getAzureADAccessToken` that works for all flows, where it is supported

### DIFF
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -442,6 +442,45 @@ export default class BaseCogniteClient {
   }
 
   /**
+   * Returns IDP id token
+   *
+   * ```js
+   * client.loginWithOAuth({ cluster: 'bluefield', ... });
+   * await client.authenticate();
+   * const idToken = await client.getIdToken();
+   * ```
+   */
+  public async getIdToken(): Promise<string | null> {
+    switch (this.flow!.type) {
+      case 'AAD_OAUTH': {
+        if (this.azureAdClient) {
+          return this.azureAdClient.getAccountToken();
+        } else {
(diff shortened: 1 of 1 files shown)


jarlah (approved):
just a new method right? looks good to me. I should remember to add client credentials here too, even if its also just returning Promise.reolsve(null), or should I return access_token from client credentials? even so this looks good.

packages/core/src/baseCogniteClient.ts
```diff
@@ -441,6 +441,45 @@ export default class BaseCogniteClient {
...
+   * const idToken = await client.getIdToken();
+   * ```
+   */
+  public async getIdToken(): Promise<string | null> {
+    switch (this.flow!.type) {
+      case 'AAD_OAUTH': {
+        if (this.azureAdClient) {
+          return this.azureAdClient.getAccountToken();
+        } else {
+          return Promise.resolve(null);
```
BugGambit, line 459: Why do it like this and not just `return null`?

packages/core/src/baseCogniteClient.ts
```diff
@@ -441,6 +441,45 @@ export default class BaseCogniteClient {
...
+        if (this.adfsClient) {
+          return this.adfsClient.getIdToken();
+        } else {
+          return Promise.resolve(null);
+        }
+      }
+      case 'OIDC_AUTHORIZATION_CODE_FLOW': {
+        if (this.authCodeFlowManager) {
+          const user = await this.authCodeFlowManager.getUser();
+          return Promise.resolve(user.id_token);
```
BugGambit, line 472: Same for this. Why not `return user.id_token;`?

PR #619, opened 2021-07-21
fix: silentlogin added for oidc

MyBitBird:


packages/core/src/authFlows/oidc_auth_code_flow.ts
```diff
@@ -101,7 +101,14 @@ export class OidcAuthCode {
             title: window.document.title,
           })
         );
-        await this.userManager.signinRedirect(this.options.loginParams);
+
+        const prompt = this.options.loginParams
```
BugGambit: Tip: use optional chaining: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

`const prompt = this.options.loginParams?.prompt ?? ''`

MyBitBird: Nice, prettier doesn't allow us to use 
`const prompt = this.options.loginParams?.prompt ?? ''`

but we can probably use:
`const prompt =  this.options.loginParams && this.options.loginParams.prompt;`

polomani (approved):

PR #620, opened 2021-07-21
feat: add custom login flow method [release]

jarlah:
with this PR we can do this:

```typescript
    await sdk.loginWithCustom(async (callbacks) => {
      return await new OidcClientCredentials(
        {} as any, // fill in  whatever required stuff is needed for this to work
        callbacks
      ).init();
    })
```

where callbacks is 

```typescript
    type Callbacks = {
        setCluster: (s: string) => void;
        setBearerToken: (s: string) => void;
        validateAccessToken: (s: string) => Promise<boolean>;
    }
```

this way we keep server side nodejs oidc impls outside js sdk

### DIFF
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -81,12 +81,17 @@ export type OIDC_AUTHORIZATION_CODE_FLOW = {
   type: 'OIDC_AUTHORIZATION_CODE_FLOW';
   options: OAuthLoginForOIDCAuthFlowOptions;
 };
+export type CUSTOM = {
+  type: 'CUSTOM';
+  options: unknown;
+};
 
 export type AuthFlowType =
   | AAD_OAUTH
   | CDF_OAUTH
   | ADFS_OAUTH
-  | OIDC_AUTHORIZATION_CODE_FLOW;
+  | OIDC_AUTHORIZATION_CODE_FLOW
+  | CUSTOM;
 
 /**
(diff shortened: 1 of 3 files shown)


packages/core/src/baseCogniteClient.ts
```diff
@@ -81,12 +81,17 @@ export type OIDC_AUTHORIZATION_CODE_FLOW = {
   type: 'OIDC_AUTHORIZATION_CODE_FLOW';
   options: OAuthLoginForOIDCAuthFlowOptions;
 };
+export type CUSTOM = {
+  type: 'CUSTOM';
+  options: unknown;
```
jarlah, line 86: not really used at all. Just for types to align really. the `options: unknown` I mean

packages/core/src/baseCogniteClient.ts
```diff
@@ -365,6 +370,42 @@ export default class BaseCogniteClient {
...
+        setBearerToken: (s: string) => void;
+        validateAccessToken: (s: string) => Promise<boolean>;
+      }
+    ) => Promise<OAuthLoginResult>
+  ): Promise<boolean> {
+    if (this.hasBeenLoggedIn) {
+      throwReLogginError();
+    }
+
+    this.flow = { type: 'CUSTOM', options: {} };
```
jarlah, line 386: needed to avoid null checking this.flow every place its used

vegardok (approved):

PR #625, opened 2021-08-02
test: do not run tests for irrelevant packages

andersfylling:
Yarn allows us to inject some lerna parameters. I figured we could add `--since master` to only run tests for packages that differ from the master branch. Say I'm adding something to beta, this change will then only validate the tests for the beta packages instead of core, stable, wells, etc.

A change in core will still trigger tests for all packages, as core is a dependency.

However, for the codecov all tests for all packages are still executed.

Added fetch-depth due to: https://github.com/actions/checkout/issues/118#issuecomment-743979327

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -99,4 +99,4 @@ jobs:
         run: yarn build
 
       - name: Test
-        run: yarn test
+        run: yarn test --since master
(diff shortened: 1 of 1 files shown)


andersfylling:
This will also help ease some pain of the unstable integration tests here and there.

andersfylling:
It should be possible to reduce the codecov time as well. Maybe `lerna ls --since master` can be used, but I need to experiment further. Removed it for now.

queCognite (approved):

bdimitrijoski (approved):

PR #627, opened 2021-08-02
fix(core): headers set by the sdk must respect case insensitive keys

andersfylling:
I noticed that the "Accept" field is added when it is not already present in the request header. However, the check is case sensitive, while the http RFC states that the http header are case insensitive.

This causes the `application/json` to be added when a user specify `accept` and not `Accept`:
```ts

    nock(mockBaseUrl)
      .get(new RegExp('/documents/preview'))
      .matchHeader('Accept', 'image/png')
      .query({ documentId: 1 })
      .once()
      .reply(200);

await this.get<ResponseType>(uri, {
      responseType: HttpResponseType.ArrayBuffer,
      headers: {
        accept: 'image/png',
      },
    });
```

The above test will fail as the accept header contains both `image/png` and `application/json`.

### DIFF
--- a/packages/core/src/__tests__/httpClient/basicHttpClient.unit.spec.ts
+++ b/packages/core/src/__tests__/httpClient/basicHttpClient.unit.spec.ts
@@ -51,8 +51,8 @@ describe('basicHttpClient', () => {
         'application/json'
       );
       expect(lengthOfHttpHeaders(alteredEmptyHeaders)).toEqual(1);
-      expect('Accept' in alteredEmptyHeaders).toBeTruthy();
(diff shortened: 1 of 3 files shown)


tommythorsen:
Does this change cause all header names to be lowercased, regardless of what case the users provide them in? If so, I'm not sure it's a good idea.

The ideal implementation would be case-insensitive, but case-preserving at the same time. Meaning that comparisons would be case-insensitive, but the original casing that the user provided the header names in, would be preserved.

andersfylling:
I do see that debugging may be more difficult if the user input is transformed. I updated the code :+1:

tommythorsen (approved):

PR #628, opened 2021-08-03
fix(core): remove slash suffix from uris [release]

andersfylling:
Sometimes the requests urls ends with `/`. This becomes a problem when using GET parameters:
`https://api.cognitedata.com/api/playground/projects/cognitesdk-js/documents/feedback/?status=CREATED`

While `/documents/feedback/?status=CREATED` responds with a 404, `/documents/feedback?status=CREATED` responds with an expected 200 status code. 

image
image


andersfylling:
To make a more unbiased argument; The RFC states that uri/path segments are to be separated by slashes. Given that get parameters aren't a segment, there should not be a slash after the path, separating the path and get parameters.

tommythorsen (approved):

PR #629, opened 2021-08-03
chore: automatic release on fix and feat

andersfylling:
Whenever a fix or a feature is added, this change will trigger the release mechanism.

Note! I'm not certain if wrapping the conditions in parenthesis is valid.


ebrattli (approved):
LGTM!

PR #639, opened 2021-08-10
test: remove duplicate testing

andersfylling:
Right now both the "lint" job and the "tests" job execute tests for node 14.x. Updated the workflow to only run node 14.x for the "lint" job, while the "tests" executes 10.x and 12.x.


jarlah (approved):
This is OK.

PR #644, opened 2021-08-11
chore: loginWithOAuth() needs SDK ver 5 now

tgib23:


jarlah (approved):
just samples? should be ok, but it wont be automatically released on chore, so if you want it published the current state of the PR wont affect much

tgib23:
@jarlah 
i thought samples are used by pulling from github, so it should be ok.
but let me know if i'm missing anything. i'm willing to fix more.

PR #645, opened 2021-08-18
fix(playground): wrong geo location relation enums [release]

andersfylling:


maaland (approved):
Looks fine, but why do you want them to be lowercase instead of uppercase?

andersfylling:
Our documents service uses lowercase enums, case sensitive. So using these will yield a 400 bad request.

PR #646, opened 2021-08-19
feat(functions): add functions for SDK

ebrattli:


packages/playground/src/api/functions/functionCallsApi.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  public list = async (
+    functionId: CogniteInternalId,
+    filter?: FunctionCallFilter,
+    limit?: number,
+    cursor?: string
+  ) => {
+    const path = this.url(`${functionId}/calls/list`);
+    return super.listEndpoint(
```
andersfylling: use this instead of super

packages/playground/src/api/functions/functionSchedulesApi.ts
```diff
@@ -0,0 +1,40 @@
...
+      this.listPostUrl,
+      {
+        data: { filter, limit },
+      }
+    );
+    return response.data;
+  };
+
+  public create = async (items: FunctionScheduleCreate[]) =>
+    super.createEndpoint(items);
```
andersfylling: - this instead of super
 - wrap function block in {}

packages/playground/src/api/functions/functionSchedulesApi.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  public delete = async (ids: CogniteInternalId[]) =>
+    super.deleteEndpoint(ids.map(id => ({ id })));
```
andersfylling: - this instead of super
 - wrap function block in {}

packages/playground/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,48 @@
...
+        data: { filter, limit },
+      }
+    );
+    return response.data;
+  };
+
+  public retrieve = (ids: IdEither[]) => super.retrieveEndpoint(ids);
+
+  public delete = async (ids: IdEither[]) => {
+    return super.deleteEndpoint(ids);
```
andersfylling: - this instead of super

packages/playground/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,48 @@
...
+    const response = await this.post<ItemsResponse<Function>>(
+      this.listPostUrl,
+      {
+        data: { filter, limit },
+      }
+    );
+    return response.data;
+  };
+
+  public retrieve = (ids: IdEither[]) => super.retrieveEndpoint(ids);
```
andersfylling: - this instead of super
 - wrap function block in {}

packages/playground/src/api/functions/functionSchedulesApi.ts
```diff
@@ -0,0 +1,40 @@
...
+import { BaseResourceAPI, CogniteInternalId } from '@cognite/sdk-core';
+import {
+  FunctionSchedule,
+  FunctionScheduleCreate,
+  FunctionSchedulesFilter,
+} from '../../types';
+
+export class FunctionSchedules extends BaseResourceAPI<FunctionSchedule> {
+  public retrieve = (ids: CogniteInternalId[], ignoreUnknownIds?: boolean) =>
+    super.retrieveEndpoint(ids.map(id => ({ id })), {
```
andersfylling: - this instead of super

packages/playground/src/__tests__/testFunction/handler.py
```diff
@@ -0,0 +1,3 @@
+def handle(data):
```
andersfylling, line 1: can you remove this? I don't think these print statements are required in the sdk

packages/playground/src/api/functions/functionCallsApi.ts
```diff
@@ -0,0 +1,74 @@
...
+        data: { data, nonce },
+      }
+    );
+    return response.data;
+  };
+
+  public list = async (
+    functionId: CogniteInternalId,
+    filter?: FunctionCallFilter,
+    limit?: number,
```
andersfylling, line 44: can you add a unit test that verifies that the query parameter is added?

packages/playground/src/api/functions/functionCallsApi.ts
ebrattli: ✅

packages/playground/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,48 @@
+import { BaseResourceAPI, CDFHttpClient, MetadataMap } from '@cognite/sdk-core';
+import { IdEither, ItemsResponse } from '@cognite/sdk';
+import { Function, FunctionCreate, FunctionFilter } from '../../types';
+import { FunctionCallsApi } from './functionCallsApi';
+import { FunctionSchedules } from './functionSchedulesApi';
+
+export class FunctionsAPI extends BaseResourceAPI<Function> {
```
andersfylling, line 7: wonder if you're missing a method: https://docs.cognite.com/api/playground/#operation/get-api-playground-projects-project-functions-function_name

This can probably be ignored

andersfylling (changes requested):
Majority is just for code consistency across the repo. To me it looks good, just verify the path parameters :+1:

packages/playground/src/__tests__/testFunction/handler.py
ebrattli: I test against the `Hello World!` in the logs test. And, we need something to do in the function anyways.

packages/playground/src/api/functions/functionsApi.ts
ebrattli: It is handled by retrieve, so I figured it is not needed.

packages/playground/src/api/functions/functionCallsApi.ts
ebrattli: list uses the filter endpoint. Thus there are only body parameters.

packages/playground/src/__tests__/testFunction/handler.py
andersfylling: ah, I see it's executed outside the sdk. my mistake

andersfylling (approved):
LGTM!

packages/playground/src/api/functions/functionSchedulesApi.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+    return response.data;
+  };
+
+  public create = async (items: FunctionScheduleCreate[]) => {
+    return this.createEndpoint(items);
+  };
+
+  public delete = async (ids: CogniteInternalId[]) => {
+    return this.deleteEndpoint(ids.map(id => ({ id })));
```
polomani, line 33: why doing this mapping? why can't just take `IdEither[]`?

polomani: (same for other places)

ebrattli: Retrieve on function schedules only takes `id`, and not `externalId`. So we don't want the user to send `externalIds` when it is not supported. https://docs.cognite.com/api/playground/#operation/byidsFunctionSchedules

PR #656, opened 2021-08-20
test: speed up tests

andersfylling:
Added the --parallel flag from lerna. This kinda crashed the majority of programs on my laptop, but i think for a dedicated CI it's perfect to max out whatever resources we have.


andersfylling:
Seems this did nothing for the overall test time

PR #657, opened 2021-08-20
fix(func-test): make function test more reliable

ebrattli:


andersfylling (approved):
Hope this jest feature works!

PR #660, opened 2021-08-20
feat(core): list azure tenants [BLOCKED]

deep-cognite:
Most of this code is copy-paste of the methods used for AAD CDF auth stuff. 

https://docs.microsoft.com/en-us/rest/api/resources/tenants/list

--

Status: BLOCKED (23.08): ref: https://cognitedata.slack.com/archives/C6KNJCEEA/p1629717852108600

### DIFF
--- a/packages/core/src/authFlows/aad.ts
+++ b/packages/core/src/authFlows/aad.ts
@@ -135,6 +135,14 @@ export class AzureAD {
     }
   }
 
+  async azureLoginRedirect(): Promise<AccountInfo | void> {
+    await this.msalApplication.loginRedirect({
+      scopes: this.userScopes,
+      extraScopesToConsent: this.getAzureScopes(),
+      redirectStartPage: window.location.href,
+    });
+  }
+
   /**
    * Logs out of current account.
    */
@@ -163,6 +171,19 @@ export class AzureAD {
     return accessToken;
   }
 
+  async getAzureToken(): Promise<string | null> {
+    if (!this.getAccount()) return null;
+
+    const {
+      accessToken,
+    }: AuthenticationResult = await this.msalApplication.acquireTokenSilent({
+      account: this.getAccount(),
+      scopes: this.getAzureScopes(),
+    });
+
+    return accessToken;
+  }
+
   /**
    * Returns azure account access token.
    * Can be used for getting user details via Microsoft Graph API
@@ -184,6 +205,10 @@ export class AzureAD {
       `https://${this.cluster}.cognitedata.com/user_impersonation`,
       `https://${this.cluster}.cognitedata.com/IDENTITY`,
     ];
+  }
+
+  private getAzureScopes(): string[] {
+    return ['https://management.azure.com/user_impersonation'];
   }
 
   private getLoginPopupRequest(
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -959,7 +959,7 @@ export default class BaseCogniteClient {
 
     let accessToken = null;
 
-    const account = this.validateAccount(azureAdClient);
+    const account = await this.validateAccount(azureAdClient);
     if (account) {
       accessToken = await azureAdClient.getAzureToken();
     }
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -127,6 +127,16 @@ export interface OAuthLoginForOIDCAuthFlowOptions extends OIDCAuthFlowOptions {
   onNoProjectAvailable?: () => void;
 }
 
+export interface AzureTenantInfo {
+  id: string;
+  defaultDomain: string;
+  displayName: string;
+  domains: string[];
+  tenantCategory: string;
+  tenantId: string;
+  tenantType: string;
+}
+
 export function accessApi<T>(api: T | undefined): T {
   if (api === undefined) {
     throw Error(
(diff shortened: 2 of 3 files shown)


packages/core/src/baseCogniteClient.ts
```diff
@@ -911,18 +921,72 @@ export default class BaseCogniteClient {
...
+      },
+    };
+    const azureAdClient = new AzureAD({ config, cluster: '', debug });
+
+    let accessToken = null;
+
+    const account = this.validateAccount(azureAdClient);
+    if (account) {
+      accessToken = await azureAdClient.getAzureToken();
+    }
```
deep-cognite: No idea if this is needed, followed the AD flow above.

packages/core/src/baseCogniteClient.ts
```diff
@@ -911,18 +921,72 @@ export default class BaseCogniteClient {
...
+  }
+
+  /**
+   * Disclaimer: This only works with AAD.
+   *
+   * Fetches the current users tenant list (https://docs.microsoft.com/en-us/rest/api/resources/tenants/list#code-try-0)
+   * @param clientId
+   * @param debug
+   */
+  public async getTenantList(
```
deep-cognite: rename to
```suggestion
  public async getAzureTenantList(
```
...?

packages/core/src/baseCogniteClient.ts
```diff
@@ -911,18 +921,72 @@ export default class BaseCogniteClient {
...
+      accessToken = await azureAdClient.getAzureToken();
+    }
+
+    if (!accessToken) {
+      return [];
+    }
+
+    const http = new CDFHttpClient(
+      'https://management.azure.com/tenants?api-version=2020-01-01',
+      this.getRetryValidator()
```
deep-cognite, line 979: Not sure if that should be included, here.

deep-cognite: Used `CDFHttpClient` since that was nicely setup for http requests.

ngerstle-cognite:
@deep-cognite 
Does this require `user_impersonation` for the Azure AD management api?
What does the consent prompt specify?

(I'm not sure we can do this if so- I don't think the customers would be happy to issue tokens with potential admin access to their azure subscription/ad just to list tenants..)

deep-cognite:
> @deep-cognite
> Does this require `user_impersonation` for the Azure AD management api?
> What does the consent prompt specify?
> 
> (I'm not sure we can do this if so- I don't think the customers would be happy to issue tokens with potential admin access to their azure subscription/ad just to list tenants..)

Yeah, we need `user_impersonation` in order to get the tenants list: https://docs.microsoft.com/en-us/rest/api/resources/providers/list-at-tenant-scope#scopes

Not sure about the consent window, forgot to take a screenshot and I don't know how to get it back now.

I did some research, and there are more people that have the same concern, @ngerstle-cognite: https://docs.microsoft.com/en-us/answers/questions/442040/list-tenants-for-current-user-without-user-imperso.html
^ This was newly posted, as well. Hopefully the MS Graph team can change the scope to User.Read or something 🙈

jarlah:
What about auth0? Other custom idps? I dont think this should live in the js sdk

jarlah:
We can create a new package that extends the functionality maybe. Extends the base client.

ngerstle-cognite:
> > @deep-cognite
> > Does this require `user_impersonation` for the Azure AD management api?
> > What does the consent prompt specify?
> > (I'm not sure we can do this if so- I don't think the customers would be happy to issue tokens with potential admin access to their azure subscription/ad just to list tenants..)
> 
> Yeah, we need `user_impersonation` in order to get the tenants list: https://docs.microsoft.com/en-us/rest/api/resources/providers/list-at-tenant-scope#scopes
> 
> Not sure about the consent window, forgot to take a screenshot and I don't know how to get it back now.
> 
> I did some research, and there are more people that have the same concern, @ngerstle-cognite: https://docs.microsoft.com/en-us/answers/questions/442040/list-tenants-for-current-user-without-user-imperso.html
> ^ This was newly posted, as well. Hopefully the MS Graph team can change the scope to User.Read or something see_no_evil

If this functionality requires that scope, it's a no-go for most customers (I wouldn't be comfortable granting that as an AAD admin myself).

ngerstle-cognite:
> What about auth0? Other custom idps? I dont think this should live in the js sdk

I do think there's a valid question around specialization. To this point, we've stated that we'll support AAD, and other IdPs meeting the requirements, and that we'll only provide guides/tutorials/native support for AAD. I think it's reasonable to add AAD specific code for retrieving a token to the SDK, as long as it's not required (an alternative call to use another IdP).
It would be fantastic to have a solid OAuth2/OIDC client that's standards oriented, and not AAD specific.


I'd also point out that this functionality is relevant for user sign in, but this SDK may be used for other clients (this is a specific request- how to use the JS sdk with backend apis/scripts/etc)- in that sense, a different package/product might be best (no one else has made a lib to handle this already?).

rek:
> What about auth0? Other custom idps? I dont think this should live in the js sdk
this sdk is full of Azure+AAD things. when there are requirements for other idps... someone can do the work to support them.

> but this SDK may be used for other clients 
again yes it might, but failing having a team (or even a person) to support this sdk, we can only contribute the code we have a requirement for

this part of the sdk already deals with a lot of similar things, adding this extra method here seems very inline with the current methodology here.

packages/core/src/authFlows/aad.ts
```diff
@@ -186,6 +207,10 @@ export class AzureAD {
     ];
   }
 
+  private getAzureScopes(): string[] {
```
rek, line 210:
```suggestion
  private getAzureManagementScopes(): string[] {
```

packages/core/src/authFlows/aad.ts
```diff
@@ -163,6 +171,19 @@ export class AzureAD {
...
   }
 
+  async getAzureToken(): Promise<string | null> {
+    if (!this.getAccount()) return null;
+
+    const {
+      accessToken,
+    }: AuthenticationResult = await this.msalApplication.acquireTokenSilent({
+      account: this.getAccount(),
+      scopes: this.getAzureScopes(),
```
rek, line 181:
```suggestion
      scopes: this.getAzureManagementScopes(),
```

packages/core/src/authFlows/aad.ts
```diff
@@ -163,6 +171,19 @@ export class AzureAD {
     return accessToken;
   }
 
+  async getAzureToken(): Promise<string | null> {
```
rek, line 174:
```suggestion
  async getAzureManagementToken(): Promise<string | null> {
```

PR #667, opened 2021-08-31
chore: update readme

andersfylling:


ebrattli:
why?

andersfylling:
No need for both of them. The `--since master` will always execute tests for all affected packages. In essence only executing the fewest tests possible to verify changes. While running all tests may only add redundant information on top of using `--since master`.

README.md
```diff
@@ -43,16 +43,7 @@ This repo contains some integration tests that require a CDF api key for `cognit
...
-Run all tests:
-
-```bash
-yarn
-yarn build
-yarn test
-```
-
-Run tests for updated package only:
-> for changes to the core package, it might be required to run tests for all packages to verify integrity.
```
ebrattli, line 55: This might be useful to know.

andersfylling: if the core has changed, then any other packages that depends on core are marked "dirty". So all tests will be run.

ebrattli (approved):
Ok. LGTM!

PR #668, opened 2021-08-31
test: separate lint and codecov

andersfylling:
Significantly improve the PR feedback by:
 - splitting tests into respective categories
 - reducing overall test time by using multiple CIs instead of one big one for codecov

longest running CI went down from 7-10min to 4min

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -7,8 +7,51 @@ on:
       - release-*
 
 jobs:
-  lint:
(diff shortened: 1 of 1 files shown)


ebrattli (approved):
lgtm. I know the setup-node action has an option for caching that you can use if you wish. Then you wouldn't need a separate step for that.

andersfylling:
won't do this here, but would be a great change to reduce the verbosity of these CI.

PR #680, opened 2021-09-15
feat(spatial): adds spatial API to playground

saagar-cognite:
- adds spatial API to playground package
- composes features & featureTypes API inside spatial API


packages/playground/src/api/spatial/types.ts
```diff
@@ -0,0 +1,139 @@
...
+export interface FeatureSearchFilter extends Limit {
+  filter: OrFilter & AndFilter & NotFilter & AllFilters;
+}
+
+interface AttributeValue {
+  attribute: string;
+  value: unknown;
+}
+
+interface AllFilters
```
pawelzwolenskicognite: can we work with json definitions for filters? if we add a new filter on the backend now we would need to change JS SDK too. If it was json-based, nothing would change here.

pawelzwolenskicognite: or maybe js map instead of json - I don't know what is better way to do that in JS (I think in Python we used dictionary, is that right @vincent-cognite ?)

saagar-cognite: @pawelzwolenskicognite I don't understand what json based means ? 🤔 
Do you want it to be accepting of any new type of filter being added with the json format of attribute value pair ?

pawelzwolenskicognite: json based: instead of using TC classes that you defined below, we use json or map (like here in python SDK: https://github.com/cognitedata/cognite-sdk-python-experimental/blob/13aff08271ff381ad91856a25e7f5afac53ff769/cognite/experimental/_api/geospatial.py#L241) - so user can send any json/map and we don't need to hardcode TS classes for new filters

(there is validation of filters on server side, so if json is wrong, user will get 400)

pawelzwolenskicognite: to be consistent with Python SDK it would be nice to use Map<String, Any> as a general filter and do not have any custom classes

@vincent-cognite can you share your opinion on that? :)

saagar-cognite: I can also add a generic Map<string, any> within filter and keep these existing ones as well.
This would allow user to have autocomplete with all possible existing options and in case someone wishes to add a custom filter, they won't face any error and anyways a 400 would save the day at the end.

@pawelzwolenskicognite @vincent-cognite

pawelzwolenskicognite: I would rather replace all the filters with Map<String, Any> so SDK will not require to be updated each time we add new filter type support on backend

rek: it seems way safter (better dx) to list the filters here explicitly and update the sdk when the backend changes. 
how frequently would happen anyway?

packages/playground/src/api/spatial/featuresAPI.ts
```diff
@@ -0,0 +1,66 @@
...
+    return this.createEndpoint<FeaturesCreateItem>(
+      features,
+      this.url(`${featureTypeExternalId}/features`)
+    );
+  };
+
+  public retrieve = (
+    featureTypeExternalId: CogniteExternalId,
+    ids: IdEither[],
+    queryParams?: { outputGeometryFormat: 'wkt' | 'geojson' }
```
rek: it feels like these should come from a type somewhere

packages/core/src/baseResourceApi.ts
```diff
@@ -174,8 +175,14 @@ export abstract class BaseResourceAPI<ResponseType> {
     );
   }
 
-  protected async searchEndpoint<FilterType>(query: FilterType) {
-    return this.callEndpointWithTransform(query, this.callSearchEndpoint);
+  protected async searchEndpoint<FilterType>(
+    query: FilterType,
+    path?: string,
```
andersfylling, line 180: can you set the default value to `this.callSearchEndpoint` to avoid breaking endpoints.

andersfylling: though it seems perhaps the callEndpointWithTransform uses the default value.

packages/playground/src/api/spatial/featuresAPI.ts
saagar-cognite: https://github.com/cognitedata/cognite-geospatial-sdk/blob/89b0eda90aab74086e47a9f9b993acfbf1f3e0aa/nodejs/src/types.ts#L24
I'll create a GeometryType like this one present in current geospatial-sdk.
Is that fine @rek ?

packages/core/src/baseResourceApi.ts
saagar-cognite: @andersfylling the support for path & queryParams is added in an optional manner.
Not sure if this breaks any end point.
Can you please help me with what change are you looking for exactly ?
callEndpointWithTransform doesn't have any default value and is not touched in this PR.

saagar-cognite: Did you mean the default path value `this.searchUrl` in `callSearchEndpoint` ?

packages/playground/src/api/spatial/types.ts
saagar-cognite: Keeping it this way for better DX.
We can change this later in case the changes become quite frequent to types filters.

packages/core/src/baseResourceApi.ts
andersfylling: looks good!

andersfylling (approved):

packages/stable/src/api/spatial/types.ts
```diff
@@ -0,0 +1,141 @@
+import { CogniteExternalId, CogniteInternalId, Limit } from '@cognite/sdk-core';
+
+export type GeometryType = 'wkt' | 'geojson';
+
+export type AttributeType =
```
andersfylling, line 5: I think you have to export these to the main types.ts file so others can access them in a consistent manner.

andersfylling (changes requested):

packages/stable/src/api/spatial/types.ts
saagar-cognite: exported after importing from this file from src/types.ts as discussed on slack thread.
@andersfylling

andersfylling (approved):

packages/stable/src/api/spatial/types.ts
```diff
@@ -0,0 +1,141 @@
...
+  | 'MULTILINESTRING'
+  | 'MULTIPOLYGON'
+  | 'GEOMETRYZ'
+  | 'POINTZ'
+  | 'LINESTRINGZ'
+  | 'POLYGONZ'
+  | 'MULTIPOINTZ'
+  | 'MULTILINESTRINGZ'
+  | 'MULTIPOLYGONZ'
+  | 'GEOMETRYCOLLECTION';
```
pawelzwolenskicognite, line 25: @saagar-cognite we are supporting more types (M and ZM geometries)
the list can be found in documentation: https://pr-1323.specs.preview.cogniteapp.com/v1.json.html#operation/createFeatureTypes

PR #687, opened 2021-10-07
feat(core): move authentication out of CogniteClient

vegardok:
BREAKING CHANGE: remove authentication internals in `CogniteClient` and replace it with an improved api to allow Apps handle authentication.

Change summarized:
*  `loginWithMETHOD` have been removed and replaced with an mandatory `getToken` argument to the `CogniteClient` constructor
* `setProject`/`setBaseUrl` removed
* `authenticate` return type have been changed from `() => Promise<boolean>` to `() => Promise<string|undefined>`, returning the token if found.

[AzureAD example using MSAL](https://github.com/cognitedata/cognite-sdk-js/tree/v6/samples/react/msal-browser-react)
[Legacy auth SPA example](https://github.com/cognitedata/cognite-sdk-js/blob/v6/samples/react/legacy-auth/src/App.tsx)
[API Key example](https://github.com/cognitedata/cognite-sdk-js/blob/v6/samples/typescript/quickstart.ts)

Motivation of change:
* The application _is_ the entity that is logging in in a OAuth perspective, not the SDK
* The SDK only cares about access tokens valid for CDF, but an app could care about access to other resources (e.g a separate app backend or IDP graph API). It is difficult/impossible to handle this with the current "internal auth" handling
* CDF have different [access scopes](https://docs.cognite.com/cdf/access/concepts/access_token_scopes.html) and only the app can know what is the appropriate scope for a certain feature/context. Handling tokens outside the SDK will enable apps to control this


andersfylling:
I haven't setup oidc yet, so might be we will see some unexpected 401. Uncertain if they enforced it for the api cluster yet.

andersfylling (approved):
LGTM!

guides/derived_SDKs.md
```diff
@@ -132,7 +132,8 @@ export class CoolThingAPI extends BaseResource<CoolThing> {
 }
 ```
 
-The `CogniteClient` class makes instances of these classes in `initAPIs()`, which is called once credentials and project name are set. The accessor helper `accessApi()` reminds the user to set credentials before using APIs.
+The `CogniteClient` class makes instances of these classes in `initAPIs()`, which in the class
```
jarlah: typo, dont understand. did you mean "which is called from the constructor"?

jarlah: or "which is in the class constructor". but then _what_ is in the class constructor?

packages/stable/README.md
```diff
@@ -66,10 +87,10 @@ quickstart();
 const { CogniteClient } = require('@cognite/sdk');
 
 async function quickstart() {
-  const client = new CogniteClient({ appId: 'YOUR APPLICATION NAME' });
-  client.loginWithApiKey({
-    project: 'publicdata',
-    apiKey: 'YOUR_SECRET_API_KEY',
+  const client = new CogniteClient({
+    appId: 'YOUR APPLICATION NAME',
+    apiKeyMode: true,
```
jarlah, line 92: it works i guess :) only two modes, either api key or bearer token.

jarlah (approved):
Added some comments but LGTM

bitoverflow:
When baseURL is provided to the constructor, getToken is not called. And authorization header is not added to requests.

Testing on @cognite/sdk version 6.1.1

PR #688, opened 2021-10-09
feat(documents)!: update pipelines request/response [skip-ci]

andersfylling:
BREAKING CHANGE: pipeline config now takes an array of strings for field mappings and the metadata in document source

https://cognitedata.atlassian.net/browse/SBS-4181

### DIFF
--- a/packages/playground/src/types.ts
+++ b/packages/playground/src/types.ts
@@ -254,19 +254,19 @@ export interface SensitivityMatcher {
 }
 
 export interface DocumentsFieldMappings {
-  title?: string[];
-  author?: string[];
-  mimeType?: string[];
-  type?: string[];
+  title?: string | string[];
+  author?: string | string[];
+  mimeType?: string | string[];
+  type?: string | string[];
   labelsExternalIds?: string[];
   sourceFile?: DocumentsSourceFile;
 }
 
 export interface DocumentsSourceFile {
-  name?: string[];
-  directory?: string[];
-  content?: string[];
-  metadata?: StringToStringArrayMap;
+  name?: string | string[];
+  directory?: string | string[];
+  content?: string | string[];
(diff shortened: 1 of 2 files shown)


packages/playground/src/types.ts
```diff
@@ -255,23 +250,23 @@ export interface SensitivityMatcher {
...
 
 export interface DocumentsFieldMappings {
-  title?: string;
-  author?: string;
-  mimeType?: string;
-  type?: string;
+  title?: string | string[];
+  author?: string | string[];
+  mimeType?: string | string[];
+  type?: string | string[];
```
arnels: The API will always return `string[]`, but support both `string` and `string[]` in requests. The API supports both on request to be backwards compatible, but we want to phase out the `string` option. So, can they be only `string[]` here?

packages/playground/src/types.ts
```diff
@@ -255,23 +250,23 @@ export interface SensitivityMatcher {
...
 
 export interface DocumentsSourceFile {
-  name?: string;
-  directory?: string;
-  content?: string;
-  metadata?: StringToStringMap;
+  name?: string | string[];
+  directory?: string | string[];
+  content?: string | string[];
+  metadata?: StringToStringMap | StringToStringArrayMap;
```
arnels: The same goes here

packages/playground/src/types.ts
andersfylling: oh yeah

andersfylling:
@arnels given this is a breaking change i also pushed in the missing "set" field for pipelines config update. I know this is messy, but I hope this will keep the version history a bit cleaner.

Edit: seems this should not be part of a breaking change PR at all. the update method is in need of a patch.

arnels (approved):

PR #689, opened 2021-10-09
feat(documents): add classifier endpoints

andersfylling:
https://cognitedata.atlassian.net/browse/SBS-4176


packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
```diff
@@ -745,4 +745,50 @@ describe('Documents unit test', () => {
...
+          items: [{ name: 'test' }],
+        });
+      const resp = await client.documents.classifiers.create([
+        { name: 'test' },
+      ]);
+      expect(resp[0].name).toEqual('test');
+    });
+    test('list by ids', async () => {
+      nock(mockBaseUrl)
+        .post(new RegExp('/documents/classifiers'), {
```
henrist: I think it should have `/byids` at the end here? https://docs.cognite.com/api/playground/#operation/documentsListClassifiersByids

If I understand https://github.com/nock/nock#specifying-path correctly I also think this RegExp pattern both matches `/documents/classifiers` as well as `/documents/classifiers/byds`. I'd recommend only using RegExp pattern where it makes sense to do so, so keeping a plain string in this case.

packages/playground/src/types.ts
```diff
@@ -75,6 +75,27 @@ export interface DocumentsSourceFileFilter {
   size?: Range<number>;
 }
 
+export interface ClassifierName {
+  name?: string;
```
henrist, line 79: Is there a reason we keep these as optional? I assume not sending the name would make the request fail when validated at the API.

For some reason I see that [the spec](https://storage.googleapis.com/cognitedata-api-docs/dist/playground.json) has not marked it as required, but actually have marked it as an enum only accepting `Document classifier` as value, so that's likely a bug. (See DocumentsClassifierCreate schema.)

queCognite: Classifier name when creating a classifier is optional. It has a default value if the name is not set.
Yes, the spec is wrong here. `Document classifier` should be default value instead of `enum`. I will fix it.

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
andersfylling: good catch!

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
```diff
@@ -5,6 +5,8 @@ import CogniteClientPlayground from '../../cogniteClientPlayground';
 import { setupMockableClient } from '../testUtils';
 import { mockBaseUrl, project } from '@cognite/sdk-core/src/testUtils';
 
+const baseURL = mockBaseUrl + `/api/playground/projects/${project}`;
```
henrist: Very nit as I know we are not consistent on this matter, and this is not important at all in a test like this, but I always prefer to keep only first char capitalized when there are acronyms in use. URL => Url

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
```diff
@@ -745,4 +747,50 @@ describe('Documents unit test', () => {
...
+        })
+        .once()
+        .reply(200, {
+          items: [{ id: 1 }, { id: 2 }, { id: 3 }],
+        });
+      const resp = await client.documents.classifiers.listByIds(
+        [1, 2, 3],
+        true
+      );
+      expect(resp.items[0].id).toEqual(1);
```
henrist, line 793: nit: Does the ordering even matter on the response, which is actually tested here (picking first item)? I'd think what you  _want_ to expect is that the collection contains the element, not its position.

packages/playground/src/api/documents/classifiersApi.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  public delete = (ids: InternalId[], ignoreUnknownIds: boolean = false) => {
+    return this.deleteEndpoint(ids, {
+      ignoreUnknownIds,
+    });
+  };
+
+  private async classifiersListByIds<ResponseType>(
+    ids: CogniteInternalId[],
+    ignoreUnknownIds?: boolean
```
henrist: nit: this is actually never undefined. Couldn't it be easier to just inline this method anyways?

packages/playground/src/types.ts
```diff
@@ -75,6 +75,27 @@ export interface DocumentsSourceFileFilter {
...
+  active?: boolean;
+  metrics: ClassifierMetrics;
+}
+
+export interface ClassifierMetrics {
+  precision?: number;
+  recall?: number;
+  f1Score?: number;
+  confusionMatrix: number[][];
+  labels: string[];
```
henrist: I'm confused why some of these are optional while others are not. They seem all optional according to the spec: https://github.com/cognitedata/service-contracts/blob/b36b07a61e23021b0d2a00f2e883cefbf407ba88/versions/playground/documents.yml#L1765-L1787

packages/playground/src/types.ts
```diff
@@ -75,6 +75,27 @@ export interface DocumentsSourceFileFilter {
...
+  name?: string;
+}
+
+export interface Classifier extends ClassifierName {
+  id: CogniteInternalId;
+  projectId?: number;
+  createdAt?: number;
+  status?: string;
+  active?: boolean;
+  metrics: ClassifierMetrics;
```
henrist: This should also be optional?

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
andersfylling: response order is required by the api design doc: https://github.com/cognitedata/service-contracts/blob/master/design_guide/API_design_guide.md#multi-get

packages/playground/src/api/documents/classifiersApi.ts
andersfylling: It's just to keep the code consistent with the rest of the SDK. You'll see that the public methods/endpoints are at the top of the resource class, and implementations in the bottom.

packages/playground/src/types.ts
andersfylling: There might have been some updates to the open api spec. Updated!

packages/playground/src/types.ts
andersfylling: updated!

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
andersfylling: updated!

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
henrist: Aha, I see.

So what does the ignoreUnknownIds actually do? According to your link, if some IDs are not found, it's not clear if it should have a partial response (null elements?) (but it then requires it to be http 400 error) or no response at all.

Should there be a test that covers the case where an ID is missing?

andersfylling: We might do tests for it in our service, but not in the sdk as we only want to provide a correct skeleton for end users to build queries and handle responses.

I haven't seen anything in the guideline whether or not the missing objects should be added as empty structs or removed all together from the response. Current implementation on our end simply do not include objects for unknown ids. So the response list length may be less than the request list length.

I believe this is the same behavior as the files api.

henrist: But I guess the SDK should be clear for how this works? E.g. how does the SDK handle the scenario where we get a 400 partial response - are we unable to read out the contents that might be present? (I'm not sure how it works, so not sure which cases we have.)

andersfylling: I believe that when using ignoreUnknownIds we will get a 200 response, instead of a 4xx.

The way you read data when the response is not a 2xx in the js sdk is by looking up the cached responses, as these are stored in a ever growing map:
image


The way to do that is:
```ts
client.getMetadata(response)
```

henrist (approved):
So this is outside the scope of this PR, but looking at the implementation of that byids endpoint it actually does not follow the requirements of the API design guide. It does not hold the guarantee that elements have the same position when ignore is set to true. The implementation also does not return elements in the same order.

I'd love to see both that solved and also this SDK to be more clear about the behaviour, since it's a bit magical now for those who uses the SDK.

packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
andersfylling: The SDK should mirror the cognite api docs perfectly, so any concerns and questions can we checked there. 

Though I am ofcourse open to suggestions on how the SDK can document this in an intuitive manner.

henrist: Aha, I see. Thanks for letting me know!

PR #705, opened 2021-10-18
fix(templates): retry templates requests [BEST-902]

peternic:
Added /templatesgroups to the retry endpoint list so that templates requests is retried if it is rate limited for instance.

The default response handler is JSON, and this fails on the templates responses, so the request fails and does is not retried.

So added a fallback to text on the jsonHandler if it fails by some reason

How templates fails now:
```
FetchError: invalid json response body at https://api.cognitedata.com/api/v1/projects/some-tenant/some-api-endpoint reason: Unexpected end of JSON input
    at /Users/someuser/projects/cognite-project/js/node_modules/node-fetch/lib/index.js:272:32
    at processTicksAndRejections (internal/process/task_queues.js:97:5) {
  type: 'invalid-json'
}```

### CODE REVIEW

**Comment by Tapped** [comment]
I think this is a bit different. This is about retrying on 500 and such, and since POST might not be idempotent only some endpoints can use this logic. However a 429 (Too many requests) is idempotent to retry regardless of method and endpoint. So I think the SDK should have special logic for handling 429 across any request being sent from the SDK.

**Review by Tapped — APPROVED** [review]
LGTM!

**Review by bdimitrijoski — APPROVED** [review]

**Comment by andersfylling** [comment]
It seems this change is causing the integration tests to become more flaky?

**Comment by peternic** [comment]
> It seems this change is causing the integration tests to become more flaky?

Can it get *more* flaky 😜  ?

On a more serious note, I will take a look a bit later.

**Comment by andersfylling** [comment]
> > It seems this change is causing the integration tests to become more flaky?
> 
> Can it get _more_ flaky stuck_out_tongue_winking_eye ?

:laughing: I have been cleaning up stuff and explicitly marked unstable tests to skip now (notified the responsible teams). There might be one or two more flaky tests, but it seem to frequently just work these days.

File: packages/core/src/httpClient/retryValidator.ts [hunk 1]
```diff
@@ -50,6 +50,10 @@ export const createUniversalRetryValidator = (
   if (retryCount >= maxRetries) {
     return false;
   }
+  // Always retry http 429 too many requests
+  if (response.status === 429) {
```
polomani, line 54:
```suggestion
```
This is already implemented, see 
```
inRange(status, 429, 430
```

polomani (changes requested):

peternic:
> > > It seems this change is causing the integration tests to become more flaky?
> > 
> > 
> > Can it get _more_ flaky stuck_out_tongue_winking_eye ?
> 
> 😆 I have been cleaning up stuff and explicitly marked unstable tests to skip now (notified the responsible teams). There might be one or two more flaky tests, but it seem to frequently just work these days.

Hmm, seems the clone() is not working in the tests. Any idea how to get around this?

peternic:
Found the issue @andersfylling, was that node-fetch (what cross-fetch uses under the hood) in versions < 3 will hang on large responses for clone on node (not for browsers).
This is fixed in v3, but no version of cross-fetch is using that.

So followed their guide how to handle this: https://github.com/node-fetch/node-fetch#custom-highwatermark

packages/core/src/httpClient/retryValidator.ts
```diff
@@ -50,6 +50,10 @@ export const createUniversalRetryValidator = (
...
+    return true;
```
andersfylling, line 55: Retry is nice. However, are we respecting `Retry-After` headers here?
Otherwise I think this might get stuck in a retry loop

peternic: Not any headers I think, but the maxRetries will hit before this.
(unless there are some headers taken into consideration in the normal retry logic we already have)

andersfylling (approved):

andersfylling (changes requested):
Wait, actually why can't your methods utilize the textResponseHandler?

peternic:
> Wait, actually why can't your methods utilize the textResponseHandler?

The reason is because we are defaulting to JSON as response handler, and not all failed requests might be setting the responsetype to 'text'. In an ideal world all api responses would set the responsetype, and we could rather default to text resonse if it was not a json response. @andersfylling 

But if you see away we could do it, I can change it.

andersfylling:
I see the request options are abstracted away, should probably expose those to the developer. For now I'm okay with using text as a failsafe for when json fails, though it feels a bit off to introduce a side effect.

> The default response handler is JSON, and this fails on the templates responses, so the request fails and does is not retried.

How come the tests for templates does not fail normally? Are the tests not verifying this?

peternic:
> I see the request options are abstracted away, should probably expose those to the developer. For now I'm okay with using text as a failsafe for when json fails, though it feels a bit off to introduce a side effect.
> 
> > The default response handler is JSON, and this fails on the templates responses, so the request fails and does is not retried.
> 
> How come the tests for templates does not fail normally? Are the tests not verifying this?

The JSON handler errors are not specific for templates. 

What do you mean by side effect?

I would say it is strange to default to a json handler without a fallback... If not fallback we should have text handler as default.

peternic:
> > I see the request options are abstracted away, should probably expose those to the developer. For now I'm okay with using text as a failsafe for when json fails, though it feels a bit off to introduce a side effect.
> > > The default response handler is JSON, and this fails on the templates responses, so the request fails and does is not retried.
> > 
> > 
> > How come the tests for templates does not fail normally? Are the tests not verifying this?
> 
> The JSON handler errors are not specific for templates.
> 
> What do you mean by side effect?
> 
> I would say it is strange to default to a json handler without a fallback...

If

andersfylling:
> I would say it is strange to default to a json handler without a fallback... If not fallback we should have text handler as default.

Yeah, that's a good point.

> What do you mean by side effect?

by side effect i mean that we are now handling plain text inside a json handler. It should be done outside the json handler, preferably by the caller.


Looking a bit deeper at the issue: `Unexpected end of JSON input`, it sounds like the server is sending a empty response? If it was plain text, I'd expect the error to be something like`Unexpected character at...`, the design doc does specify that endpoints should return a `{}` for empty responses. So if that is the case, then the service in question needs to be patched. I don't see any documentation for templates, so perhaps it's not too late to change? otherwise we should do some psuedo code like
```ts
const contentLength: number = +(res.headers.get("Content-Length") || 0);
return (res.json() ? contentLength > 0 : {}) as Promise<ResponseType>
```

However, we can add a failfast if the parser fails, and try the text handler in the rawRequest method. Perhaps we should just wrap line 208 in a try catch for SyntaxError (or is the exception something else?).

That being said, how can I reproduce this? Is it only when 429 takes place? Meaning ambassador is at fault? Then it might make more sense to simply not try and parse the body for 429, thought it would be nice with a consistent json error.

I don't think the sdk should accommodate services that sends out invalid json, as the fault lays elsewhere. But if the `Content-Type` in the response, is not `application/json`, then yes, I agree that the SDK here makes it difficult to specify response type in the endpoint implementation. But I don't believe we handle this by patching the json handler to return plain text in case it fails. Even though it seems like a simple way to fix this.

I'm being strict here, but if @vegardok or @jarlah or @ebrattli or @bdimitrijoski etc. over-rules me and approves, I won't push back and let it merge in peace :+1:

peternic:
> > I would say it is strange to default to a json handler without a fallback... If not fallback we should have text handler as default.
> 
> Yeah, that's a good point.
> 
> > What do you mean by side effect?
> 
> by side effect i mean that we are now handling plain text inside a json handler. It should be done outside the json handler, preferably by the caller.
> 
> Looking a bit deeper at the issue: `Unexpected end of JSON input`, it sounds like the server is sending a empty response? If it was plain text, I'd expect the error to be something like`Unexpected character at...`, the design doc does specify that endpoints should return a `{}` for empty responses. So if that is the case, then the service in question needs to be patched. I don't see any documentation for templates, so perhaps it's not too late to change? otherwise we should do some psuedo code like
> 
> ```ts
> const contentLength: number = +(res.headers.get("Content-Length") || 0);
> return (res.json() ? contentLength > 0 : {}) as Promise<ResponseType>
> ```
> 
> However, we can add a failfast if the parser fails, and try the text handler in the rawRequest method. Perhaps we should just wrap line 208 in a try catch for SyntaxError (or is the exception something else?).
> 
> That being said, how can I reproduce this? Is it only when 429 takes place? Meaning ambassador is at fault? Then it might make more sense to simply not try and parse the body for 429, thought it would be nice with a consistent json error.
> 
> I don't think the sdk should accommodate services that sends out invalid json, as the fault lays elsewhere. But if the `Content-Type` in the response, is not `application/json`, then yes, I agree that the SDK here makes it difficult to specify response type in the endpoint implementation. But I don't believe we handle this by patching the json handler to return plain text in case it fails. Even though it seems like a simple way to fix this.
> 
> I'm being strict here, but if @vegardok or @jarlah or @ebrattli or @bdimitrijoski etc. over-rules me and approves, I won't push back and let it merge in peace 👍

Good points! 

I initiated to look into the Ambassador, so then maybe making the 429 from Ambassador either set the type or return an empty {} would fix this.

peternic:
> That being said, how can I reproduce this? Is it only when 429 takes place? Meaning ambassador is at fault? Then it might make more sense to simply not try and parse the body for 429, thought it would be nice with a consistent json error.

Not sure if it is only on 429s, I am afraid that we might get other responses into this since we are defaulting to json though.

To test it you can just do a simple for loop and ask for either:
```
for( var i=0; i<50; i++) {
  // client.templates.groups.list();
  client.relationships.list({ filter: { targetTypes: ['timeseries']}, limit: 10 });
}
```

andersfylling:
> Not sure if it is only on 429s, I am afraid that we might get other responses into this since we are defaulting to json though.

I think this case can trigger when our gateway blocks requests in general as well. I do think a try catch on line 208 would be nice regardless. It would add a fallback to text for all handlers if there is a failure, instead of just the json handler. If it's do-able?

peternic:
> > Not sure if it is only on 429s, I am afraid that we might get other responses into this since we are defaulting to json though.
> 
> I think this case can trigger when our gateway blocks requests in general as well. I do think a try catch on line 208 would be nice regardless. It would add a fallback to text for all handlers if there is a failure, instead of just the json handler. If it's do-able?

I can move the logic out yes, should be doable. But will be the same clone logic though

peternic:
@andersfylling Have moved the logic to fallback to text for any of the handlers if they fail.

packages/core/src/httpClient/basicHttpClient.ts
```diff
@@ -197,11 +205,17 @@ export class BasicHttpClient {
     const responseHandler = BasicHttpClient.getResponseHandler<ResponseType>(
       request.responseType
     );
-    const data = await responseHandler(res);
+    // Cloning to fallback on text response if response is failing the responsehandler.
+    // node-fetch < 3.0 will hang on clone() for large responses, that is why the parallel promises https://github.com/node-fetch/node-fetch#custom-highwatermark
+    const resClone = res.clone();
+    const [data, textFallback] = await Promise.all([
```
andersfylling, line 203: don't the other handlers need to be updated for this as well?

andersfylling: to extend on this: what happens if we get a 429 when requesting a file?

eg. using `arrayBufferResponseHandler`.

peternic: Hmm, the responseHandler here is either arrayBufferResponseHandler, jsonResponseHandler, or textResponseHandler.

So if I am not mistaken, if either one of these responsehandlers throws an error, it will be catched and then it will respond with the text version.

That was the plan at least 😅

andersfylling: is this change still needed? 
image

peternic: No, forgot to revert that, now it should be reverted. Thanks!

peternic: Any final thoughts @andersfylling ?

andersfylling: I'm going through the error handling now. Tests (fetch mock) would be great, to actually verify that this fixes the issues.

peternic: Have added a simple test, that fails without the fix, and works with the fix 👍🏻

andersfylling (approved):

PR #707, opened 2021-10-21
feat(documents pipeline update): remove deprecated str support

andersfylling:
BREAKING CHANGE: removes support for str values in certain fields when updating pipeline

### DIFF
--- a/packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
+++ b/packages/playground/src/__tests__/api/documentsApi.uint.spec.ts
@@ -84,11 +84,11 @@ describe('Documents unit test', () => {
               TERMS: ['secret', 'confidential', 'sensitive'],
             },
             fieldMappings: {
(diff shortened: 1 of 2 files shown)


arnels (approved):

tommythorsen (approved):
Looks fine to me. While it technically is a breaking change, nobody are actually calling these at the moment, so it will not break anything in practice.

Is this consistent with other endpoints? Do we typically allow single strings as well as arrays?

henrist:
Shouldn't the "BREAKING CHANGE" description also be part of a commit message? Not sure how the auto-versioning works here

andersfylling:
The github auto merge just ignored the tests. 

Yeah I add the breaking change footer when merging.

andersfylling:
@tommythorsen Yes. we re-use the types so this should affect all related endpoints for pipeline config.

PR #708, opened 2021-10-21
chore(doc, samples): improve samples and doc related to authentication 

vegardok:
1. adds another react app that integrates with both CDF and the Microsoft graph
2. adds a node example for client credentials using `@azure/msal-node`
3. reorganize samples into `nodejs` and `react` folders
4. update guides/authentication.md to cover all normal cases (API keys, legacy flow and auth code w. pkce and client credentials using Azure)


samples/nodejs/oidc-typescript/.gitignore
```diff
@@ -0,0 +1 @@
+quickstart.js
```
andersfylling, line 1: why is this ignored?

andersfylling (approved):
LGTM! it's a large change so i might have missed something, but generally looks nice.


Also nice that u updated to node 16!

samples/nodejs/oidc-typescript/.gitignore
vegardok: `tsc quickstart.ts` will result in `quickstart.js` in the root dir

PR #716, opened 2021-11-05
feat(documents): use generics for documents types

andersfylling:
BREAKING CHANGE: simplified types

Prefer use of generics over duplicating types with different values.

Openapi schemas that are just an alias are inlined. eg. 
 - `LabelList` => `Label[]`


henrist:
Introducing generics will likely make it harder to get compatibility to auto-generated types based on the schema later? I feel that we should strive to reuse the same types as exposed in the API schema if possible. What do you think?

A little side note: It's nice that the commits in general allow me to look at the individual changes to ease reviewing (since looking at the whole PR is very hard in this case). I think the commit types are a bit off (I consider none of these to be `chore`). It would be nice with a little rebase to squash the commits that belong together (e.g. the `chore: move classifiers to own file` commits) - I often use `git rebase -i origin/master` to go through a branch to clean up a bit.

packages/playground/src/api/documents/pipelinesTypes.ts
```diff
@@ -28,7 +28,7 @@ export interface DocumentsSourceFile {
   name?: string[];
   directory?: string[];
   content?: string[];
-  metadata?: StringToStringArrayMap;
+  metadata?: Map<string[]>;
```
henrist, line 25: Maybe avoid the use of `Map` as name since it shadows a [global](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)?

henrist: Maybe just use `Record<string, string[]>` ?

andersfylling: I'll look into if we can use Record, due to backwards capability. And thanks, I forgot about `Map`!

henrist: What do you mean by "backwards compatibility" in this case?

tommythorsen (approved):
Looks good! :+1:

I second using a different name than `Map`, though.

andersfylling:
Closing as I now have a script for autogenerated types

PR #720, opened 2021-12-09
chore(geospatial): updates spatial to geospatial & type name changes

saagar-cognite:
- Updates v1 to accommodate spatial -> geospatial in end points
- fixes type names & externalId names.


packages/stable/src/api/geospatial/featureAPI.ts
```diff
@@ -5,29 +5,29 @@ import {
...
+    return this.createEndpoint<FeatureCreateItem>(
       features,
       this.url(`${featureTypeExternalId}/features`)
     );
   };
 
   public retrieve = (
     featureTypeExternalId: CogniteExternalId,
-    ids: IdEither[],
+    externalIds: IdEither[],
```
pawelzwolenskicognite: type should be ExternalId[] instead of IdEither[]

packages/stable/src/api/geospatial/types.ts
```diff
@@ -65,20 +65,20 @@ export interface FeatureTypes {
   searchSpec?: {
     [searchName: string]: {
```
pawelzwolenskicognite: maybe `indexName` would be better?

packages/stable/src/api/geospatial/types.ts
```diff
@@ -49,12 +49,12 @@ export interface FeaturesCreateItem {
...
-export interface Features extends FeaturesCreateItem {
+export interface Feature extends FeatureCreateItem {
   createdTime: Date;
   lastUpdatedTime: Date;
 }
 
-export interface FeatureTypes {
+export interface FeatureType {
   id: CogniteInternalId;
   externalId: CogniteExternalId;
```
pawelzwolenskicognite: What is CogniteExternalId and how it is different from ExternalId?

packages/stable/src/api/geospatial/types.ts
```diff
@@ -65,20 +65,20 @@ export interface FeatureTypes {
...
         | ['_created_at']
         | ['_updated_at']
         | ['_external_id'];
     };
   };
 }
 
-export interface FeatureTypesCreateItem {
+export interface FeatureTypeCreateItem {
   externalId: CogniteExternalId;
```
pawelzwolenskicognite: ExternalId?

packages/stable/src/api/geospatial/featureTypeAPI.ts
```diff
@@ -0,0 +1,26 @@
+import {
+  BaseResourceAPI,
+  CursorAndAsyncIterator,
+  ExternalId,
+} from '@cognite/sdk-core';
+import { FeatureType, FeatureTypeCreateItem } from './types';
+
+export class FeatureTypeAPI extends BaseResourceAPI<FeatureType> {
```
pawelzwolenskicognite, line 13: are you going to add update method too?

packages/stable/src/api/geospatial/types.ts
saagar-cognite: CogniteExternalId just tells that id is a string whereas
ExternalId is { externalId : id } shaped object. Generally used in search.

saagar-cognite: https://github.com/cognitedata/cognite-sdk-js/blob/708763f5568ad4dc4ddc48e3e2bedc2f2bc732c6/packages/core/src/types.ts#L37

packages/stable/src/cogniteClient.ts
```diff
@@ -215,7 +215,7 @@ export default class CogniteClient extends BaseCogniteClient {
       this.httpClient,
       this.metadataMap
     );
-    this.spatialApi = this.apiFactory(SpatialAPI, 'spatial');
+    this.geospatialApi = this.apiFactory(GeospatialAPI, 'geospatial');
```
rek, line 218: this is breaking and will require a major bump right?

saagar-cognite: @rek 
Yes. I've added BREAKING CHANGE: description in the commit message footer.
I'm hoping just adding that to one of the commit works 🤔

packages/stable/src/api/geospatial/featureTypeAPI.ts
saagar-cognite: added

pawelzwolenskicognite (approved):

packages/stable/src/api/geospatial/types.ts
saagar-cognite: extended ExternalId interface. 
I'm hoping that is what you were looking for here.

packages/stable/src/cogniteClient.ts
rek: i think the important part is the message when you 'squash and merge'

andersfylling: ~you need to rename the title to `feat` instead of `chore`.~

andersfylling: ~or perhaps just `fix`, it doesn't seem like a typical chore to do a breaking change to me~

andersfylling: I was wrong =D

andersfylling (approved):

saagar-cognite:
Converted to draft as we're still expecting more changes to come in after the attributes to properties change.
Will convert it back to review once all breaking changes are in.
@rek @pawelzwolenskicognite @andersfylling

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,267 @@
...
+
+interface AndFilter {
+  and?: NonRecursiveFilters[];
+}
+
+interface NotFilter {
+  not?: NonRecursiveFilters;
+}
+
+interface SortFilter {
```
pawelzwolenskicognite: sort is not a filter

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,267 @@
...
+  sort?: string[];
+}
+
+interface AllowCrsTransformationFilter {
```
pawelzwolenskicognite: it is not a filter, it is just a boolean flag

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,267 @@
...
+      size: number;
+      description?: string;
+      optional?: boolean;
+    };
+
+export interface Properties {
+  [property: string]: Property;
+}
+
+export interface Geospatial {
```
pawelzwolenskicognite, line 71: what is this interface for?

packages/stable/src/api/geospatial/featureAPI.ts
```diff
@@ -0,0 +1,91 @@
...
+    return this.createEndpoint<FeatureCreateItem>(
+      features,
+      this.url(`${featureTypeExternalId}/features`)
+    );
+  };
+
+  public retrieve = (
+    featureTypeExternalId: CogniteExternalId,
+    externalIds: ExternalId[],
+    params?: FeatureOutputParams
```
pawelzwolenskicognite: maybe `outputParams` is better name?

packages/stable/src/api/geospatial/types.ts
saagar-cognite: This is place holder for the Geospatial class as it has feature & featureType inside it.

packages/stable/src/api/geospatial/featureAPI.ts
saagar-cognite: sure.

packages/stable/src/api/geospatial/types.ts
saagar-cognite: naming it just Sort seems too generic.
Is SortParams a better name ?

packages/stable/src/api/geospatial/featureAPI.ts
```diff
@@ -0,0 +1,91 @@
...
+        path: this.url(`${featureTypeExternalId}/features/byids`),
+        items: request,
+        params,
+      })
+    );
+  };
+
+  public update = (
+    featureTypeExternalId: CogniteExternalId,
+    changes: FeatureCreateItem[]
```
pawelzwolenskicognite, line 42: Request body for update endpoint is different than for create endpoint. Does it actually work in that form?

packages/stable/src/api/geospatial/featureAPI.ts
```diff
@@ -0,0 +1,91 @@
...
+    return this.deleteEndpoint(
+      externalIds,
+      params,
+      this.url(`${featureTypeExternalId}/features/delete`)
+    );
+  };
+
+  public search = (
+    featureTypeExternalId: CogniteExternalId,
+    params: FeatureSearchFilter = {}
```
pawelzwolenskicognite: I would suggest to rename `params` to `filter` here and also in other methods.

packages/stable/src/api/geospatial/featureAPI.ts
```diff
@@ -0,0 +1,91 @@
...
+    return this.updateEndpoint(
+      changes,
+      this.url(`${featureTypeExternalId}/features/update`)
+    );
+  };
+
+  public delete = (
+    featureTypeExternalId: CogniteExternalId,
+    externalIds: ExternalId[],
+    params: FeatureOutputParams = {}
```
pawelzwolenskicognite: I would suggest to rename `params` to `output` or `outputParams` here and also in other methods.

pawelzwolenskicognite:
@saagar-cognite we will have one more breaking change with naming (not very big from the JS SDK perspective: adding `st` prefix for filter and aggregate names) that is required before Geospatial will be publicly available. We are blocked by the code freeze now, but it will come in early January. I will keep you in touch

packages/stable/src/api/geospatial/types.ts
pawelzwolenskicognite: We actually call it `orderBy`. Maybe `orderSpec` ?

packages/stable/src/api/geospatial/types.ts
pawelzwolenskicognite: should then feature & featureType be specified in the interface? or they are not accessible from that interface?

packages/stable/src/api/geospatial/types.ts
andersfylling: I would recommend using the same name as in the service-contracts repo - if there is a definition already. The consistency might promote ease of use across Cognite, as the type is already named and defined in the documentation.

Ofcourse, it's "strongly" typed, so it's not a big deal. But a nice thought when you're uncertain about naming.

example from the assets api; we see a `ExternalEvent` type referenced.
image

saagar-cognite: Oh that seems like a correct approach @andersfylling 
So according to the following screenshot
Screenshot 2021-12-23 at 1 14 24 PM

Should I be renaming 
`AllowCrsTransformationFilter` -> `GeospatialAllowCrsTransformation`
`FeatureSearchFilter` -> `GeospatialFeatureFilter`
`FeatureOutputParams` -> `GeospatialOutput`
and so on ?

saagar-cognite: But sort doesn't show any named type, so should I just keep it like string[] ? @andersfylling 
or better to name a `GeospatialSort` or `Sort` to be used ?

packages/stable/src/api/geospatial/featureAPI.ts
saagar-cognite: but filter is a parameter inside params 🤔

packages/stable/src/api/geospatial/types.ts
andersfylling: > Should I be renaming
AllowCrsTransformationFilter -> GeospatialAllowCrsTransformation
FeatureSearchFilter -> GeospatialFeatureFilter
FeatureOutputParams -> GeospatialOutput
and so on ?

Yes. It's up to you, in the end. We do it for everything under the documents service (moving to generating types directly from openapi soon for /documents, so this makes it easier to avoid breaking changes for the grep team).

andersfylling: > But sort doesn't show any named type, so should I just keep it like string[] ? @andersfylling
or better to name a GeospatialSort or Sort to be used ?

Does sort need to be a standalone object?

pawelzwolenskicognite:
@saagar-cognite we have introduced the change with `st` prefix for geospatial operators. The full spec is under the usual link (https://pr-1323.specs.preview.cogniteapp.com/v1.json.html). Please contact me if you have any questions

saagar-cognite:
> @saagar-cognite we have introduced the change with `st` prefix for geospatial operators. The full spec is under the usual link (https://pr-1323.specs.preview.cogniteapp.com/v1.json.html). Please contact me if you have any questions

Sure, @pawelzwolenskicognite .

packages/stable/src/api/geospatial/types.ts
saagar-cognite: Probably sort doesn't need to be standalone. Thanks @andersfylling .

saagar-cognite:
@pawelzwolenskicognite @andersfylling This is ready for review now.

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,245 @@
...
+  property: string;
+  pattern: string;
+}
+
+interface GeospatialAllowCrsTransformation {
+  allowCrsTransformation?: boolean;
+}
+
+interface RecursiveFilters {
+  or?: NonRecursiveFilters[];
```
pawelzwolenskicognite: RecursiveFilters can have other Recursive filters as children, e.g.
```
{
  "and": [
      { "equals": ... },
      { "or": [..., ..., ] }
  ]
}
```
What is the reason that you distinguish between recursive and non-recursive filters?

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,245 @@
...
+      optional?: boolean;
+    }
+  | {
+      type: GeometryPropertyType;
+      srid?: number;
+      description?: string;
+      optional?: boolean;
+    }
+  | {
+      type: 'STRING';
```
pawelzwolenskicognite, line 63: why STRING is separated?

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,245 @@
...
+        | { property: string; lte: string | number };
+    }
+  | {
+      stWithin: GeospatialPropertyAndValue;
+    }
+  | {
+      stIntersects: GeospatialPropertyAndValue;
+    }
+  | {
+      completelyWithin: GeospatialPropertyAndValue;
```
pawelzwolenskicognite: stContainsProperly

packages/stable/src/api/geospatial/types.ts
```diff
@@ -0,0 +1,245 @@
...
+  properties: Properties & {
+    _created_at: { type: 'LONG' };
+    _updated_at: { type: 'LONG' };
+    _external_id: { type: 'STRING'; size: 32 };
+  };
+  searchSpec?: {
+    [indexName: string]: {
+      properties:
+        | string[]
+        | ['_created_at']
```
pawelzwolenskicognite: API returns `externalId`, `lastUpdatedTime` and `createdTime`

packages/stable/src/api/geospatial/types.ts
saagar-cognite: The distinguish is just for better breakdown.
Both are Recursive and Non-Recursive are local variables only and exported is complete filter only.

packages/stable/src/api/geospatial/types.ts
saagar-cognite: As we've `size` only for STRING type and srid only with Geometrical types, hence, these clear breakdowns help while sending filters with correct fields being sent with correct type.

pawelzwolenskicognite: clear now 👍

packages/stable/src/api/geospatial/types.ts
pawelzwolenskicognite: but this breakdown is limiting the power of API (nested recursive filters)

pawelzwolenskicognite: also, is plural for class/interface names a convention in JS?

packages/stable/src/api/geospatial/types.ts
saagar-cognite: @pawelzwolenskicognite This is featureType and in the example this is what is returned after creation.
Can you please confirm if this is correct or not ?
Screenshot 2022-01-04 at 12 44 35 PM

pawelzwolenskicognite: looks like the example is outdated - we will fix it, good catch!

btw. that should be tested in some automatic tests?

pawelzwolenskicognite: @saagar-cognite the example was fixed (refresh https://github.com/cognitedata/service-contracts/pull/1323)

packages/stable/src/api/geospatial/types.ts
saagar-cognite: @pawelzwolenskicognite I've removed the recursive & non recursive.
Can you check if it seems correct now ?

pawelzwolenskicognite: yes, much better, thanks!

pawelzwolenskicognite:
@saagar-cognite I have no more remarks and everything looks fine for me, great job!
I think the PR should also be reviewed by somebody with better JS experience than me :)

PR #722, opened 2021-12-09
fix(spatial): changes api endpoint from spatial to geospatial

saagar-cognite:


andersfylling:
is this still relevant? @saagar-cognite

PR #724, opened 2021-12-21
fix: use case-insensitive lookup for http header

andersfylling:
Given that the rfc states http header keys are to be case-insensitive, we should not be using direct map lookup.

I also noticed the x-request-id seem to be lowercase in responses, and updated it as such. Though the new lookup function should fix any case-sensitive issues.

### DIFF
--- a/packages/core/src/httpClient/httpHeaders.ts
+++ b/packages/core/src/httpClient/httpHeaders.ts
@@ -1,0 +1,3 @@
+export interface HttpHeaders {
+  [key: string]: string;
+}
--- a/packages/core/src/httpClient/httpHeaders.ts
+++ b/packages/core/src/httpClient/httpHeaders.ts
@@ -1,3 +1,19 @@
 export interface HttpHeaders {
   [key: string]: string;
 }
+
+/** @hidden */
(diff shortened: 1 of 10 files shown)


packages/core/src/httpClient/httpHeaders.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+/** @hidden */
+export function getHeaderField(
+  headers: HttpHeaders,
+  key: string
+): string | undefined {
+  const lowercaseKey = key.toLowerCase();
+
+  for (const headerKey in headers) {
```
ebrattli, line 12: Is there a function find such that you can write ```headers.find(headerKey => headerKey.toLowerCase() == lowercaseKey)``` or something like that?

ebrattli (approved):
Anyways, LGTM!

packages/core/src/httpClient/httpHeaders.ts
andersfylling: Not that I know of. You can do some casting and then map it, but it becomes fairly verbose imo.

PR #729, opened 2022-01-13
feat(annotations): add sdk for annotations api (CXT-488)

cyriaquebrousse:
Add basic support for Annotations API (see [specification](https://docs.cognite.com/api/playground/#tag/Annotations)).

Support for annotation types is not included in this PR for the sake of speed, which means that `annotationType` (str) is a hardcoded enum in this PR, and `data` is a generic object, which means that its contents are not typechecked by the SDK.
Support for annotation types will be added in a subsequent PR.

### DIFF
--- a/packages/playground/src/index.ts
+++ b/packages/playground/src/index.ts
@@ -2,4 +2,3 @@
 
 export { default as CogniteClientPlayground } from './cogniteClientPlayground';
 export * from './types';
-export * from './api/annotations/types';
--- a/packages/playground/src/index.ts
+++ b/packages/playground/src/index.ts
@@ -2,3 +2,4 @@
 
 export { default as CogniteClientPlayground } from './cogniteClientPlayground';
 export * from './types';
+export * from './api/annotations/types';
--- a/packages/playground/src/types.ts
+++ b/packages/playground/src/types.ts
@@ -18,6 +18,8 @@ import {
 // This file is here mostly to allow apis to import { ... } from '../../types';
 // Overriding types should probably be done in their respective API endpoint files, where possible
 
+export * from './api/annotations/types';
+
 export interface IntIn {
   in: number[];
 }
--- a/packages/playground/src/types.ts
+++ b/packages/playground/src/types.ts
@@ -13,6 +13,12 @@ import {
   FileMimeType,
   Metadata,
   ItemsWrapper,
+  SetField,
+  SinglePatchString,
+  SinglePatchRequiredString,
+  SinglePatch,
+  IdEither,
+  InternalId,
 } from '@cognite/sdk';
 
 // This file is here mostly to allow apis to import { ... } from '../../types';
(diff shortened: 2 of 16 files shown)


packages/playground/src/types.ts
```diff
@@ -537,3 +543,68 @@ export interface FunctionSchedulesFilter {
   createdTime?: Date;
   cronExpression?: string;
 }
+
+/* ************************
+   Annotations API types
```
andersfylling: can you put this in a types.ts file inside the annotations api folder?
Then just do the following here:
```ts
export * from './api/annotations/types';
```

packages/playground/src/types.ts
```diff
@@ -537,3 +543,68 @@ export interface FunctionSchedulesFilter {
...
+export interface AnnotationFilter {
+  filter: AnnotationFilterProps;
+}
+
+export interface AnnotationFilterProps {
+  annotatedResourceType: AnnotatedResourceType;
+  annotatedResourceIds: IdEither[];
+  annotationType?: AnnotationType;
+  creatingApp?: string;
+  creatingUser?: string | null;
```
andersfylling: what does null do here?

peetcremer1: `null` is a special value that signals that the annotation was not created by a particular user, i.e. the annotation was created by a service.

peetcremer1:
I did not see any particular code to handle the special value of `creatingUser` upon retrieving Annotations. On the python-sdk side of things, I had to add such code, see forexample here: https://github.com/cognitedata/cognite-sdk-python-experimental/blob/CXT-460/cognite/experimental/data_classes/annotations_v2.py#L107

Most likely this is not necessary here as JS does a better job of distinguishing `undefined` and `null`, wheras python only has `None`. However, we should double check that within the JS SDK a retrieved annotation which was created with `creatingUser: null` has indeed `creatingUser: null` when fetched again.

packages/playground/src/types.ts
```diff
@@ -537,3 +543,68 @@ export interface FunctionSchedulesFilter {
...
+  linkedResourceType?: LinkedResourceType;
+  linkedResourceId?: CogniteInternalId;
+  linkedResourceExternalId?: CogniteExternalId;
+  status: AnnotationStatus;
+}
+
+export interface AnnotationChangeById extends InternalId, AnnotationUpdate {}
+
+export interface AnnotationUpdate {
+  update: {
```
peetcremer1: `status` seems to be missing here

packages/playground/src/types.ts
```diff
@@ -537,3 +543,68 @@ export interface FunctionSchedulesFilter {
   createdTime?: Date;
   cronExpression?: string;
 }
+
+/* ************************
+   Annotations API types
+************************ */
+
+export type AnnotatedResourceType = 'file';
+export type AnnotationType =
```
peetcremer1: Perhaps it is better to not have this as a type for now and just let the user input a String, as we have currently no good way to fill in the possible values automatically.
Then, when we have a proper integration with the `annotation-types` repo, we can create this type/enum automatically and it has all the available values.

packages/playground/src/types.ts
cyriaquebrousse: Status cannot be updated, according to the service contracts, see https://docs.cognite.com/api/playground/#operation/annotationsUpdate. Does this need to be updated?

peetcremer1: I think it needs to be updated. How else would you approve / reject an annotation before the feedback loop is implemented? It is also allowed by the API currently, see here:
https://github.com/cognitedata/context-api/blob/master/src/api/annotations_v2/schemas.py#L161

cyriaquebrousse: https://github.com/cognitedata/service-contracts/pull/1583

packages/playground/src/types.ts
cyriaquebrousse: Agreed; changed.

packages/playground/src/types.ts
cyriaquebrousse: Done

packages/playground/src/index.ts
```diff
@@ -2,3 +2,4 @@
 
 export { default as CogniteClientPlayground } from './cogniteClientPlayground';
 export * from './types';
+export * from './api/annotations/types';
```
andersfylling: move this to the types.ts file, then we have all the playground type/references in one location.

andersfylling (approved):
LGTM!

~see above comment :+1:~

peetcremer1 (approved):
LGTM

chingyulin (approved):
LGTM!

PR #736, opened 2022-01-17
chore: changing code owners for the javascript sdk

cognite-ornellas:


jarlah (approved):
😀

andersfylling:
~You probably want to update the other fields as well.~

~grep would like notifications for when there are changes to things in /documents, but we don't need it for core and all of playground.~

cognite-ornellas:
> ~You probably want to update the other fields as well.~
> 
> ~grep would like notifications for when there are changes to things in /documents, but we don't need it for core and all of playground.~

I saw you already changed it. Thank you!

PR #737, opened 2022-01-18
test: increase global timeout for tests

andersfylling:


henrist:
See https://cognitedata.slack.com/archives/G011ZCXB2BE/p1642501541011700?thread_ts=1642500317.011100&cid=G011ZCXB2BE

In general I wouldn't expect any of our API endpoints except document preview to take more than 5 seconds? Do you have any examples of failures?

If we have issues with timeouts in general we might also consider changing https://jestjs.io/docs/configuration#testtimeout-number

andersfylling:
There are several cases where our tests fail due to timeouts. People have also sent me messages about the tests failing due to time outs. 

This has also happened for other services, and the direct cause isn't known as of now.

henrist:
> This has also happened for other services, and the direct cause isn't known as of now.

Then it sounds like it makes sense to increase the general timeout instead of doing it individually? I think the int tests should ideally run separately from the others so they can also have their own config, but that can be looked more into later.

henrist:
Here's an example of timeout for another service: https://github.com/cognitedata/cognite-sdk-js/runs/4894382847?check_suite_focus=true

henrist (approved):
I think this makes sense. :+1:

It would be nice if the commit/merge message explains the background/reason for this, so that is easily available when checking the line commit message.

andersfylling:
Good point! I'll update the commit description

pedroharbs (approved):

PR #740, opened 2022-01-21
chore: codegen for documents service

andersfylling:


package.json
```diff
@@ -36,9 +36,12 @@
     "docs:bundle": "rm -rf docs/ && yarn allsdk docs:bundle --stream && touch docs/.nojekyll",
     "validateDocLinks": "find . -name \\*.md -not -path '*/node_modules/*' -print0 | xargs -0 -n1 yarn markdown-link-check -c ./markdownlint.json",
     "test-snippets": "yarn allsdk test-snippets --stream",
-    "test-samples": "SKIP_PREFLIGHT_CHECK=true yarn allsamples test --stream"
+    "test-samples": "SKIP_PREFLIGHT_CHECK=true yarn allsamples test --stream",
+
+    "generate-types:playground:documents": "sh scripts/codegen/build.sh"
```
henrist:
```suggestion
    "generate-types:playground:documents": "scripts/codegen/build.sh"
```

(And make sure the script has +x)

scripts/codegen/build.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+set -e
+
+FOLDER="./tmp/service-contracts"
+# rm -rf ./tmp
+mkdir -p ./tmp
+
+git -C ${FOLDER} pull || git clone git@github.com:cognitedata/service-contracts.git ${FOLDER}
```
henrist: Since this is a public repo I'd advise to go against the public endpoints from https://docs.cognite.com/

scripts/codegen/index.js
```diff
@@ -0,0 +1,263 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
henrist, line 1: Shouldn't there be some tests for all this?

If we were to update dependencies we wouldn't know its failing before someone tries to generate updated types.

scripts/codegen/README.md
```diff
@@ -0,0 +1,2 @@
+generate typescript types for the grep team. Note that this does not create endpoints, and may re-defined already defined common types found in the sdk or core.
```
henrist, line 1: Can we write a proper Markdown formatted README?

scripts/codegen/build.sh
```diff
@@ -0,0 +1,19 @@
...
+cd ${FOLDER}
+yarn install
+chmod +x build.sh
+./build.sh
+cd ../..
+
+yarn install
+node scripts/codegen/index.js
```
henrist: Can we split up the tasks of generating the types vs. linting and building? So one can work on this with type generation while not doing the rest

henrist (commented):
Some general notes:

Since this is a multi-package repo, maybe it would make sense to have this as a package?

I'd like to see this being run in CI, and it would be nice to keep track of which spec it was previously generated for. When running in CI, we can't have it always pick the latest spec, since that would create failures for every time. It should be stable and reproducable without errors. Since the public docs are not versioned (like e.g. a Git commit), I suggest we _copy_ the spec file into this repo (with a script that can be run to update), and ensure CI generates types from that file, and fail on changes. That would then also work as a kind of "test" for the codegen script, and we ensure the generates types are always in sync. In the future, a GH Action can pull the latest spec on a schedule and create PRs for updates.

andersfylling:
> we can't have it always pick the latest spec, since that would create failures for every time.

What do you mean? How will this always create failures?

andersfylling:
In the past, whenever there was a change to our service api or openapi spec, we would go into here and manually update this. I believe simplifying the workload to run a script helps significantly.

I don't think I agree that we have to pull in the openapi spec into this repository for version management. This sdk should be updated in accordance with the master branch, as that is the documentation displayed in the cognite documentation page.

I do agree that it would be nice to setup PR's automatically as this is now possible, and having PR's created and linked to PR's in the service contracts repo would be sweet, but do we need all of these nice features right now? Can't that be an extension to this?

*Maybe* we can trigger a CI to run via repository dispatch: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch

I haven't looked into any baker logic.

henrist:
> What do you mean? How will this always create failures?

"... every time _the spec changes_" that is

We should have changes go through a commit/PR and not be automatically included, since that would cause us to not have deterministic builds. Having input change on a new build for a unrelated changes is bad, and might leave a previously working tree in a failing state on rebuild.

That is why you want to lock the input to the build as much as possible. Since there is no immutable public resource of the API spec, a solution is to copy the spec into the repo. To generate updated types, it then starts by updating the copied spec.

To say it another way: If the spec if changed, it shouldn't be a new _release_ just by triggering a new build (that would make the build non-deterministic, and also hard to keep source control over the built library).

andersfylling:
We can't keep a copy of the openapi spec for every repo we want to have codegen.

I feel like this is a lot of work for an API we plan to eventually stop supporting?

henrist:
> We can't keep a copy of the openapi spec for every repo we want to have codegen.

Why?

I'd say that's exactly what you want to have. It helps you know what you are coding for, and you have source control over the version and changes of it.

> I feel like this is a lot of work for an API we plan to eventually stop supporting?

I assume the long-term goal would be to generate stuff for more APIs, not only this one in playground?

tommythorsen:
If I understood things correctly, there's another team that will take more ownership over this sdk, no? Do they want code generation?

andersfylling:
> I'd say that's exactly what you want to have. It helps you know what you are coding for, and you have source control over the version and changes of it.

It would be a dependency struggle. Imagine having to keep N repos in sync, when they have exactly the same information. It's easier to just write typescript then. Cause then you're just having a way to represent the api, there shouldn't be an open api to represent an openapi spec just for the sake of codegen.

It would be better to introduce tooling that verifies the openapi contracts aren't changed in a breaking manner.

Again tho, this all goes back to my original suggestion. One repository where we use a declerative language, and then generate openapi, js sdk, python, etc. from that. It would be a lot cleaner and more powerful than the current approach.

andersfylling:
> If I understood things correctly, there's another team that will take more ownership over this sdk, no? Do they want code generation?

I have brought it up to them, but did not get a response on slack.

@cognitedata/developer-tooling will you be aiming for code generation in the future?

tommythorsen:
> > I'd say that's exactly what you want to have. It helps you know what you are coding for, and you have source control over the version and changes of it.
> 
> It would be a dependency struggle. Imagine having to keep N repos in sync, when they have exactly the same information. It's easier to just write typescript then. Cause then you're just having a way to represent the api, there shouldn't be an open api to represent an openapi spec just for the sake of codegen.

Let me elaborate a bit on this. I do agree that there should only be a single source of truth for the specification. However, like with dependency management, setting up something where you always fetch the "latest" version of a dependency has a lot of potential for going wrong. This is why lots of systems use lock files to lock dependencies to specific versions and ensure people don't get lots of unexpected errors when they try to build unrelated things.

It would be ok if we could point to specific SHAs in the service-contracts repo for the contract. Sort of like described here: https://github.com/cognitedata/service-contracts#keep-the-spec-in-your-existing-repo. However, I saw the comment about this being a public repo and that linking to files in a private repo might not be so nice. I agree with that. If we want to keep this repo public, that leaves us with the option of fetching contracts from docs.cognite.com.

docs.cognite.com does not, as far as I can see, have versioned contracts. The only option is to get the latest version. That means that the only way we have to "lock" the version of the contract, is to store it in this repo. The contract stored in this repo should not be something people are meant to edit. The only way of changing it should be to fetch a new version from docs.cognite.com. We can't really enforce this, but we can encourage it in various ways. We can inject a big ugly header in the file, stating that people should not edit this file, but instead run some named script that fetches new versions from docs.cognite.com. It might also be possible for us to preprocess the contract slightly and maybe parse it and re-serialize it as json without any unnecessary whitespace, making it not look like something that should be edited by hand.

scripts/codegen/README.md
janisharbs: I agree, because we are trying to run here, and getting some errors.

scripts/codegen/index.js
janisharbs: Agree.

scripts/codegen/README.md
janisharbs: Maybe write some steps like how to run and etc.

cognite-ornellas: Maybe we could also describe a bit more about how to run the codegen and, if we need need something else in order to run it, describe it too.

andersfylling: some of the output errors are related to cognite tooling, the oas merger, which are expected for playground.

henrist:
Yep - thanks for elaborating and explaining this is in a simpler manner @tommythorsen. Does that description make more sense @andersfylling ?

In regard to keeping thing up-to-date I'd recommend creating a clear process and then automate it. This will also correctly produce new versions of the SDK when there are changes to the spec, as otherwise there would be no clear way to detect that, other than doing a new build and comparing with a previous build.

andersfylling:
@henrist Yes, I'm more on board now. 

That being said, can someone explain to me the practical difference between manually removing/adding/editing a lock file vs manually running a command when new types need to be generated? I get it with libraries, but this case seems different to me as we want the sdk's to always reflect the documentation.

To me it seems both blocks the same way(?).  Seems to me that a manual approach is just as safe?

If we completely automate this, won't we still have to go in and make a PR to update the lock file? during that time it can race, causing the version to be outdated.

I guess we could create reference a sha sum created by the xtag of a desired service from here (using this to recursively link together all the relevant schemas): https://storage.googleapis.com/cognitedata-api-docs/dist/playground.json

Still though, if we don't introduce some policy around renaming types with our service contracts, we'll still introduce breaking changes automatically. Regardless of a lock file or manually running a yarn command for generating the types.

Or am I misunderstanding something?

That being said, I do prefer that the type names in the documentation directly reflects the type names I will be working with, but it's not a must with a statically typed language.


Currently what I do to update types now:
 1. create a branch
 2. run `yarn generate-types playground:documents`
 3. git commit ...
 4. git push ...
 
 How does this look if we introduce a lock file?

henrist:
> That being said, can someone explain to me the practical difference between manually removing/adding/editing a lock file vs manually running a command when new types need to be generated? I get it with libraries, but this case seems different to me as we want the sdk's to always reflect the documentation.

> when new types need to be generated

How would you know when that is needed? The "lock" would help you keep track of that.

> sdk's to always reflect the documentation

You can't really guarantee this unless you move the code generation to runtime.

> practical difference

You loose track of what was the source of the current types (unless you encode that in the generated code).

You can't easily check if something has changed, as you have to check the _result_ of a process, instead of just see that the _input_ has changed.

Same input should produce same result.

I'd say the code generation should be done _every time_ the code is built. This ensure nobody starts modifying this manually, and keeps it in sync.

If you fetch a potentially updated remote resource every build you break reproducability.

If the spec changes in a way that is not supported by code generation, you have suddenly broken the SDK and can no longer build it successfully. If you keep track of the input you can keep the previously working state while fixing the code generation at the same time you introduce the changes.

henrist:
> If we completely automate this, won't we still have to go in and make a PR to update the lock file? during that time it can race, causing the version to be outdated.

Not sure why a race matters here. You will never have perfect consistency here, there can always at best be eventual consistency. If this scenario happens you will get a new PR with the latest changes.

For me - key parts are:

* Reproducability - don't get surprises, keep the build stable
* Source control of what goes into the build - by locks, copying files etc, depending on what is available
* Automation - don't let the generated types diverse (no manual editing)

> How does this look if we introduce a lock file?

You must split the process. Think of this as a just a regular dependency. You can update the dependency and you can install it. Same here - you can point it to a newer version and you can generate code for the active version. The last part should happen as part of the build pipeline, while the first should happen as part of new commits, preferably automated.

andersfylling:
> How would you know when that is needed? The "lock" would help you keep track of that.

If you introduce a new field, you know when you need to update any sdks/openapi representations. How does the lock help you keep track of that? I'm not asking about how it can be compared, cause comparing a sum is straight forward. I mean in what case will the lock not change when you update a type? Is it just to automatically give us a todo list of repos that must be updated, or should they auto update when the sum changes? cause then we still have the issue of type names causing breaking changes.

> If you fetch a potentially updated remote resource every build ...

I'm realizing now how annoying it would be for contributors if their PR fails cause they haven't generated code for some random service. And this allows us to isolate the update process to one PR.

:+1: 

> If the spec changes in a way that is not supported by code generation, you have suddenly broken the SDK and can no longer build it successfully. If you keep track of the input you can keep the previously working state while fixing the code generation at the same time you introduce the changes.

:+1: 


I see my perspective was a bit different, I agree this is a must have. I appreciate that you explained this further @tommythorsen @henrist

andersfylling:
I also wonder if a lock file per team would be best? Then teams can go in and approve updates as they need manually, just to avoid it accumulating in one PR. The code generation can run for a single specified service. Just to divide the workload.

What if we start of with a whitelist of services that wants code gen (us). Then it should be somewhat straight forward to extend it?

henrist:
> If you introduce a new field, you know when you need to update any sdks/openapi representations. How does the lock help you keep track of that? I'm not asking about how it can be compared, cause comparing a sum is straight forward. I mean in what case will the lock not change when you update a type? Is it just to automatically give us a todo list of repos that must be updated, or should they auto update when the sum changes? cause then we still have the issue of type names causing breaking changes.
> 
> >

The lock represents a input to the build, similar as with regular dependencies. We're calling it a lock here, but it's basicly the same thing (strictly speaking "lock" would be more correct if we target a range, to freeze the exact version).

The "JSON file defining the contract of the service" is a (development) dependency like others. You use it as input to the build to produce generated code.

> Is it just to automatically give us a todo list of repos that must be updated

I think your view of perspective is a bit opposite of what I'd think myself. You (this SDK) depend on some other resource, but it doesn't imply that the resource knows you are depending on it, and neither should it need to know. But of cause it should avoid to do things in a breaking manner. Other companies might even build their own SDKs you don't know of.

> I also wonder if a lock file per team would be best? Then teams can go in and approve updates as they need manually. The code generation can run for a single specified service. Just to divide the workload.

I don't see why human approvals are needed for this. The updates should flow as automatically as possible.

One reason I could see for keeping different versions for different "team resources" would be that breaking changes in the service contract might block other updates (e.g. due to types needing to be updated in tests). However I'm not sure this will actually be that large of a problem (and it should not stay in that state), and I'd rather not optimize for it personally. It adds complexity (and time to the build I guess) I'm not sure we need?

This is why I'm also a bit nitty about our service specs (and consider lots of changes breaking), since changes such as names of components, structure etc, will lead to different generated code.

> 
> What if we start of with a whitelist of services that wants code gen (us). Then it should be somewhat straight forward to extend it?

Yes, if we move forward with code generation I think it makes sense to not do everything at once.

Btw - you might be interested to dig a bit into how e.g. GitHub generates their SDK. Some nice read from e.g. https://github.blog/2020-04-09-from-48k-lines-of-code-to-10-the-story-of-githubs-javascript-sdk/

andersfylling:
> I don't see why human approvals are needed for this. The updates should flow as automatically as possible.

You earlier mentioned that the sdk code gen could break. Sometimes it might be buggy but still succeed. I think in the beginning it would be nice to manually approve it to see how it works, before we fully automate it. We can then also choose to delay releases in case we are planning several breaking changes in a short time period.

Edit: having it on per service would help decrease the cognitive load of verifying.

henrist:
What I mean is that manual verification shouldn't be _required_. That will not make things scale, at least when we eventually have this for lots of different languages. We have to trust the code/type generation. Since this isn't established immediately it might be convenient to not automate everything just yet, however this should be a responsibility of maintainers of this repo, not the teams that owns the specs/services.

If the code/types from this library is wrong due to the spec, then it's not this library that is wrong. It's the spec that needs to be fixed.

andersfylling:
> What I mean is that manual verification shouldn't be required. That will not make things scale, at least when we eventually have this for lots of different languages

I think the expectations of this PR has gotten a bit high. We can work on this in iterations. The main point here is to provide a foundation for code generation to reduce dev time. Whenever we introduce features/changes, we have a responsibility to update the SDK. This helps reduce time spent on that, and possibly reduce bugs as well.

I believe that having simply a command that auto-generates the types for you, which is most of the work, will help scale everything a lot. It currently takes at most 2 seconds to generate the types for /documents, compared that to having manually implement everything by hand. You can cover a lot more ground, in a significantly shorter time. We only have a handful of SDKs, there isn't a need to fully automate this, though I think it would be awesome if we did at a later point.


> I don't see why human approvals are needed for this. The updates should flow as automatically as possible.

> We have to trust the code/type generation.

To an extent yes. I don't agree to blindly trust it. That level of automation will be up to @cognitedata/developer-tooling to take on, as the SDK will then become purely their responsibility. Currently we help out by verifying our part of it is correct. Complete automation moves all of the responsibility over to them.

My point is that I shouldn't aim for that in this PR. This is to reduce the workload for grep. From what I can see it's normal for individual teams to maintain their types/endpoints in SDKs. Which is usually fairly alright when a service is in V1 and not in beta/alpha/playground.

> If the code/types from this library is wrong due to the spec, then it's not this library that is wrong. It's the spec that needs to be fixed.

I agree it's the goal / ideal, and it's where I want us to be as well. But I don't think it's fair to say the first PR for this is bug free.


-----------

> One reason I could see for keeping different versions for different "team resources" would be that breaking changes in the service contract might block other updates (e.g. due to types needing to be updated in tests). However I'm not sure this will actually be that large of a problem (and it should not stay in that state), and I'd rather not optimize for it personally. It adds complexity (and time to the build I guess) I'm not sure we need?

I tried adding a lockfile per service. I realized the lockfile for playground is huge. +23k lines, around 90KB (~40KB minified). So I agree it will have to be one lockfile per package. However, I won't introde that until every service in said package have moved to code generation. I simply can't accept the risk introduced of generating it for all services.

Build time was fine though, the openapi => typescript takes a second ish. Less so when you don't update the lockfile.


> Btw - you might be interested to dig a bit into how e.g. GitHub generates their SDK. Some nice read from e.g. [github.blog/2020-04-09-from-48k-lines-of-code-to-10-the-story-of-githubs-javascript-sdk](https://github.blog/2020-04-09-from-48k-lines-of-code-to-10-the-story-of-githubs-javascript-sdk/)

sweet :+1:

PR #741, opened 2022-01-21
fix(documents): extension should be outside sourceFile

henrist:
The code for updating the auto-generated file is not in master yet, so I've patched it manually. See https://github.com/cognitedata/cognite-sdk-js/pull/740


henrist:
The integration tests are failing due to a flake

andersfylling (approved):

PR #749, opened 2022-02-02
feat(documents): add document search API to stable

tommythorsen:
Only the /documents/search endpoint is available for the time being.

Jira: [SBS-4652](https://cognitedata.atlassian.net/browse/SBS-4652)


packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,369 @@
...
+   */
+  securityCategories?: number[];
+  metadata?: Record<string, string>;
+}
+
+/**
+ * GeoJson representation of a geometry.
+ */
+export interface DocumentGeometry {
+  /** Type of the shape. Currently we support "polygon", "linestring" and "point". */
```
henrist: Are the types in this file hand-written or somehow generated? If the latter case it would be nice with some instructions for how to keep it up-to-date with future API changes.

I notice that the comments here is not the same as in the API docs, and the values here is no longer correct. It would also be nice if the type names matched the schema used for the API, to reduce cognitive load and reduce type name changes in future automation (as clients of this library will have to migrate their code if they import the types).

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+                },
+              },
+            },
+          ],
+        });
+      const response = await client.documents.search({});
+      const geoLocation = response.items[0].item.geoLocation;
+      expect(geoLocation?.type).toEqual('Point');
+      expect(geoLocation?.coordinates).toBeDefined();
+      // @ts-ignore
```
henrist: I think all these ts-ignore rules are needed due to lack of proper types?

If we would need it for these tests, wouldn't a consumer of the library also need it? This makes me think this should be redundant and something we should fix.

I see this in the types, that should be fixed:

```
export type DocumentShapeCoordinates = any[];
```

The API docs behind this is likely the root cause - since it has `type: array` with no further definition

packages/stable/src/api/documents/documentsApi.ts
```diff
@@ -0,0 +1,49 @@
...
+   * });
+   * ```
+   */
+  public search = (
+    query: DocumentSearchRequest
+  ): Promise<DocumentSearchResponse> => {
+    return this.searchDocuments<DocumentSearchResponse>(query);
+  };
+
+  private async searchDocuments<ResponseType>(
```
andersfylling, line 36: You have to add:

`return this.addToMapAndReturn(response.data, response);`

This is so users can get headers or other data. It's built into all the helper methods of the library, but when you write custom methods you need to specify this manually.

packages/stable/src/api/documents/documentsApi.ts
```diff
@@ -0,0 +1,49 @@
...
+import { BaseResourceAPI, CDFHttpClient, MetadataMap } from '@cognite/sdk-core';
+
+import {
+  Document,
+  DocumentSearchResponse,
+  DocumentSearchRequest,
+} from '../../types';
+
+export class DocumentsAPI extends BaseResourceAPI<Document> {
+  constructor(...args: [string, CDFHttpClient, MetadataMap]) {
```
andersfylling: I think the constructor can be removed as it doesn't do anything beyond the default behavior.

andersfylling (changes requested):
Few things that have to/can be updated.

packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,369 @@
...
+}
+
+export interface DocumentSort {
+  /**
+   * List of fields to sort by, currently only supports 1 field.
+   * Syntax: `["<fieldname>:asc|desc"]`. Default sort order is `asc` with short syntax `["<fieldname>"]`.
+   *
+   * @example ["externalId:desc"]
+   */
+  sort?: string[];
```
henrist: I also notice this is the sort structure for the playground API, not the v1 API. I think we need to redo this file to get things correct and in sync.

andersfylling: we could also write one monster request that touches every field. If the api returns 400, then we know its a little off.

packages/stable/src/api/documents/types.ts
tommythorsen: It's hand-written. Well, I copied the playground types and modified them to match the v1 spec.

I suppose I could remove these comments. I would not like to manually have to keep them in sync with the spec. That sounds like a lot of work for little benefit.

packages/stable/src/api/documents/documentsApi.ts
tommythorsen: Oh, maybe I could instead use one of those helper methods that have it built in? I'll look into that

packages/stable/src/api/documents/types.ts
henrist: I think this is a strong argument for generating the types (#740).

We should focus on the developer experience, so comments should be included. However I'd rather exclude comments than including the wrong comments, if we have to choose.

Since we are not there yet (generating types), can we try to be compatible with it? To preserve developer experience we don't want to rename these later.

Since I spotted at least another defect in the types - maybe @andersfylling can help generating this for a one-shot now? I think this was initially generated for the playground API. That way we get proper names, fix other defects to the types, and will be compatible with future code gen as far as I understand without introducing breaking changes.

packages/stable/src/api/documents/documentsApi.ts
andersfylling: Maybe you can use `callSearchEndpoint`.

packages/stable/src/api/documents/types.ts
andersfylling: The v1 spec is funny. I can generate the entire playground api types, without issues, but the v1 might have some weird cyclic reference. Causing a never ending lookup loop (suspicion). It keeps generating forever, ending up with a crash due to memory usage. Even 4GiB isn't enough.

Might be that the script just have to check the type record before exploring a type to avoid redundancy.

tommythorsen: The new filter specification is definitely cyclic. In particular, the bool filters refer back to the common filter type

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
tommythorsen: The cause for this was simply that the `coordinates` field is optional. I added a `!` behind it, like so:

```js
expect(geoLocation?.coordinates![0]).toEqual(2.324);
```

This removes the need for the `@ts-ignore`

I have not done anything about the `any[]`. Looking at the implementation in the `unstructured-search` repo, it looks like this depends a bit on the type of shape, so it'll get a bit complex if we want to do it exact. I think the `any[]` here is about on par with how it's defined in the contract.

packages/stable/src/api/documents/types.ts
tommythorsen: I have fixed the sort structure now, and added an integration test for it. There are also integration tests for the other major things.

packages/stable/src/api/documents/documentsApi.ts
tommythorsen: Done

packages/stable/src/api/documents/documentsApi.ts
tommythorsen: I ended up following your first suggestion, which seems to work fine

andersfylling (approved):
lgtm

andersfylling:
can you add the following to codeowners?

```
/packages/stable/src/api/documents @cognitedata/grep
/packages/stable/src/__tests__/api/documents @cognitedata/grep
```

https://github.com/cognitedata/cognite-sdk-js/pull/750

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
henrist: The `any[]` is aligned with the service spec, but I think we should look later to see if we can improve this (in the service spec, and also here). Currently a client have _no idea_ what it might contain (as it is defined as anything) without looking into external resources, while it actually can only hold

```ts
type Position = [number, number]
type Coordinates =
  | Position // Point
  | Position[] // MultiPoint, ListString
  | Position[][] // MultiLineString, Polygon
  | Position[][][] // MultiPolygon
```

This would narrow everything down to numbers and a few lists, and making things a lot simpler to reason about.

.github/CODEOWNERS
```diff
@@ -1,3 +1,6 @@
 * @cognitedata/developer-tooling
 
 /packages/playground/src/api/documents @cognitedata/grep
+/packages/playground/src/__tests__/api/documents @cognitedata/grep
+/packages/stable/src/api/documents @cognitedata/grep
+/packages/stable/src/__tests__/api/documents @cognitedata/grep
```
henrist: This does not match the tests files. Maybe it's easily solved by added a `*` at the end of those?

andersfylling: Ah, yes, true. It isn't a documents folder, under __tests__, it's a file matching `documents*`

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -0,0 +1,100 @@
+// Copyright 2022 Cognite AS
+
+import CogniteClient from '../../cogniteClient';
+import { setupLoggedInClient } from '../testUtils';
+
+describe('Documents integration test', () => {
```
henrist, line 6: What is the canonical source of this test data, in case something gets mixed up (e.g. more files added to the same project that would break these tests) - and for me to actually understand what we are testing against?

packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export type DocumentSortItem = DocumentFilterProperty & {
+  order: DocumentSortOrder;
+};
+
+export interface DocumentSearchLimit {
+  limit?: number;
+}
+
+export interface Highlight {
```
henrist: Is it intentional that this does not have the `Document` prefix as it has in the service spec?

packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export interface DocumentSort {
+  sort?: DocumentSortItem[];
+}
+
+export type DocumentSortOrder = 'asc' | 'desc';
+
+export type DocumentSortItem = DocumentFilterProperty & {
+  order: DocumentSortOrder;
```
henrist: This should be optional? `order?:` ?

packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,189 @@
...
+};
+
+export interface DocumentSearchLimit {
+  limit?: number;
+}
+
+export interface Highlight {
+  name?: string[];
+  content?: string[];
```
henrist: These should not be optional. However the spec is also missing required attribute for these.

packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export interface Document {
+  id: CogniteInternalId;
+  externalId?: CogniteExternalId;
+  title?: string;
+  author?: string;
+  createdTime?: DocumentTimestamp;
```
henrist: This is not optional

packages/stable/src/api/documents/types.ts
```diff
@@ -0,0 +1,189 @@
...
+  assetIds?: CogniteInternalId[];
+  labels?: DocumentLabel[];
+  geoLocation?: DocumentGeoJsonGeometry;
+  datasetId?: CogniteInternalId;
+  securityCategories?: number[];
+  metadata?: Record<string, string>;
+}
+
+export interface DocumentGeoJsonGeometry {
+  type?: DocumentShapeType;
```
henrist: type is not optional

henrist (changes requested):

packages/stable/src/api/documents/types.ts
andersfylling: Yes, default value is `asc`.

.github/CODEOWNERS
tommythorsen: I have updated this now

packages/stable/src/api/documents/types.ts
tommythorsen: fixed

packages/stable/src/api/documents/types.ts
tommythorsen: fixed

packages/stable/src/api/documents/types.ts
tommythorsen: fixed

packages/stable/src/api/documents/types.ts
tommythorsen: I've added the Document prefix now

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
tommythorsen: Most of the tests will work regardless of what kinds of files are in the project, as long as there are a few files there. The only exception is maybe `search with query` which requires that there is at least one file with the word "test" somewhere in the project.

andersfylling (approved):

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
henrist: These integration tests have external dependencies and assumtions that is not explicit, so it would help a lot to give a better understanding of this. Have you added new data? Are there seed data available somewhere else? What would we do in case someone wipe the project data? There is a hidden cost here. Might be there are things in the SDK that covers this, but I'm not sure since I haven't touched this here before. I can't find any documentation for integration tests in this repo from a quick look for it.

henrist (approved):
If types are generated later I'm sure something will be considered breaking from the developer perspective whatever effort we put now. I think the work done so far is very nice!

Currently we are only focusing on happy paths in the tests. With the Documents API we have a few failure cases that's not likely to be that rare. E.g. using a filter that is not supported for a given propertly. I can't understand from this work how that must be handled by a client. Ideally there should be tests for this, and perhaps documentation somewhere.

packages/stable/src/api/documents/types.ts
tommythorsen: I'm fixing this in the contract first: https://github.com/cognitedata/service-contracts/pull/1614

tommythorsen: fixed

tommythorsen:
I think this PR is ready to be merged. @cognitedata/developer-tooling: would one of you be able to review?

pedroharbs (approved):

henrist:
Since this now include generated code, I advise this is not merged before the code generation logic has been merged. As otherwise it would refer to code and process that does not exist.

andersfylling:
> As otherwise it would refer to code and process that does not exist.

The `gen` part of the filename?

Forgot I had added a gen comment.

But these can be removed. I just forgot about them. Note that if we do get codegen later on and we want the `gen` tag in the filename, removing it now will make it harder to diff later on.

andersfylling:
I'm surprised I didn't see any failing tests in my PR before I merged..

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -0,0 +1,454 @@
+// Do not modify this file!
+// It was generated by the command "yarn generate-types:stable:documents".
+// Instead update the code generation logic or the service contracts.
```
tommythorsen:
```suggestion
// It has been automatically generated and should not be edited manually
// If you need to make changes to this file, please reach out to tommy.thorsen@cognite.com or anders.fylling@cognite.com
```

tommythorsen:
> Since this now include generated code, I advise this is not merged before the code generation logic has been merged. As otherwise it would refer to code and process that does not exist.

I'm not so fond of keeping PRs open for extended periods of time. Instead, I have modified the comment in the top of the `types.gen.ts` file so that instead of referencing a tool and process that does not exist, it says to reach out to me or Anders.

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -0,0 +1,454 @@
+// Do not modify this file!
+// It has been automatically generated and should not be edited manually
+// If you need to make changes to this file, please reach out to tommy.thorsen@cognite.com or anders.fylling@cognite.com
```
tommythorsen:
```suggestion
// If you need to make changes to this file, please reach out to tommy.thorsen@cognite.com
// or anders.fylling@cognite.com
```

PR #752, opened 2022-02-03
feat: oidc integration tests

pedroharbs:
- Integration tests for the OIDC auth feat.
- Moving of api key tests for better organization.

To attend: [SDKTP-144]( https://cognitedata.atlassian.net/browse/SDKTP-144)


andersfylling:
Doesn't seem like the tests are working?

janisharbs:
@andersfylling I fixed little problems with the tests, now should be working, let's see the CI.

andersfylling:
I suggest removing the api key folder, it's going to be dead code/unable to run anyways. It will be a lot less to review, and there shouldn't be a need to have an explicit oidc folder. So the team/service defined tests should probably not have to be altered, just the setup of the test client. Though I can have forgotten a part of this.

But reviewing 4k loc additions seems a bit much to me for moving from api keys to oidc - any way we can reduce/simplify this?

andersfylling:
also note that I'm not a code owner so my approval of this will still keep you blocked.

henrist:
As far as I can see most files in `oidc` folder is originally duplicated from `apikey`, however there are more files in the `apikey` dir. (I did `meld apikey oidc` to ease the review of the added lines.)

Seems the files that are not really integration tests are left in the old dir. I suggest moving them (if that's the right thing to do) to avoid this confusion during the PR. Ideally to review such large change as this it's a great help if the commits are organized so it can be reviewed logically step by step.

BugGambit:
IMO, this PR shouldn't be merged. It doesn't only introduce tests for OIDC but is testing all APIs twice.
If we want to test OIDC, we should only test the login flow.

PR #764, opened 2022-02-25
chore: export ADFS auth class [release]

vegardok:
With v6+, apps are responsible for handling auth and there aren't any publicly available ADFS libraries.


harish0489-cognite (approved):
LGTM.

PR #769, opened 2022-03-02
chore: ran code gen script for playground:documents

arnels:
Ran the code gen script for playground:documents to include the latest API definitions.

Jira task: https://cognitedata.atlassian.net/browse/SBS-4803


packages/playground/src/api/documents/types.gen.ts
```diff
@@ -1,3 +1,7 @@
+// Do not modify this file!
```
henrist, line 1: Can you change this to be same as https://github.com/cognitedata/cognite-sdk-js/blob/4394e60273fa4fe33be731289beb2b09dac57f31/packages/stable/src/api/documents/types.gen.ts#L1-L4 ?

henrist (changes requested):

packages/playground/src/api/documents/types.gen.ts
arnels: Fixed

henrist (approved):

henrist:
Flaky tests - I triggered a rerun

andersfylling:
I wonder if the context team did a breaking change in their api. The error complains about a type issue in the request.

andersfylling:
@cognitedata/contextualization could you take a look?

henrist:
> I wonder if the context team did a breaking change in their api. The error complains about a type issue in the request.

Ah, good spot! I just assumed flakyness a bit too quick.

Maybe the test itself is originally wrong/invalid?

andersfylling:
> 

Yeah, integration tests in this SDK are testing a bit too much internal behavior rather than just if the endpoint + types are correct (+ timeout issues). So they often get flaky.

These used to work before, I believe @cyriaquebrousse implemented them.

henrist:
I didn't pay attention, but next time use "fix" for this change and it will also auto-release. :+1:

PR #770, opened 2022-03-03
test: temporary skip the tests for the annotations endpoints

arnels:
Temporary skip the tests for the annotations endpoints as they are currently not working, blocking other [PRs](https://github.com/cognitedata/cognite-sdk-js/pull/769).

### DIFF
--- a/packages/playground/src/__tests__/api/annotationsApi.int.spec.ts
+++ b/packages/playground/src/__tests__/api/annotationsApi.int.spec.ts
@@ -48,7 +48,7 @@ const FILE_FILTER: AnnotationFilterProps = {
   annotatedResourceIds: [{ externalId: ANNOTATED_FILE_EXTERNAL_ID }],
 };
 
-describe('Annotations API', () => {
+describe.skip('Annotations API', () => {
   let client: CogniteClientPlayground;
   let stableClient: CogniteClient;
   const createdAnnotationIds: InternalId[] = [];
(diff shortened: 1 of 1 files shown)


arnels:
@cognitedata/contextualization

henrist:
I suggest you add a comment in the code as well to describe the situation and why it is (temporary) disabled

arnels:
> I suggest you add a comment in the code as well to describe the situation and why it is (temporary) disabled

Added

packages/playground/src/__tests__/api/annotationsApi.int.spec.ts
```diff
@@ -48,6 +48,10 @@ const FILE_FILTER: AnnotationFilterProps = {
   annotatedResourceIds: [{ externalId: ANNOTATED_FILE_EXTERNAL_ID }],
 };
 
+/**
+ * Annotations API tests are temporarily skipped due to the tests failing, which blocks non-related releases.
+ * The Annotation endpoint has changed behavior, leading these tests to produce 400 errors.
+ */
```
henrist: nit: I'd say that this type of text should be a regular comment (`//` or `/*`) and not a "JSDoc" (or TSDoc or whatever we would call this) using `/**`. A JSDoc or similar would be appropriate if we were to _document_ an interface, symbol or something similar. This is internal (implementation) details to this file only describing _why_ the code is as it is.

```suggestion
// Annotations API tests are temporarily skipped due to the tests failing, which blocks non-related releases.
// The Annotation endpoint has changed behavior, leading these tests to produce 400 errors.
```

henrist (approved):

pedroharbs (approved):

PR #789, opened 2022-04-29
fix: fixing missing token when call sdk methods using Promises

pedroharbs:
Fixing missing token problem when call SDK methods with multiple promises, reported by @peternic.

### DIFF
--- a/packages/core/src/baseCogniteClient.ts
+++ b/packages/core/src/baseCogniteClient.ts
@@ -133,8 +133,11 @@ export default class BaseCogniteClient {
 
     this.httpClient.set401ResponseHandler(async (_, retry, reject) => {
       try {
+        const previousToken = this.retrieveTokenValueFromHeader();
+
         const newToken = await this.authenticate();
-        if (newToken && newToken !== _.headers['api-key']) {
+
+        if (newToken && newToken !== previousToken) {
           retry();
         } else {
           reject();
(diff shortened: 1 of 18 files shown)


packages/core/src/baseCogniteClient.ts
```diff
@@ -147,11 +145,18 @@ export default class BaseCogniteClient {
...
+        API_KEY_HEADER,
+        options.getToken().toString()
+      );
+    }
   }
 
   public authenticate: () => Promise<string | undefined> = async () => {
     try {
-      const token = await this.getToken();
+      const token = this.getToken;
```
cognite-ornellas: shouldn't be a method call?

janisharbs: I removed the necessity to be a method. We aren't receiving a Promise now, just a string. If you wanna we can connect to discuss more regarding it. The tests are passing, and I did some tests with the reported scenario and now is working. @cognite-ornellas

packages/core/src/baseCogniteClient.ts
```diff
@@ -35,7 +35,7 @@ export interface ClientOptions {
   /** URL to Cognite cluster, e.g 'https://greenfield.cognitedata.com' **/
   baseUrl?: string;
   project: string;
-  getToken: () => Promise<string>;
```
roligheten: How does this work with the "OIDC authentication using code authorization w pkce" example? Afaik for token refresh to work the getToken needs to be able to resolve a new token if the old one has expired. Without promises here I don't think it is possible to write a single function that can resolve a new access token externally as the function must resolve immediately. 

Case where this is important: Long-lived clients authenticated via client credentials such as APIs

cognite-ornellas (approved):

packages/core/src/baseCogniteClient.ts
janisharbs: You're right, but, the problem are exactly here when you try to get logged using api-keys using only promises without async/await. I did solve this problem on the new feature that will be on the next release, using the new auth-wraper that will do automatically the token refresh and of course the api key on this way too. Maybe we can discuss more about in a meeting, please text-me on slack and we can schedule.

cognite-ornellas (approved):

cognite-ornellas (approved):

PR #791, opened 2022-05-09
feat: show warning message on version mismatch

samir-ajdarpasic:
Depending on the repository where @cognite/sdk is used it can happen that it resolves to an older @cognite/sdk-core version and it's really hard to debug or know what's going on. This way we will at least see a warning telling us that the resolved version is wrong.

### DIFF
--- a/packages/stable/src/cogniteClient.ts
+++ b/packages/stable/src/cogniteClient.ts
@@ -3,9 +3,11 @@ import {
   accessApi,
   apiUrl,
   BaseCogniteClient,
+  ClientOptions,
   RetryValidator,
 } from '@cognite/sdk-core';
-import { version } from '../package.json';
+import { CogniteAPIVersion } from '@cognite/sdk-core/src/utils';
+import { version, dependencies } from '../package.json';
 import { AssetMappings3DAPI } from './api/3d/assetMappings3DApi';
 import { Files3DAPI } from './api/3d/files3DApi';
 import { Models3DAPI } from './api/3d/models3DApi';
@@ -172,12 +174,31 @@ export default class CogniteClient extends BaseCogniteClient {
   private geospatialApi?: GeospatialAPI;
   private documentsApi?: DocumentsAPI;
 
+  constructor(options: ClientOptions, apiVersion: CogniteAPIVersion = 'v1') {
+    super(options, apiVersion);
+    this.checkCoreVersionMatch();
+  }
+
   protected get version() {
     return version;
   }
 
+  protected get coreVersion() {
+    // @ts-ignore linter is complaining even though version is protected and can be accessed
+    return super.version;
+  }
+
   protected getRetryValidator(): RetryValidator {
     return retryValidator;
+  }
+
+  private checkCoreVersionMatch() {
+    // check if sdk-core version matches with the one in the dependencies
+    if (!dependencies['@cognite/sdk-core'].includes(this.coreVersion)) {
+      console.warn(
+        `VERSION MISMATCH! \n The resolved @cognite/sdk-core version ${this.coreVersion} doesn't match the required version ${dependencies['@cognite/sdk-core']} from @cognite/sdk. This might lead to unexpected behavior and bugs.`
+      );
+    }
   }
 
   protected initAPIs() {
--- a/packages/stable/src/cogniteClient.ts
+++ b/packages/stable/src/cogniteClient.ts
@@ -6,6 +6,7 @@ import {
   ClientOptions,
   RetryValidator,
 } from '@cognite/sdk-core';
+import { satisfies } from 'semver';
 import { CogniteAPIVersion } from '@cognite/sdk-core/src/utils';
 import { version, dependencies } from '../package.json';
 import { AssetMappings3DAPI } from './api/3d/assetMappings3DApi';
@@ -194,7 +195,7 @@ export default class CogniteClient extends BaseCogniteClient {
 
   private checkCoreVersionMatch() {
     // check if sdk-core version matches with the one in the dependencies
-    if (!dependencies['@cognite/sdk-core'].includes(this.coreVersion)) {
+    if (!satisfies(this.coreVersion, dependencies['@cognite/sdk-core'])) {
       console.warn(
         `VERSION MISMATCH! \n The resolved @cognite/sdk-core version ${this.coreVersion} doesn't match the required version ${dependencies['@cognite/sdk-core']} from @cognite/sdk. This might lead to unexpected behavior and bugs.`
       );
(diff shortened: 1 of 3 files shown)


rek:
This would be very useful to merge. Perhaps make a new commit with a better message. Eg: bump version or docs

janisharbs:
@samir-ajdarpasic Can you change the commit message to a small one? Actually with this one the linting step shall not pass.

packages/stable/src/cogniteClient.ts
```diff
@@ -172,14 +174,33 @@ export default class CogniteClient extends BaseCogniteClient {
...
 
+  private checkCoreVersionMatch() {
+    // check if sdk-core version matches with the one in the dependencies
+    if (!dependencies['@cognite/sdk-core'].includes(this.coreVersion)) {
+      console.warn(
+        `VERSION MISMATCH! \n The resolved @cognite/sdk-core version ${this.coreVersion} doesn't match the required version ${dependencies['@cognite/sdk-core']} from @cognite/sdk. This might lead to unexpected behavior and bugs.`
+      );
+    }
+  }
+
```
henrist: I'm not sure if this will work as you expect:

If a user installs `@cognite/sdk` and also an incompatible version of `@cognite/sdk-core` the core package will exist multiple places in the dependency tree due to how version resolution is done. With NPM you will not easily be able to put an incompatible version for this import, while with yarn you can force it with resolutions.

For the (normal?) situations where the user accidentally installs an incompatible version but don't override the transitive version, this logic will not catch that situation.

Also note that you probably want to do a semver check here. The value in package.json is a range, while the `coreVersion` is a specific version. While the `includes` work at the point of release, it will break on next patch release even though it would be considered a compatible release (and version resolution will pick it).

Have you seen any other projects that tries to solve the same problem? Maybe there is some useful inspiration around.

samir-ajdarpasic: @henrist The monorepos had special tsconfig setup and it would always start searching for the package from the root `node_modules`. That problem is fixed but there's still cases where people don't delete old `node_modules` directories and sdk resolves the wrong sdk/core version. Since the upgrade we had multiple cases of these errors and it was hard to debug, that's how this PR came to life. 

About the versioning: is there a package that we already use or should I install one. I knew my `includes` implementation is too simple but I just wanted to get this conversation started.

henrist: I don't think these situations will be caught by this? The import from this file should be compatible _unless_ there is a manual resolution to override the allowed range.

However from the _users code_ (consider `src/myfile.ts`) it might be seeing a different version that is not compatible, since it will resolve relative to root and not from the library code.

Consider this:

```bash
yarn add @cognite/sdk
yarn add @cognite/sdk-core@4.0.0
```

An import to `@cognite/sdk-core` from your source dir will give you 4.0.0, while the import from `@cognite/sdk` (the one you are modifying) will see a different version (4.1.1 as of writing).

If you use a resolution to override the core version, yarn will show a warning during package install. I think that is the only case you will detect with that code.

Regarding package for version comparison: I think `satisfies` from https://www.npmjs.com/package/semver would do what you want.

samir-ajdarpasic: @henrist I completely understand your point. But while upgrading the `@cognite/react-container` package to use the new SDK version I encountered this problem. The reason was multiple `tsconfig` settings and resolutions in `package.json`. It took me so long to find out what was going on because my editor would point me to the correct version, but when the code would run it would resolve an old one. And I've seen this behaviour with apps outside of the monorepo too because they all could have different configurations and if they don't clean the `node_modules` but just install it doesn't resolve the core version correctly. 

That is the reason why I think something like this that checks at runtime could be a helpful thing. But if you think that we don't want this then I'll close the PR

rek: I don't think we should close this. This work does fix _a_ problem, though potentially not all _problems_ as you pointed out. Either way it would still be beneficial for a certain set of users/situations in close to the current state.

samir-ajdarpasic: Updated the PR to use `semver` for version check

samir-ajdarpasic: @henrist Do you think it is ok now, can we merge this in ?

packages/stable/src/cogniteClient.ts
```diff
@@ -3,9 +3,12 @@ import {
   accessApi,
   apiUrl,
   BaseCogniteClient,
+  ClientOptions,
   RetryValidator,
 } from '@cognite/sdk-core';
-import { version } from '../package.json';
+import { satisfies } from 'semver';
+import { CogniteAPIVersion } from '@cognite/sdk-core/src/utils';
```
henrist, line 10: Let's make this an actual export rather than hooking into the internals?

yarn.lock
```diff
@@ -3919,6 +3919,11 @@ geojson@^0.5.0:
   resolved "https://registry.yarnpkg.com/geojson/-/geojson-0.5.0.tgz#3cd6c96399be65b56ee55596116fe9191ce701c0"
   integrity sha1-PNbJY5m+ZbVu5VWWEW/pGRznAcA=
 
+geojson@^0.5.0:
```
henrist, line 3922: I wouldn't expect the lockfile to change like this

henrist (commented):
The code itself is fine. I'm neutral on what this actually solves, but if you feel certain this proves useful then go for it! You'll need an approval from @cognitedata/developer-tooling though.

yarn.lock
samir-ajdarpasic: I just did `yarn` when I cloned the repo, I guess this was missing in the lock but I can revert this change.

bdimitrijoski:
@samir-ajdarpasic This is a very old PR. What is the status? Can we merge it or close it?

samir-ajdarpasic:
@bdimitrijoski Sorry, I've seen your comment just now. Yeah it's a very old PR, been waiting for an approval. Not sure if it's worth merging it now or not ?

andersfylling:
If an SDK requires a specific core version, why can't this be specified in the package.json for the SDK? Wouldn't that be a better approach, or is it not possible?

samir-ajdarpasic:
> If an SDK requires a specific core version, why can't this be specified in the package.json for the SDK? Wouldn't that be a better approach, or is it not possible?

@andersfylling That is specified. The problem is that there could be multiple versions of core installed and the resolution to the correct version failed a couple of times. The bug was extremely difficult to find, so then I made this PR to at least warn us that it didn't resolve to the correct version. Luckily the bug has not come up for a long time, so don't know how useful this PR is.

andersfylling:
Sounds like a dependency resolution that should be fixed in yarn/npm. Do you have any build system that handles any kind of caching? like bazel?

This issue may come back when there are changes to the core package, but for now it probably won't be a concern. I'll leave this for @cognitedata/development-experiences to decide.

samir-ajdarpasic:
@andersfylling Yeah it was a dependency resolution problem. Happened in the `applications` monorepo which uses bazel but we fixed it there. Then I saw it happen to apps that were installing `@cognite/react-container` from npm. But like I said it's been so long since the last occurrence. I will close the PR.

PR #792, opened 2022-05-11
feat(documents): content and list endpoints + stricter types

andersfylling:
https://cognitedata.atlassian.net/browse/SBS-4994

### DIFF
--- a/packages/stable/src/__tests__/api/documentsApi.int.spec.ts
+++ b/packages/stable/src/__tests__/api/documentsApi.int.spec.ts
@@ -51,7 +51,7 @@ describe('Documents integration test', () => {
 
   beforeAll(async () => {
     client = setupLoggedInClient();
-    fileId = getFileId(client);
+    fileId = await getFileId(client);
   });
 
   test('search with limit 1', async () => {
--- a/packages/stable/src/__tests__/api/documentsApi.int.spec.ts
+++ b/packages/stable/src/__tests__/api/documentsApi.int.spec.ts
@@ -3,12 +3,34 @@
 import CogniteClient from '../../cogniteClient';
 import { setupLoggedInClient } from '../testUtils';
 import { DocumentSearchResponse } from '@cognite/sdk-stable/dist';
+import { TextEncoder } from 'util';
 
 describe('Documents integration test', () => {
   let client: CogniteClient;
+  let fileId: number;
 
   beforeAll(async () => {
     client = setupLoggedInClient();
+
+    // ensure we have a file for testing
+    try {
+      const resp = await client.documents.list({
+        limit: 1,
+      });
+      fileId = resp.items[0].id;
+    } catch (error) {
+      const fileContent = new TextEncoder().encode('test data');
+      const file = await client.files.upload(
+        {
+          name: 'test.txt',
+          mimeType: 'text/plain',
+        },
+        fileContent,
+        false,
+        true
+      );
+      fileId = file.id;
+    }
   });
 
   test('search with limit 1', async () => {
@@ -147,5 +169,19 @@ describe('Documents integration test', () => {
       const resp = await client.documents.preview.pdfTemporaryLink(document.id);
       expect(resp.temporaryLink).toBeDefined();
     });
+  });
+
+  test('list', async () => {
+    const response = await client.documents.list({
+      limit: 1,
+    });
+    expect(response.items).toHaveLength(1);
+    expect(response.items[0]).toBeDefined();
+    expect(response.items[0].id).toBeDefined();
+  });
+
+  test('content', async () => {
+    const response = await client.documents.content(fileId);
+    expect(response).toBeDefined();
   });
 });
--- a/packages/stable/src/__tests__/api/documentsApi.int.spec.ts
+++ b/packages/stable/src/__tests__/api/documentsApi.int.spec.ts
@@ -9,34 +9,44 @@ const getFileId = async (
   client: CogniteClient,
   deadline: number = 15000
 ): Promise<number> => {
-  try {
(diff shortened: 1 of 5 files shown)


henrist:
If we are about to do breaking changes now I'd also like to see `DocumentCountAggregate` renamed, as it is confusing since it will not be present in the upcoming aggregate endpoint (only from search endpoint). I'm not convinced changes like this warrant a major version change though.

andersfylling:
Feel free to take a look, ill hold this off until then

packages/stable/src/api/documents/documentsApi.ts
```diff
@@ -57,4 +74,15 @@ export class DocumentsAPI extends BaseResourceAPI<Document> {
 
     return this.addToMapAndReturn(response.data, response);
   }
+
+  private async documentContent<ResponseType>(
+    id: CogniteInternalId
+  ): Promise<ResponseType> {
+    const response = await this.get<ResponseType>(this.url(`${id}/content`), {
```
henrist: Shouldn't `ResponseType` always be string, so it's not needed?

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -3,12 +3,34 @@
...
+      const file = await client.files.upload(
+        {
+          name: 'test.txt',
+          mimeType: 'text/plain',
+        },
+        fileContent,
+        false,
+        true
+      );
+      fileId = file.id;
```
henrist: Can we make this loop until the document is actually available? The current way this will work is that first time (for an empty test project) the tests will unexpectedly fail, and it will not be clear why it fails, while it is just waiting for the pipeline to process the file. Then "suddenly" it will start working again. We should avoid such behaviour in our tests.

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -3,12 +3,55 @@
...
+
+  while (deadline > 0) {
+    try {
+      const resp = await client.documents.list({
+        limit: 1,
+      });
+      return resp.items[0].id;
+    } catch (error) {
+      const timeout = 3000;
+      await new Promise((resolve) => setTimeout(resolve, deadline));
```
henrist: Did you intend to use `timeout` here?

Note that on the last iteration it will now sleep and then always fail since deadline will be `<= 0`

```suggestion
      await new Promise((resolve) => setTimeout(resolve, timeout));
```

Can we also check `error` properly? We should only catch the not found errors, and not e.g. network errors.

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -3,12 +3,55 @@
...
+const getFileId = async (
+  client: CogniteClient,
+  deadline: number = 15000
+): Promise<number> => {
+  try {
+    const resp = await client.documents.list({
+      limit: 1,
+    });
+    return resp.items[0].id;
+  } catch (error) {
```
henrist: I think it's unexpected to get an exception for an empty result. Is this because you get an exception for index out of bounds in the list lookup? I think it would be better to check the length instead.

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
andersfylling: oh, good you spotted the use of deadline.

henrist:
Can we improve the final commit message (changelog entry) - as "new endpoints" doesn't say much?

Maybe `feat(documents): add list and content methods, update types from spec` ?

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -3,12 +3,55 @@
...
+  while (deadline > 0) {
+    try {
+      const resp = await client.documents.list({
+        limit: 1,
+      });
+      return resp.items[0].id;
+    } catch (error) {
+      const timeout = 3000;
+      await new Promise((resolve) => setTimeout(resolve, timeout));
+      deadline -= timeout;
```
henrist: I suggest you move this up, because now you are doing: "get, get, sleep, get, sleep, fail"

Better to have something like "get, sleep, get, sleep, get, fail"

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -3,12 +3,55 @@
...
+    }
+  }
+
+  throw new Error(
+    'unable to ingest a file into the documents service within time'
+  );
```
henrist: If you're open for an alternative :smile: 

```suggestion
  // Use a previous uploaded file/document if available,
  // otherwise ingest a new file.

  async function lookup() {
    const resp = await client.documents.list({
      limit: 1,
    });
    return resp.items.length > 0 ? resp.items[0].id : undefined;
  }

  const found = await lookup();
  if (found) {
    return found;
  }

  const fileContent = new TextEncoder().encode('test data');
  await client.files.upload(
    {
      name: 'test.txt',
      mimeType: 'text/plain',
    },
    fileContent,
    false,
    true
  );

  // It takes some time for a uploaded file to become available
  // in the documents API.

  let timeLeft = deadline;
  while (timeLeft > 0) {
    const timeout = Math.min(timeLeft, 3000);
    await new Promise((resolve) => setTimeout(resolve, timeout));
    timeLeft -= timeout;

    const foundOnRetry = await lookup();
    if (foundOnRetry) {
      return foundOnRetry;
    }
  }

  throw new Error(
    'unable to ingest a file into the documents service within time'
  );
```

henrist (approved):

PR #793, opened 2022-05-12
feat(documents): aggregate api

andersfylling:
This includes a change to BaseResourceAPI to enable the use of a operation-specific item type for aggregates, which is not the same as the REST resource. This follows our API design guide that states that aggregations can use pagination, which we use in the documents API. In the documents API only the `allUniqueValues` supports pagination.


packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -215,4 +218,30 @@ describe('Documents integration test', () => {
...
+    const resp = await client.documents.aggregate({
+      filter: {
+        equals: {
+          property: ['type'],
+          value: 'PDF',
+        },
+      },
+      aggregate: 'count',
+    });
+    expect(resp.items).toBeGreaterThanOrEqual(1);
```
henrist:
```suggestion
    expect(resp.items).toHaveLength(1);
```

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -215,4 +218,30 @@ describe('Documents integration test', () => {
...
+    expect(resp.items[0].count).toBeGreaterThanOrEqual(1);
+  });
+
+  test('document aggregate unique', async () => {
+    const resp =
+      await client.documents.aggregate<DocumentsAggregateUniqueValuesResponse>({
+        aggregate: 'uniqueValues',
+        properties: [{ property: ['mimeType'] }],
+      });
+    expect(resp.items).toBeGreaterThanOrEqual(0);
```
henrist:
```suggestion
    expect(resp.items.length).toBeGreaterThanOrEqual(1);
```

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -215,4 +218,30 @@ describe('Documents integration test', () => {
...
+        },
+      },
+      aggregate: 'count',
+    });
+    expect(resp.items).toBeGreaterThanOrEqual(1);
+    expect(resp.items[0].count).toBeDefined();
+    expect(resp.items[0].count).toBeGreaterThanOrEqual(1);
+  });
+
+  test('document aggregate unique', async () => {
```
henrist:
```suggestion
  test('document aggregate uniqueValues', async () => {
```

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
```diff
@@ -385,4 +385,52 @@ describe('Documents unit test', () => {
...
+          value: 'PDF',
+        },
+      },
+      aggregate: 'count',
+    });
+    expect(resp.items).toHaveLength(1);
+    expect(resp.items[0]).toEqual({ count: 3456 });
+  });
+
+  test('document aggregate unique', async () => {
```
henrist:
```suggestion
  test('document aggregate uniqueValues', async () => {
```

packages/stable/src/api/documents/documentsApi.ts
```diff
@@ -65,6 +67,27 @@ export class DocumentsAPI extends BaseResourceAPI<Document> {
     return this.documentContent(id);
   };
 
+  /**
+   * []()
```
henrist: ?

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -1,9 +1,6 @@
 // Do not modify this file!
-// It has been automatically generated and should not be edited manually
-// If you need to make changes to this file, please reach out to tommy.thorsen@cognite.com
-// or anders.fylling@cognite.com
-
-import { CogniteInternalId, CogniteExternalId } from '@cognite/sdk-core';
+// It was generated by the command "yarn generate-types stable:documents".
+// Instead update the code generation logic or the service contracts.
```
henrist: This should not be changed (yet)?

henrist (commented):
Should there also be some tests for allUniqueValues? Do we want to actually fetch all pages for the user in this case, or otherwise see if we can use the generator pattern (?) used elsewise for this?

andersfylling:
> Do we want to actually fetch all pages for the user in this case, or otherwise see if we can use the generator pattern (?) used elsewise for this?

@henrist 
Not sure if we can use the generator here. BaseResource#listEndpoint utilizes makeAutoPaginationMethods. However, the BaseResource is current configured to use a Document[] as it's generic type. This fits the need for /search and /list, not so much for aggregate.

It would work better if aggregate was moved to it's own resource base inherited class, perhaps we can try that. The SDK would instead look like:
```ts
documents.aggregate.count(..)
documents.aggregate.uniqueValues(..)
documents.aggregate.allUniqueValues(..)
```

I split these into separate methods as each case uses different request and response structures, so it seems natural. Perhaps this should have been 3 sub-endpoints (`/documents/aggregate/count`, etc.) instead of using a constant in the request payload?

henrist:
@andersfylling 

I support the idea of trying out a subresources in the SDK, and like the idea of different methods.

Regarding the API for the service I think it's fine to stay as-is, and I'd probably do the same again. In my head we have a separation between low level and high level abstractions here. The API to the service is low level. It changes infrequent and is pretty resistant against changes. It's more about passing operations with data around, and in thise case REST-ish calls. It's difficult to be (and stay) precise in this context. This SDK is more like a mix between both, and in this case it would be a higher level abstraction since we don't map 1-1 on the API and have additional behaviour such as the auto-pagination.

packages/core/src/baseResourceApi.ts
```diff
@@ -312,11 +322,11 @@ export abstract class BaseResourceAPI<ResponseType> {
       .reduce((a, b) => [...a, ...b], []);
   }
 
-  protected addNextPageFunction<QueryType extends FilterQuery>(
-    endpoint: ListEndpoint<QueryType, ResponseType[]>,
-    cursorResponse: CursorResponse<ResponseType[]>,
+  protected addNextPageFunction<QueryType extends FilterQuery, Item>(
```
andersfylling, line 325: Item should perhaps default to ResponseType, thought it doesn't seem to have mattered?

packages/stable/src/api/documents/aggregateApi.ts
```diff
@@ -1,53 +1,78 @@
...
+      aggregate: 'allUniqueValues',
+    });
   };
 
-  private async aggregate<
-    Request extends DocumentsAggregateRequest,
-    Response extends DocumentsAggregateResponse
-  >(request: Request): Promise<Response> {
-    const response = await this.post<Response>(this.url(), {
+  protected callAggregateCursorEndpointWithPost = async <
```
andersfylling: private

henrist (approved):

henrist:
@cognitedata/developer-tooling Can you please help review this?

henrist:
Last failures far are due to flakes or issues during integration tests. Will trigger another rerun later if it persists.

cognite-ornellas (approved):

PR #801, opened 2022-05-19
feat: code generation for types based on openapi doc

andersfylling:
# Second draft
Second draft of code generation, see [previous version here](https://github.com/cognitedata/cognite-sdk-js/pull/740). There's a few changes such as:
 - implemented in typescript
 - unit / integration tests
 - new library (seems swagger-typescript-api is currently on "pause")
 - configuration file per service
 - a way to opt in/opt out of the package versionfile
 - ~generated types are sorted for a better git diff~ (bugged feature, wrote a something that sorts the entire json spec)
 - several smaller commands instead of one large command
 - avoids exporting already exported types (codegen per service may re-define an existing type)
 - generate types for inlined requests (with automatic naming - can be toggled off in the config file per service)

# Code (type) Generation
The SDK types should mirror the open api definition. By generating types we avoid human errors and inconsistencies in the SDK.

This also saves developers time, provides a higher likelihood of correctness as well as faster SDK updates. (This may now be automated in the future).

Whenever types generated seems incorrect, the ideal solution is that the openapi spec have to be updated - as this might indicate incorrect schema definitions. Whenever a real bug with the code generation is identified, we reproduce these thanks to the versionfile (formerly referred to as "lockfile") to reproduce the bugged behavior and come up with patches (see https://github.com/cognitedata/cognite-sdk-js/pull/740#pullrequestreview-859316537).

## Concerns

By using open api as a source, we now depend on naming conventions defined there. Changes to existing types is essentially a breaking change as the old type would no longer exist and dependents will fail to compile unless they update their imported types.

Therefore any team/service that wants this feature have to enable it manually, to become aware how this impact their work on the open api contracts.

This is a new feature and should not be assumed to be perfect. Our open API contracts also have no constraints and we don't know how certain features may affect the generated output here. We therefore have this as a opt in feature with the ability to put specific services on pause in case any temporary issues arise.

## Open API Snapshot
We require a snapshot of the open api contracts in order to generate code, this allows us to:
1. verify with github action that generated types are not manipulated / changed by hand within a pull request.
2. reprocude buggy behavior.

Each package (stable, playground, etc.) that supports code generation will have their own snapshot, shared among services. The goal is to continuously update the package snapshot, and run code generation respectively. Automatic version estimate is another problem which is not covered by this pull request - eg. should changes to types be a fix, feature or breaking change in the SDK when automating updates.

A service may specify their own snapshot if there are bugs or issues preventing it from using the package snapshot. This is done by specifying a "path" in the configuration file. This should be considered a last resort as a service is effectively put on "pause" until it's snapshot/config is manually updated.

## Config file
A configuration file is defined for the package and per service. Here a service may tweak settings to achieve desired behavior such as disabling "magic" behavior or specifying a custom open api snapshot.

## Example repo setup
Here the `documents` service uses the package open api snapshot (default), while `vision` uses their own snapshot, the `files` service have not enabled code generation as it has no configuration file (.codegen.json):
```markdown
stable/
├─ src/
│  ├─ .cognite-api-snapshot
│  ├─ types.gen.ts
│  ├─ imports.gen.ts
│  ├─ api/
│  │  ├─ vision/
│  │  │  ├─ .cognite-api-snapshot
│  │  │  ├─ .codegen.json
│  │  │  ├─ types.gen.ts
│  │  ├─ documents/
│  │  │  ├─ .codegen.json
│  │  │  ├─ types.gen.ts
│  │  ├─ files/
```

```json
{
	"service": "documents",
	"filter": {
		"serviceName": "documents"
	},
	"inlinedSchemas": {
		"autoNameRequest": true
	}
}
```

```json
{
	"service": "vision",
	"filter": {
		"serviceName": "vision"
	},
	"inlinedSchemas": {
		"autoNameRequest": false
	},
	"snapshot": {
		"path": "packages/stable/src/api/vision/.cognite-api-snapshot"
	}
}
```

The `codegen generate-types` command will generate types for the package (meaning all configured services; `documents` and `vision`). 

Example:
```console
yarn codegen generate-types --package stable
```

# General code flow
In order to generate types for a package there must be at least one service configured for code generation. Any service folder (under `./packages/{package}/api/{service}`) that contains a `.codegen.json` is considered to be a configured service (use `yarn codegen:configure` to enable code generation for your service).

Types are generated per service, with respect to their config file. The types are stored as a `types.gen.ts` file within the service folder.

A "export file" is created which explicitly exports types from each service. This is done to avoid exporting types with the same name, each service currently have their own definition for shared types.

Say you have two services `documents` and `files`. They both have defined the `LabelList` type. Since openapi schema names are globally unique, we assume that any type defined in multiple services have the same definition, so we keep track of already exported types to avoid re-exporting them.

> Fictive example
```ts
export { LabelList, FilesResponse } from './api/files/types.gen'; // types: LabelList, FilesResponse
export { DocumentsSearchResponse } from './api/documents/types.gen'; // types: LabelList, DocumentsResponse
```

# Commands
> Subject to change!

```markdown
# yarn codegen 
 * generate-types: generates types for all configured services
 * fetch-latest: fetches the latest open api and saves it as a snapshot
 * configure: generates a default config for the service or package
```

# Scenarios
> Using the "documents" service for these examples

## Enable code generation for a package
```console
# configure the package to use cognite open api v1
yarn codegen configure --package my-package --version v1

# fetches the latest open api contract, creating a snapshot
yarn codegen fetch-latest --package my-package

# generate types for package (generates no new types)
yarn codegen generate-types --package my-package
```

Then update the `types.ts` in the package `src` folder to export from `exports.gen.ts`:
```ts
export * from './exports.gen';
```

Update the github action script at `.github/workflows/code-gen.yaml` to run for your new package:
```yaml
package: ["stable", "playground", "my-package"]
```

## Enable code generation for a service
```console
yarn codegen configure --package stable --service documents
```

And generate types for the package and newly configured documents service:
```console
yarn codegen  generate-types --package stable
```

## Disable code generation for a service
For whatever reason the code generation have to be disabled for a service, you simply delete the `codegen.json` file for the respective service(s), and re-run code generation:
```console
rm ./packages/stable/src/api/documents/codegen.json
yarn codegen generate-types --package stable
```

This is useful when a service has just enabling code generation. Sometimes there might be issues in their openapi spec and they want to work on both their openapi contracts and their js-sdk types. The nice effect of generating types based off an openapi spec is that the developer gets a different perspective on their openapi definitions, it may be easier to identify issues when looking at code instead of a large yaml definition.

Update your service config to refer to a open api path:
```json
{
	"service": "documents",
	"filter": {
		"serviceName": "documents"
	},
	"inlinedSchemas": {
		"autoNameRequest": false
	},
	"snapshot": {
		"path": "/tmp/v1.json"
	}
}
```
> Config file is validated whenever you try to generate types

You don't have to run a snapshot command as you are pointing to a local path. The snapshot is simply read from the location when it's required, instead you can generate type right away:
```console
yarn codegen generate-types --package stable
```

# Issues

## Commands: configure & snapshot
~Currently the `yarn codegen:configure` and `yarn codegen:snapshot` is decoupled. It can be argued that the `yarn codegen:snapshot` should always read from a service configuration (when a --service argument is specified). This would help the PR review process as the config file also details where the snapshot originates from (though without a guarantee anyways).~

fixed...

~Currently shared types are not generated by themselves. This mean that any service defined first in the export file will have ownership over a subset of the shared types definitions. If the first service uses a local outdated versionfile, this may cause the SDK to have outdated shared types, which can be a problem that needs manual intervention.~

~This can be fixed by identifying shared types and specified as the first export statement:~
 ~1. generating types for all services once~
 ~2. identify shared types~
 ~3. run type generation from the package global versionfile for identified shared types using a components/schemas filter.~
 ~4. Adjust the first line of the export file to use the shared generated types.~

~This fix may also cause services with oudated shared types, to now have incorrect types. However, I'd argue that we should prefer our shared types to be as up to date as possible and instead push relevant service to stay up to date.~
_Fixed in commit https://github.com/cognitedata/cognite-sdk-js/pull/801/commits/bf96dd94a58c7138cd2bcc6c2f68bc552d4ff9f8_

# Other

Required for this PR to pass:
 - Disable repository/branch requirement for node v12 tests
	A new workflow job called `tests-legacy` was added instead. Code generation doesn't work with node v12, and we won't add support for generating types via node v12 anyways. 

related tasks: SBS-4943, SBS-4945, SBS-4950, SBS-4947, SBS-5107, SBS-5122


.github/workflows/code-gen.yaml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          fetch-depth: 2
+
+      - name: configure node js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+
+      - name: Install Dependencies
+        run: yarn install
```
henrist:
```suggestion
        run: yarn install --frozen-lockfile
```

.github/workflows/code-gen.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+      - name: Run code generation for ${{ matrix.package }} package
+        run: yarn codegen --package=${{ matrix.package }} generate
+      
+      - name: Fix lint issues
+        run: yarn lint:fix --since=master
```
henrist: I don't see why linting is mixed in here? I'd not expect it to be relevant (at least as a separate command).

I suggest to exclude generated files from linting (e.g. in `.eslintignore`) instead of running the linter on the files. Otherwise I think this should be done as part of the generate command, and not require another lint run (I think I prefer to exclude it).

.github/workflows/code-gen.yaml
henrist: Can we also look at the existing workflow files and make them more similar?

.github/workflows/tests.yaml
```diff
@@ -156,7 +156,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12, 14]
+        node-version: [14, 16]
```
henrist, line 159: Let's mark the PR for breaking change?

package.json
```diff
@@ -67,11 +72,16 @@
     "lerna": "^3.22.1",
     "markdown-link-check": "^3.8.1",
     "nock": "^10.0.6",
+    "openapi-types": "^11.0.1",
     "prettier": "^2.5.1",
     "rollup": "^2.22.2",
+    "swagger-typescript-api-nextgen": "^10.0.1",
```
henrist, line 76: I'm a bit reluctant into relying on newly forks. Is `swagger-typescript-api` the right tool for the job?

packages/codegen/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "homepage": "https://github.com/cognitedata/cognite-sdk-js#readme",
+  "main": "dist/index.js",
+  "types": "dist/src/index.d.js",
+  "scripts": {
+    "clean": "rm -rf dist/ docs/ codeSnippets/",
+    "test": "jest --config=../../jest.config.js --testPathPattern=/codegen/",
+    "test:debug": "node --inspect-brk ../../node_modules/.bin/jest --config=../../jest.config.js --testPathPattern=/codegen/ --runInBand",
+    "lint": "eslint 'src/**/*.ts'",
+    "lint:fix": "yarn lint --fix",
+    "watch": "rollup -cw",
```
henrist: clean, rollup, es-check and docs doesn't seem relevant for this package

packages/codegen/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "esCheck": "es-check es5 './dist/index.js'",
+    "docs": "typedoc --options typedoc.js --tsconfig tsconfig.json src/index.ts",
+    "docs:bundle": "yarn docs && mkdir -p ../../docs && cp -r docs/* ../../docs/"
+  },
+  "dependencies": {},
+  "devDependencies": {
+  },
+  "files": [
+    "dist"
+  ]
```
henrist: Probably not relevant as we don't pack (at least shouldn't need to)

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,342 @@
...
+      await fs.writeFile(
+        pathUtil.resolve(this.options.outputDir, CodeGen.outputFileName),
+        fileComment + content
+      );
+    });
+
+    await file.cleanup();
+  };
+
+  // private operationResponseWithoutRef = (operation: OpenApiOperationObject): boolean => {
```
henrist: Cleanup?

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,342 @@
...
+      }
+    }
+
+    return operations;
+  };
+
+  private autoNameInlinedResponses = (): [string, OpenApiSchema][] => {
+    return [];
+    // console.error('inline responses are not currently supported');
+    // process.exit(1);
```
henrist: Cleanup?

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,342 @@
...
+    //  3. copy responses over to schemas (normalise: add "Response" suffix)
+    //  4. process query parameter, add name and add to schemas (little magic)
+    //  1. name inline responses and add to schemas (magic) (normalise: add "Response" suffix)
+    //  2. name inline requests and add to schemas (magic) (normalise: add "Request" suffix)
+    //  6. discard error types (error responses as these aren't needed)
+    //  7. make sure everything is in order (sorted, maybe use prettier?)
+    //  8. run swagger code gen on the filtered open api spec
+    //  9. use swagger's onCreateComponent hook to remove types to be imported from core
+    // 10. update file comment and imports
+    // 11. write to file
```
henrist: Cleanup?

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,342 @@
+// Copyright 2020 Cognite AS
```
henrist: 2022?

packages/codegen/src/commands.ts
```diff
@@ -0,0 +1,269 @@
...
+  public delete = async (options: ConfigFileOptions) => {
+    const directory = await versionFileDirectoryPath(options);
+    const config = new ConfigurationManager({
+      directory: directory,
+    });
+
+    try {
+      await config.delete();
+    } catch (error) {
+      throw new Error('Config file does not exist - did nothing');
```
henrist: This might hide other errors, e.g. a read only file system. I suggest checking if the file exists explicitly rather than assuming all errors are not founds.

packages/codegen/src/configuration.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (!['local', 'global'].includes(config.scope)) {
+      throw new Error('unknown scope specified');
+    }
+  };
+
+  public write = async (
+    config: ConfigurationOptions
+  ): Promise<ConfigurationOptions> => {
+    this.validate(config);
+    const json = JSON.stringify(config, null, 4);
```
henrist: We usually use 2 spaces. Se e.g. `package.json` files and `.editorconfig`

packages/codegen/src/configuration.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    await fs.writeFile(this.path, json, { flag: 'wx' });
```
henrist: Do we care if the file already exists (the `x` flag)? I guess no flags are needed here

packages/codegen/typedoc.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
```
henrist: Is typedoc needed for this package? I think we can remove

packages/codegen/src/versionfile.ts
```diff
@@ -0,0 +1,82 @@
...
+  };
+
+  private createUrl = (): string => {
+    if (typeof this.options.version === 'undefined') {
+      throw new Error('Cognite API version was not configured');
+    }
+    return `https://storage.googleapis.com/cognitedata-api-docs/dist/${this.options.version}.json`;
+  };
+
+  public downloadFromURL = async (url?: string): Promise<OpenApiDocument> => {
```
henrist: nit: URL vs Url (I prefer Url)

packages/codegen/src/versionfile.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+/**
+ * VersionFileManager handles creating and updating a version file.
+ *
+ * A version file is a copy of the open api json spec at a specific point in time.
+ * To keep the version file "small", the irrelvant paths are filtered out before saving.
+ */
+export class VersionFileManager {
+  public static readonly filename = '.cognite-api-version';
```
henrist: I find this name a bit confusing, especially since the "version" field inside the file never changes. And maybe the file shouldn't be hidden by default?

In my head what we are looking at is a snapshot from some time. How about something like `openapi-spec-snapshot.json` ?

packages/codegen/src/versionfile.ts
```diff
@@ -0,0 +1,82 @@
...
+ * VersionFileManagerOptions options for configure the version file.
+ */
+export type VersionFileManagerOptions = {
+  directory: string;
+} & Partial<VersionOption>;
+
+export type LocalVersionFileOptions = {
+  path: string;
+  filename?: string;
+};
```
henrist: nit: I'd call for using interfaces instead of type aliases thoughout, unless you need the power of expressions. See recommendations in https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases

.github/workflows/code-gen.yaml
```diff
@@ -0,0 +1,41 @@
...
+ pull_request:
+    branches:
+      - master
+    types: [opened, edited, synchronize, reopened]
+
+jobs:
+  stable:
+    strategy:
+      matrix:
+        package: ["stable"]
```
henrist: I suggest we include playground as well, so that's not missed later.

Do we need to have a build matrix? I'd think that we could have one command generate types for everything (or delegate if needed) - that would mean we remove the package argument. Maybe each package.json file that needs codegen should have a run-script for it, and use Lerna to delegate?

CONTRIBUTING.md
```diff
@@ -55,6 +55,41 @@ docs(contributing-readme): add example of commit with subject line
...
+A service must have a versionfile (essentially just a copy of the openapi spec) and a configuration (which openapi paths to parse). After
+a versionfile and configuration have been created, a user may update their versionfile, delete it (eg. type generation doesn't work for yuor service), switch between local and global versionfiles, etc. before generating types.
+
+### Versionfile
+> Newly configured defaults to a local versionfile scope.
+
+A versionfile has two scopes; local (service specific) or global (package specific). Services that links to a package global versionfile are all updated at the same time. If you require to update types for your service only, independently from other services in the package, it's recommended to use a local versionfile.
+```sh
+# configure your service to have a local versionfile
+> yarn codegen:configure --package=stable --service=documents --versionfile-scope=local update
```
henrist: Since this is supposed to be a script (`sh` above) the `> `s should be removed.

scripts/codegen/configure.ts
```diff
@@ -0,0 +1,45 @@
...
+    } catch (error) {
+        console.error("ERROR: unable to create config: " + error)
+    }
+}).command("update", "update a configuration", (y) => y, async (argv) => {
+    try {
+        let config = argv;
+        if (typeof argv['versionfile-scope'] !== "undefined") {
+            config["scope"] = argv['versionfile-scope'];
+        }
+        await cmd.update(config);
```
henrist: This will reinitialize the file on every update. I suggest we patch the file instead.

scripts/codegen/configure.ts
```diff
@@ -0,0 +1,45 @@
...
+            config["scope"] = argv['versionfile-scope'];
+        }
+        await cmd.create(config);
+        console.log("ok");
+    } catch (error) {
+        console.error("ERROR: unable to create config: " + error)
+    }
+}).command("update", "update a configuration", (y) => y, async (argv) => {
+    try {
+        let config = argv;
```
henrist: This will include lots of other values that is present in `argv`. I think it's better to select what you expect and build a new object.

scripts/codegen/codegen.ts
```diff
@@ -0,0 +1,28 @@
...
+    const nodeVersion = process.versions.node.split(".");
+    const majorVersion = parseInt(nodeVersion[0]);
+    if (majorVersion < 16) {
+        console.error("NodeJS version must be v16 or higher");
+        process.exit(1);
+    }
+
+    try {
+        await cmd.generate({...argv});
+        console.log("ok");
```
henrist: remove?

scripts/codegen/notes.txt
```diff
@@ -0,0 +1,45 @@
+# generate a new versionfile in the stable package. This file points to openapi v1 and has created a configuration for the documents service.
```
henrist: Remove file?

tsconfig.common.json
```diff
@@ -1,4 +1,19 @@
 {
+  // Most ts-node options can be specified here using their programmatic names.
+  "ts-node": {
+    //"extends": "ts-node/node16/tsconfig.json",
+    // It is faster to skip typechecking.
+    // Remove if you want ts-node to do typechecking.
```
henrist: Let's remove all these comments?

tsconfig.common.json
```diff
@@ -1,4 +1,19 @@
...
+    "transpileOnly": true,
+    "files": true,
```
henrist: nit: I don't think this is needed?

scripts/codegen/codegen.ts
```diff
@@ -0,0 +1,28 @@
+import { CodeGenCommand } from '../../packages/codegen/src';
```
henrist: Just an idea: I think it would be nice to move all this codegen scripts into a single cli entrypoint on the codegen package

packages/stable/src/exports.gen.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 Cognite AS
+export {} from './types.gen';
```
henrist: Something odd here

packages/stable/src/exports.gen.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright 2020 Cognite AS
```
henrist: Should we add a `linguist-generated=true` rule in `.gitattributes` to match on these autogenerated files? Similar as how baker files etc are done in other repos

packages/codegen/src/configuration.ts
henrist: It would be nice to add training newlines as well, according to the rule in `.editorconfig`. Especially since this file is legit to edit manually.

packages/codegen/src/commands.ts
```diff
@@ -0,0 +1,338 @@
...
+
+  public update = async (options: ConfigFileUpdateOptions) => {
+    const directory = await versionFileDirectoryPath(options);
+    const config = new ConfigurationManager({
+      directory: directory,
+    });
+
+    try {
+      await config.delete();
+      await config.write(this.defaultConfig(options));
```
henrist: I suggest overwriting the file instead. This helps certain editors keep the correct (inode) reference to the file

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+    ) as OpenApiDocument;
+  }
+
+  /**
+   * runSwaggerGenerate generates types for all component schematics.
+   */
+  private runSwaggerGenerator = async (): Promise<void> => {
+    const fileComment = `// Do not modify this file!
+// It was generated by the command "yarn codegen".
+// Instead update the code generation logic or the open api constracts.\n\n`;
```
henrist:
```suggestion
// Instead update the code generation logic or the open api contracts.\n\n`;
```

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+  isReferenceObject,
+  OpenApiPathsObject,
+  filterPaths,
+  operationsInPath,
+  ReferenceWalker,
+} from './openapi';
+import { ConfigurationOptions } from './configuration';
+
+/**
+ * PathFilter creates a new set of openapi paths that pass the filter implementation.
```
henrist: This doesn't look to belong here

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+    schema?: StringFilter;
+  };
+  outputDir: string;
+} & Omit<ConfigurationOptions, 'filter'>;
+
+export class CodeGen {
+  static outputFileName = 'types.gen.ts';
+  readonly versionFile: OpenApiDocument;
+
+  constructor(versionFile: OpenApiDocument, readonly options: CodeGenOptions) {
```
henrist: `options` unused?

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+
+/**
+ * PathFilter creates a new set of openapi paths that pass the filter implementation.
+ */
+export type StringFilter = (str: string) => boolean;
+
+export const PassThroughFilter: StringFilter = (): boolean => true;
+
+export const ServiceNameFilter = (service: string): StringFilter => {
+  const r = new RegExp('^/api/.+/projects/{project}/' + service);
```
henrist: This actually matches for "service name prefix". May I suggest we append `($|/)` to make it match on equals instead?

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+  filter: {
+    path: StringFilter;
+    schema?: StringFilter;
+  };
+  outputDir: string;
+} & Omit<ConfigurationOptions, 'filter'>;
+
+export class CodeGen {
+  static outputFileName = 'types.gen.ts';
+  readonly versionFile: OpenApiDocument;
```
henrist: I think in this context it would make more sense to just name this "specFile" or something. I find the "version" term more confusing than helpful

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+
+  constructor(versionFile: OpenApiDocument, readonly options: CodeGenOptions) {
+    this.versionFile = JSON.parse(
+      JSON.stringify(versionFile)
+    ) as OpenApiDocument;
```
henrist: Why is this serialized and deserialized? It seems to be the same types input and output

scripts/codegen/versionfile.ts
```diff
@@ -0,0 +1,37 @@
+import { VersionFileCommand } from '../../packages/codegen/src';
+import { yargsBase } from './command';
+
+const cmd = new VersionFileCommand();
+
+const parser = yargsBase().options({
+    package: { type: 'string', demand: true,  describe: "Js sdk package (stable, beta, etc.)" },
+    service: { type: 'string', demand: false, describe: "REST service to generate types for" },
+    version: { type: 'string', demand: false, describe: "Cognite api version (v1, playground, etc.)" },
+    "from-path": { type: 'string', demand: false, describe: "Specify a local openapi spec instead" },
```
henrist: This is nice so it's easy to work with the spec and see what changes are caused by it!

scripts/codegen/configure.ts
```diff
@@ -0,0 +1,45 @@
...
+        }
+        await cmd.update(config);
+        console.log("ok");
+    } catch (error) {
+        console.error("ERROR: unable to update config: " + error)
+    }
+}).command("delete", "delete a configuration", (y) => y, async (argv) => {
+    try {
+        await cmd.delete({...argv});
+        console.log("ok");
```
henrist: Remove?

henrist: See other places too

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -48,120 +43,20 @@ export type DocumentListRequest = DocumentListFilter &
   DocumentListLimit &
   DocumentCursor;
 
-export interface DocumentSearch {
```
henrist, line 51: Just an open question: Is it possible (easy) to sort the types by name or something, to avoid types moving around?

CONTRIBUTING.md
```diff
@@ -55,6 +55,41 @@ docs(contributing-readme): add example of commit with subject line
 
 A commit hook makes sure the syntax is followed. Automated commit messages such as `Merge pull request` are handled.
 
+## Type generation
+This SDK support generating typescript types from the Cognite open api specification. The idea is to use openapi as a source of truth. Any incorrect or missing types should be fixed/added in the openapi spec instead of manually adjusting the generated types. This also helps to keep documentation up to date.
+
+The scripts prefix `codegen` are reserved for type generation. The core logic is found in `packages/codegen` with a deeper explenation.
+
+To quickly setup type generation for your service you may run the generate command with `--auto`:
+> Note: using `--auto` will setup a local versionfile and not the package global versionfile. You can later switch to the global versionfile once you have confirmed type generation works correctly for your service (you may also have to update your openapi spec).
```
henrist: I suggest we swap this - always prefer global file but opt in to local file as a strategy for conflict resolution.

packages/codegen/README.md
```diff
@@ -0,0 +1,28 @@
...
+Generate typescript types/interfaces from the Cognite open api contract. Only json files supported. 
+
+The current version aims to generate the simplest representations. Endpoints and tests may require some unique or opinionated logic and are therefore not currently covered.
+
+## Flow
+
+To run code generation for a package the script behaves as follows:
+
+ 1. Identify any services with a config file.
+ 2. Evaluate the relevant lockfile for identified services.
```
henrist: rephrase lockfile

packages/codegen/README.md
```diff
@@ -0,0 +1,28 @@
...
+ 3. Generate relevant types/interfaces.
+ 4. Write types to a types.gen.ts file.
```
henrist: In general - can we add ``` ` ``` around filenames like this?

packages/codegen/README.md
```diff
@@ -0,0 +1,28 @@
...
+ 5. Create an export statement for each service, that avoids exporting existing types.
+ 6. Write export statements to a types.gen.ts file at the package level.
+
+See codegen.ts#main for high level understanding of the flow.
```
henrist: This reference looks incorrect (and I guess it's a bit redundant)

packages/codegen/rollup.config.js
```diff
@@ -0,0 +1,44 @@
+import typescript from '@rollup/plugin-typescript';
```
henrist: We don't use the output of rollup for this package so maybe just remove?

packages/codegen/src/configuration.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright 2020 Cognite AS
+import { promises as fs } from 'fs';
+
+export type ConfigurationOptions = {
+  version: string;
+  service: string;
+  autoNameInlinedRequest: boolean;
+  scope: string;
```
henrist:
```suggestion
  scope: 'local' | 'global';
```

And maybe consider changing to `service` and `package` as scopes? I find `local` and `global` a bit confusing

packages/codegen/src/commands.ts
```diff
@@ -0,0 +1,338 @@
...
+  };
+}
+
+export class CodeGenCommand {
+  public generate = async (options: PackageOption & Partial<ServiceOption>) => {
+    if (typeof options.service !== 'undefined') {
+      await this.generateForSingleService({
+        ...options,
+        service: options.service,
+      });
```
henrist: This doesn't have any effect?

henrist: I'm not sure if we should even support generating for only one service. It adds complexity to figuring out shared types (which I think is incorrect now if we were to only generate for one service).

packages/codegen/src/commands.ts
```diff
@@ -0,0 +1,338 @@
...
+      (schemaName) => sharedTypes.includes(schemaName)
+    );
+    return generatedTypeNames;
+  };
+
+  private createExportStatement = (
+    fromDirectory: string,
+    types: string[],
+    blacklist: string[]
+  ): [string, string[]] => {
```
henrist: I'd recommend returning an object rather than array. It helps a lot to understand what data there is without look at the implementation.

Something like:

```ts
private createExportStatement = (
    fromDirectory: string,
    types: string[],
    blacklist: string[]
  ): { code: string, skipped: string[] } => {
    // ...
  }

// and then
const { code } = createExportStatement(...)
```

CONTRIBUTING.md
```diff
@@ -55,6 +55,41 @@ docs(contributing-readme): add example of commit with subject line
...
+
+```sh
+yarn codegen --package=stable --version=v1 --service=documents --auto generate
+```
+
+A service must have a versionfile (essentially just a copy of the openapi spec) and a configuration (which openapi paths to parse). After
```
henrist: I think it's a bit "hidden" in the text _which_ endpoints types are actually generated for. Can we make that more clear? Ideally I think we could shorten down all this text around.

Can a "service" start by only generating types for a single endpoint, or is it "all or nothing"?

henrist (commented):
This is a lot of work hence quite a few comments. I haven't finished looking through the "code generation" code yet. I'd like to point out that this brings a lot of complexity (and 1700+ lines of code, excluding fixtures/generated), and I wonder if we could have solved it a bit simpler with less complexity initially. I want to challenge you to see if there is some code/text that can be stripped down/simplified without loosing essential functionality.

The CLI in general looks nice. Maybe we can have it all through one entrypoint, so `yarn codegen` could hold the important documentation and access to all commands?

I think the feedback is coming a bit late here. It would be nice if we could arrange feedback earlier in the process or maybe do some pair programming or design talk earlier to drive some discussions.

packages/codegen/src/codegen.ts
andersfylling: Seems its a general issue in the repo, it's automatically done by eslint. see .eslintrc.js, I'm leaving it as is as it's a seperate issue.

CONTRIBUTING.md
andersfylling: A service has to, currently, generate types for all service owned endpoints.

I made the config/StringFilter option in mind so we could extend this if we wanted to.

packages/codegen/src/configuration.ts
andersfylling: I like this idea

.github/workflows/code-gen.yaml
andersfylling: generating that export file. I don't really use templates, so it seems easier to just run prettier as it automatically fixes this.

packages/codegen/src/codegen.ts
andersfylling: Ay, I just noticed that the .eslintrc.js doesn't update it if there is already a year specified. Updating comments.

CONTRIBUTING.md
andersfylling: I just discovered there's a `console` option. Adjusted.

packages/stable/src/exports.gen.ts
andersfylling: Ah. It's because code has only been generated for only one service. This gets populated as soon as we have codegen for an additional service, so I'm not sure it's worth the extra effort of making generated code nice when only one service has it enabled.

.github/workflows/code-gen.yaml
henrist: Can we just accept the output the code generation produces?

packages/codegen/src/codegen.ts
andersfylling: isn't `$` for end of string identification? perhaps `\/?` instead.

henrist: That wont change anything since you dont match until the end of the string, so it will match the same as before

packages/stable/src/api/documents/types.gen.ts
andersfylling: Yeah, the sorting option for the codegen library is buggy. I figure I can just sort the entire openapi object recursively, just need to check performance first..

CONTRIBUTING.md
andersfylling: Wasn't that my suggestion earlier today - or am I misunderstanding something?

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,352 @@
...
+export type StringFilter = (str: string) => boolean;
+
+export const PassThroughFilter: StringFilter = (): boolean => true;
+
+export const ServiceNameFilter = (service: string): StringFilter => {
+  const r = new RegExp('^/api/.+/projects/{project}/' + service);
+  return (path: string): boolean => r.test(path);
+};
+
+export type CodeGenOptions = {
```
andersfylling: I think this could be simplified.

package.json
andersfylling: The author has been inactive since February. I'm uncertain why. There were a few features that weren't published yet, so I made the switch. I wanted sort types (broken) and auto loading the prettier config.

andersfylling:
@henrist are you counting lines of code or just lines in typescript files? I get around 1.2k loc.

CONTRIBUTING.md
henrist: Might be

henrist:
> @henrist are you counting lines of code or just lines in typescript files? I get around 1.2k loc.

For the current tree:

```bash
git diff origin/master --numstat | grep -v "\.gen\." | grep -v "\.json" | grep -v "yarn.lock" | grep -v ".cognite-api"
```

```
1       1       .eslintrc.js
41      0       .github/workflows/code-gen.yaml
1       1       .github/workflows/tests.yaml
1       0       .npmrc
35      0       CONTRIBUTING.md
1       1       README.md
28      0       packages/codegen/README.md
174     0       packages/codegen/src/__tests__/testCodeGen.uint.spec.ts
44      0       packages/codegen/src/__tests__/testOpenApi.uint.spec.ts
29      0       packages/codegen/src/__tests__/testVersionfile.int.spec.ts
51      0       packages/codegen/src/__tests__/testVersionfile.uint.spec.ts
316     0       packages/codegen/src/codegen.ts
338     0       packages/codegen/src/commands.ts
64      0       packages/codegen/src/configuration.ts
4       0       packages/codegen/src/index.ts
125     0       packages/codegen/src/openapi.ts
42      0       packages/codegen/src/utils.ts
81      0       packages/codegen/src/versionfile.ts
2       1       packages/stable/src/types.ts
28      0       scripts/codegen/codegen.ts
10      0       scripts/codegen/command.ts
45      0       scripts/codegen/configure.ts
37      0       scripts/codegen/versionfile.ts
1       1       scripts/versionAndPush.sh
```

```bash
| awk '{a += $1} END {print a}'
# 1499
```

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,316 @@
...
+  /**
+   * generateTypes
+   *
+   * generate all types with a given versionfile and configuration. Returns name of all generated types afterwards.
+   */
+  public generateTypes = async (): Promise<string[]> => {
+    this.filterPaths();
+    this.removeSchemasWithoutPathRelations();
+    this.copyResponsesToSchemas();
+    this.copyQueryParametersToSchemas();
```
henrist: I'm not a large fan of code like this. It makes the flow of operations and effects harder to follow, and e.g. breaks in the case of method being called twice and such, since it's building state in the object.

I'd recommend being explicit about the values you pass around, as in removing `this.versionFile` at least. I'm neutral on doing mutations on the values in the methods though - that should be fine as long as we are working on a deep copy.

henrist: As an example here it was tricky to understand that the `this.versionFile` used in `runSwaggerGenerator` is _not_ the same as what the object was initialized with. I'd say that in the context of `runSwaggerGenerator` that variables has a bad and confusing name. It could be a lot more clear if the signature was something like `private runSwaggerGenerator = async (transformedSpec: OpenApiDocument) =>` instead

I also think it would be more clear if the side effects (such as writing `types.gen.ts`) was not hidden so deep and mixed with other operations. It's also confusing in regard to `generateTypesFromSchemas` vs `generateTypes`. Invoking only the first will not do what the second is doing, but both end up writing a types file (in the same place with different content?).

packages/codegen/src/__tests__/testCodeGen.uint.spec.ts
```diff
@@ -0,0 +1,174 @@
+// import path from 'path';
```
henrist: Having `test` in the name is a bit redundant as it's `.spec.` that matters. So I suggest renaming these files to just `codeGen.uint.spec.ts` etc)

packages/codegen/src/codegen.ts
henrist: Forget my comment - I didn't know we could declare class members with `readonly` in a constructor parameter

packages/codegen/src/codegen.ts
henrist: If your intent is to deep copy maybe you can use structuredClone instead: https://developer.mozilla.org/en-US/docs/Web/API/structuredClone

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,316 @@
...
+  operationsInPath,
+  ReferenceWalker,
+} from './openapi';
+import { ConfigurationOptions } from './configuration';
+
+export type StringFilter = (str: string) => boolean;
+
+export const PassThroughFilter: StringFilter = (): boolean => true;
+
+export const ServiceNameFilter = (service: string): StringFilter => {
```
henrist: nit: These are just functions but looks like classes or types with their capital first letter - `passThroughFilter` and `serviceNameFilter`

packages/codegen/src/codegen.ts
```diff
@@ -0,0 +1,316 @@
...
+
+      this.versionFile.components!.schemas![typeName] = schema;
+    }
+
+    for (const [typeName, schema] of this.autoNameInlinedRequests()) {
+      if (typeName in this.versionFile.components!.schemas!) {
+        console.error(
+          `unable to use automatic name for inlined request type. "${typeName}" already exists in the component schemas`
+        );
+        process.exit(1);
```
henrist: I think throwing an exception is better here - since you're not really in the CLI level here

packages/codegen/src/commands.ts
```diff
@@ -0,0 +1,338 @@
...
+  ): Promise<void> => {
+    const download = async (): Promise<OpenApiDocument> => {
+      if (typeof options['from-path'] !== 'undefined') {
+        return await vfm.downloadFromPath({ path: options['from-path'] });
+      }
+      return await vfm.downloadFromURL();
+    };
+
+    const document = await download();
+    await vfm.write(document);
```
henrist: nit: If you can I'd recommend using ternary operator over control flow - it often simplifies things (and notice how to check for undefined):

```suggestion
    const document =
      options['from-path'] !== undefined
        ? await vfm.downloadFromPath({ path: options['from-path'] })
        : await vfm.downloadFromURL();

    await vfm.write(document);
```

packages/codegen/src/openapi.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+export function operationsInPaths(
+  paths: OpenApiPathsObject
+): OpenApiOperationObject[] {
+  const filtered = filterPaths(paths);
+  return Object.values(filtered)
+    .map(operationsInPath)
+    .reduce((acc, v) => acc.concat(v), []);
+}
```
henrist: These seem unused?

.github/workflows/code-gen.yaml
andersfylling: Is the package argument then implicit via lerna? => Would that cause a command to affect all packages at the same time?

henrist: It would cause the package argument to go away, since you will run the codegen on one package at a time

henrist: Probably just finish what's began now instead of doing too much changes to this.

CONTRIBUTING.md
andersfylling: From what I can tell, it is. Perhaps you were thinking of something additional to my suggestion earlier? I've updated the PR description to reflect my suggestion so please take a look when you get back from vacay :sun_with_face:

andersfylling: From what I can see, it is. Perhaps you were thinking about something additional?

I've updated the PR description and I'm in the progress of updating the code. Please take a look when you're back from vacay :sun_with_face:

packages/codegen/src/commands.ts
andersfylling: > which I think is incorrect now if we were to only generate for one service

I don't think we will be generating types for only one service in the future anyways.

packages/codegen/src/openapi.ts
andersfylling: See codegen.ts line 88

packages/codegen/src/codegen.ts
andersfylling: it's a bit of sugar, might be cleaner to just drop it tbf

.github/workflows/code-gen.yaml
andersfylling: Just out of curiosity.

so basically `cd ./packages/stable && yarn codegen ...`? 

And then `yarn codegen generate` triggers on all packages?

packages/codegen/src/versionfile.ts
andersfylling: Yeah, snapshot is a better fit. Updated

.github/workflows/code-gen.yaml
henrist: See how the other commands like build and lint works

.github/workflows/code-gen.yaml
andersfylling: IT has to be formatted at some point. I'm not sure how to execute prettier on the string before saving it, so this seemed easier. `yarn lint...` has to be run anyways for a PR to be approved.

henrist: Why does it have to be formatted? You can just place the file in .eslintignore

packages/codegen/src/codegen.ts
andersfylling: `generateTypesFromSchemas` will only look at the schemas. While `generateTypes` will try to derive more types than just those defined in `schemas`.

andersfylling: I don't think schemas are altered in `generateTypes`, only new schemas are added (unless unused ones are filtered out).

scripts/codegen/codegen.ts
andersfylling: I tried using the CommandModule as you suggested. However, I feel more confused now. Especially when running the help argument for a sub command. So I put this change on ice, perhaps we can take a look together at a later point as I'm finding the behavior a bit weird right now?

.github/workflows/code-gen.yaml
andersfylling: I like that idea of disabling linting for generated files in general, however that means this wont add the copyright comment at the top of the file anymore (I figure).

henrist: If you'd like to preserve that I'm sure it's easy to just add that line manually, rather than having to call into the code formatter. I don't like that you have to run two commands when working on this, as you won't get the instant feedback of what changes when you regenerate code, and the (lint fix) command might change other files you are working on.

andersfylling: Sure, it's probably not a big deal that the copyright statement is outdated after a year or more, so ill update it.

package.json
henrist: I see. I'm a bit afraid that we will run into the same problem with the fork. I think if really needed we could always fork this out outself (should be fine since this is a dev dep), however I'm asking the question if we are choosing the right tool or if there are other more well-maintained alternatives that would suit us better.

.github/workflows/code-gen.yaml
```diff
@@ -0,0 +1,41 @@
+name: Code Generation
```
henrist: nit: let's rename this file to `codegen.yaml` so we don't have both `codegen` and `code-gen` ?

scripts/codegen/codegen.ts
```diff
@@ -0,0 +1,22 @@
...
+import { yargsBase } from './command';
+
+const cmd = new CodeGenCommand();
+
+const parser = yargsBase().options({
+    package: { type: 'string', demand: true,  describe: "Js sdk package (stable, beta, etc.)" },
```
henrist: These files (in the top level `scripts` dir) is not linted or formatted. I think the easiest fix is to move it into the codegen package so it will be linted by its setup

scripts/codegen/codegen.ts
henrist: What is confusing? Sure, we can look at it later?

packages/codegen/src/codegen.ts
henrist: It's fine. I just weren't aware of the feature.

packages/stable/src/api/documents/types.gen.ts
henrist: Let's scope it out for now. If we see frequent changes to the ordering and that's annoying we can look into it later.

CONTRIBUTING.md
henrist: I see this content is no longer present. It seems now that you have to change the config to be able to regenerate with a custom path, e.g. for testing some change? I liked the way it was easy previously to generate types for local spec with least effort (e.g. while working on service contracts) - and I'm not sure if it needs to be coupled to a config as it would only be temporary?

packages/codegen/src/commands.ts
henrist: Any reason this is still left around? The code still seems wrong/incomplete around shared types

packages/codegen/src/openapi.ts
henrist: Still unused. Are you mixin this with the next method `operationsInPath` ?

scripts/codegen/codegen.ts
andersfylling: I assume you're handling this now ^^,

packages/playground/.eslintignore
```diff
@@ -0,0 +1 @@
+packages/*/src/exports.gen.ts
```
henrist, line 1: Can you check if you have EditorConfig set up? This should have a trailing newline

henrist:
Note that the PR description is out of date.

package.json
andersfylling: That is a good point, I'm leaving this unresolved to make others aware.

We talked about it, but I'll re-iterate for other reviewers:
All types that are generated have been moved into the `components.schemas` of the open api definition before running code generation. This should allow us to move to different libraries without issues. Ofcourse the way they may generate code (like prefering types for enums, or type comments may differ), but the future will tell if this becomes an issue.

andersfylling:
There was a merge issue, a invalid import statement after annotations moved to stable. Fixed this up, and decided to redo the merge just in case there were any other issues.

Edit: seems tests are "stuck" now (doesn't register as finished even though they are), but this seems to be an issue in other PR's as well so I think it's unrelated to my changes here.

andersfylling:
Talked to @cognite-ornellas about removal of node v12. Out of concern that future changes may break backwards support, we're instead going to only run tests for codegen when the nodejs version is v14 and higher.

cognite-ornellas (approved):

PR #804, opened 2022-05-19
fix: 401 not authorized

cognite-ornellas:


peternic:
Seems to solve our problem 👍🏻

packages/core/src/baseCogniteClient.ts
```diff
@@ -133,10 +133,9 @@ export default class BaseCogniteClient {
 
     this.httpClient.set401ResponseHandler(async (_, retry, reject) => {
       try {
-        const previousToken = this.retrieveTokenValueFromHeader();
+        const previousToken = this.retrieveTokenValueFromHeader(_.headers);
```
henrist: `_` is typically used as "I don't care about the argument". I suggest naming it now that you do.

pedroharbs (approved):

PR #806, opened 2022-05-20
feat(annotations): support for suggest endpoint

peetcremer1:
- Added support for the new `/suggest` endpoint on the Annotations API
- Added a new interface `AnnotationSuggest` for it, which does not contain the `status` field as it is implicitly set to "suggested"
- Have the `AnnotationCreate` interface derive from `AnnotationSuggest``

### DIFF
--- a/packages/playground/src/api/annotations/types.ts
+++ b/packages/playground/src/api/annotations/types.ts
@@ -46,7 +46,7 @@ export interface AnnotationChangeById extends InternalId, AnnotationUpdate {}
 
 export interface AnnotationUpdate {
   update: {
-    annotationType?: SetField<AnnotationPayload>;
+    annotationType?: SetField<AnnotationType>;
     data?: SetField<AnnotationPayload>;
     linkedResourceType?: SinglePatch<LinkedResourceType>;
     linkedResourceId?: SinglePatch<CogniteInternalId>;
--- a/packages/playground/src/api/annotations/types.ts
+++ b/packages/playground/src/api/annotations/types.ts
@@ -24,7 +24,11 @@ export interface AnnotationModel extends AnnotationCreate {
   lastUpdatedTime: Date;
 }
 
-export interface AnnotationCreate {
+export interface AnnotationCreate extends AnnotationSuggest {
+  status: AnnotationStatus;
+}
+
+export interface AnnotationSuggest {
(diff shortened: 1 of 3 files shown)


packages/playground/src/api/annotations/types.ts
```diff
@@ -24,7 +24,11 @@ export interface AnnotationModel extends AnnotationCreate {
   lastUpdatedTime: Date;
 }
 
-export interface AnnotationCreate {
```
miladcognite, line 27: Outside the scope of this PR, but while you're at it :innocent: could you also fix `annotationType` in `AnnotationUpdate`.
Should be `annotationType?: SetField<AnnotationType>;`

miladcognite (approved):
LGTM :+1:

cyriaquebrousse (approved):

packages/playground/src/api/annotations/types.ts
peetcremer1: `AnnotationUpdate` is already like that :-) 
https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/playground/src/api/annotations/types.ts#L46

However, there was another thing that was missing: `creatingAppVersion` in the `AnnotationFilterProps` which I added here: https://github.com/cognitedata/cognite-sdk-js/pull/806/commits/637e8c90c3ff8f70f919e14a25b428dab800d437

miladcognite: hmm...shouldn't `annotationType?: SetField<AnnotationPayload>; ` -> `annotationType?: SetField<AnnotationType>; ` or am I missing something?

peetcremer1: Ah yes, good catch!
Fixed it here: https://github.com/cognitedata/cognite-sdk-js/pull/806/commits/f3c9c3e5d9e00344450a430c02680a665268ae91

peetcremer1:
Also made `annotatedResourceId` required on `AnnotationSuggest` and `AnnotationCreate` and deprecated `annotatedResourceExternalId`, as it is required by the API now in this commit: https://github.com/cognitedata/cognite-sdk-js/pull/806/commits/fd0c34243c186a019ec6c63d78f4ea9d60760965

marvrez (approved):

cognite-ornellas (approved):

PR #812, opened 2022-06-02
test: add retry ability

andersfylling:
It also wouldn't surprise me that running the same tests in parallel may cause some race conditions that affects how flaky the tests are. I specified that the codecov test should run only after the tests for affected packages are successful, and that the test suite should be retried up to 3 times.

github action: https://github.com/nick-fields/retry

I don't see any issues with the action, but please correct me if I'm incorrect. Their codebase is fairly small.


andersfylling:
Hmm, wonder if there's an issue with the github ci tool.

PR #820, opened 2022-06-10
chore: codegen improvements

henrist:
See the individual commits


packages/codegen/package.json
```diff
@@ -1,14 +1,14 @@
 {
   "private": true,
   "name": "@cognite/sdk-codegen",
+  "version": "0.0.0-development",
```
andersfylling, line 4: why did you add version?

packages/codegen/src/snapshot.ts
```diff
@@ -23,11 +23,12 @@ interface ServiceOpenApiOptions extends PathOption {
   filename?: string;
 }
 
+const defaultFilename = 'cognite-openapi-snapshot.json';
```
andersfylling: Could we keep this hidden as this file shouldn't be edited by a user.  Adding `.json` suffix sounds good. I like the `api` => `openapi` change as well.

packages/codegen/src/snapshot.ts
```diff
@@ -68,12 +69,12 @@ export class OpenApiSnapshotManager {
...
       return json as OpenApiDocument;
     } catch (error) {
-      throw new Error(`Unable to download open api contract: ${error}`);
+      throw new Error(`Unable to download OpenAPI contract: ${error}`);
     }
   };
 
   public write = async (openapi: OpenApiDocument): Promise<OpenApiDocument> => {
-    const json = JSON.stringify(openapi);
+    const json = JSON.stringify(openapi, undefined, '  ');
```
andersfylling, line 77: I think some of the snapshots are around several tens of thousands of lines. Could we perhaps just keep this size small by avoiding readable formatting?

I don't think we often need to review the snapshot file. When we do, we could always just make it a cleaner format to help us out.

packages/codegen/package.json
andersfylling: Rather: can you verify that this won't publish/release the codegen package on changes?

package.json
```diff
@@ -25,6 +25,9 @@
   "jest": {
     "testTimeout": 15000
   },
+  "engines": {
+    "node": ">=14"
```
andersfylling, line 29: nice

packages/codegen/package.json
henrist: The commit message should explain. Yarn complains when packages does not contain a version (it's required for a package.json file)

packages/codegen/src/snapshot.ts
henrist: Why do you think it needs to be hidden? There are files like `package-lock.json` or `yarn.lock` which are similar

packages/codegen/src/cli.ts
```diff
@@ -0,0 +1,75 @@
...
+  handler: async (argv) => {
+    await generateTypes({
+      package: argv.package as string,
+    });
+  },
+};
+
+const configureCommand: CommandModule = {
+  command: 'configure',
+  describe: 'Create configuration file to enable type generation for a service',
```
andersfylling: "package or service"

packages/codegen/src/snapshot.ts
henrist: Why do the line count matter?

Currently, when you show the diff, you will get _one line_ with the full diff. At least this makes it readable when you want to see what has changed.

packages/codegen/src/cli.ts
henrist: I suggest we just remove this. See https://github.com/cognitedata/cognite-sdk-js/pull/821

packages/codegen/src/snapshot.ts
andersfylling: `lock` kinda deters people from manually editing it. The reason I made it hidden was to avoid people editing it. Why do you want to make it visible?

henrist: I don't think the fact that its "hidden" will make people not change it. You do modify files like `.gitignore` right?

I think it looks odd to be "hidden" and it doesn't need to be.

I think the "snapshot" part of the name is the best we have that should give a hint that it's not a file to be manually updated.

packages/codegen/package.json
andersfylling: Version is only required if you plan to publish a package. I don't believe that's a goal here.

henrist: Yarn currently complains:

```
$ yarn
warning Missing version in workspace at "/home/henrste/projects/cognitedata/cognite-sdk-js/packages/codegen", ignoring.
```

I don't want to see this warning every time I run `yarn` to install deps. It clutters other potential important warnings, and is just annoying.

andersfylling: Seems like a bug in yarn then, would make more sense to do a PR to yarn instead of adjusting our codebase. I really don't think we should do releases for this package, which I'd argue is more significant than a unnecessary warning.

packages/codegen/src/snapshot.ts
andersfylling: > I think it looks odd to be "hidden" and it doesn't need to be.

It seems to me this is a personal preference, and I don't see why this trumps a decision I made a week or two ago. It's hidden:
 - to deter users from editing it
 - it's owned by a command and is therefore placed in the "background" rather than in the front with all the code/config files

packages/codegen/package.json
henrist: `"private": true` will prevent Lerna/NPM from publishing the package. See https://github.com/lerna/lerna/tree/main/commands/publish#usage and https://docs.npmjs.com/cli/v8/configuring-npm/package-json#private

We have `--no-private` in the `lerna version` command (see `release.yml` workflow), so this package should not be changed by that neither.

packages/codegen/src/snapshot.ts
henrist: I've re-hidden the file.

packages/codegen/package.json
andersfylling: https://github.com/lerna/lerna/issues/3042

Edit: 
looking at https://github.com/cognitedata/cognite-sdk-js/commit/1704e1d4c42de076a6867d7e0a81f25f9ae5f500

It seems there's no issue. So go ahead and keep version in :+1: . I still think the better approach would be to send a patch to yarn as that's the problem, as having a version for a package that is never published doesn't make sense imho.

henrist: I've tested this against a local registry. The version and publish command does not touch the codegen package at all, so this should be fine. I can't comment on that issue, but this might be because we are using more flags and specific variants of the commands.

```
Successfully published:
 - @cognite/sdk-beta@5.1.2
 - @cognite/sdk-core@4.4.1
 - @cognite/sdk-playground@5.3.2
 - @cognite/sdk@7.6.2
lerna success published 4 packages
```

packages/codegen/src/cli.ts
henrist: I've fixed the text

henrist:
Anything you find outstanding for this now @andersfylling?

andersfylling (approved):

andersfylling:
Looks good! Let's get this merged and I'll ping tooling to review it. Hopefully we can this code gen out and see if any problems arise.

PR #821, opened 2022-06-10
chore: drop cli to write config to simplify

henrist:
I suggest we don't have a CLI command to generate the configuration.

For package-specific configs there also needs to be files like `.eslintignore`.

For service-specific configs I think it's easier to keep a template for it, which can also help document what the various fields actually means. I think this demystifies the process, making it more understandable.

### DIFF
--- a/packages/codegen/README.md
+++ b/packages/codegen/README.md
@@ -51,13 +51,30 @@ it block other updates.
 
 ## Configuring type generation for a service
 
-```bash
-yarn codegen configure --package stable --service my-service
+Create a `codegen.json` file in the service directory (e.g.
+`packages/stable/src/api/documents/codegen.json`) using this
+template:
+
+```json
+{
+  "service": "SERVICE-NAME",
+  "filter": {
+    "serviceName": "SERVICE-NAME"
+  },
+  "inlinedSchemas": {
+    "autoNameRequest": true
+  }
+}
 ```
 
-This will generate a `codegen.json` file for the given service
-so types will be created on next run.
+Replace `SERVICE-NAME` with the approriate value, such as `documents`.
 
 The types extracted are based on the paths in the OpenAPI document
 that matches the name of the service. It is currently not possible
 to select individually paths other than this.
+
+Now you can generate types which should include types for your service. E.g:
+
+```bash
+yarn codegen generate-types --package stable
+```
--- a/packages/codegen/src/cli.ts
+++ b/packages/codegen/src/cli.ts
@@ -1,6 +1,5 @@
 // Copyright 2022 Cognite AS
 import yargs, { CommandModule, Options } from 'yargs';
-import { createConfiguration } from './configuration';
 import { generateTypes } from './generate';
 import { updateSnapshot } from './snapshot';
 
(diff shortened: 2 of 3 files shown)


andersfylling:
It would be nice to ensure a new package has a correct setup for code generation. I'd rather have a `enable` or `configure` command that adds required files for the package - especially now since there are multiple.

 > I think this demystifies the process, making it more understandable.

What does it demystify? By default the default config should be adequate, and a user shouldn't have to adjust it. Automating that setup seemed natural. Was there a issue around having a command that setup a basic code generation config for a service?

henrist:
> It would be nice to ensure a new package has a correct setup for code generation. I'd rather have a `enable` or `configure` command that adds required files for the package - especially now since there are multiple.
> 
> > I think this demystifies the process, making it more understandable.
> 
> What does it demystify? By default the default config should be adequate, and a user shouldn't have to adjust it. Automating that setup seemed natural. Was there a issue around having a command that setup a basic code generation config for a service?

Adding new packages isn't something we do every day. I don't even know if we will have any future packages like we have today. And the current implementation lacks e.g. `.eslintignore`. Maybe we also want to move things into package-specific `package.json`. Why should we write code to support all this?

It demystifies by not having the additional layer in between modifying files in the repo (the "magical" layer), but instead being able to explain what needs to be done and how things work. We'll likely need to explain this anyways, so removing the code means less to maintain (= simpler code).

andersfylling:
> Maybe we also want to move things into package-specific package.json. 

This hasn't really been planned.

> the current implementation lacks e.g. .eslintignore.

Can't we just do `await fs.writeFile('packages/*/src/exports.gen.ts\n', path);`, when we create a package config? If it already exists, lets use a append flag.

> Why should we write code to support all this?

Well, it's already written. The reason I have this as a command is to ensure that everything is setup correctly and ready straight way. No user error or confusion. If it's a big issue, I'll say okay to dropping the command for handling a package setup, because you're right it will rarely happen. However, it needs to be documented how to use this on a new package (it doesn't matter if we rarely create new packages, it should be documented).

The task is essentially for a service, without a lightweight command:
 - copy some default minimal config from documentation
 - create a config file
 - paste the content into the config file
 - edit it to correctly use the filter and service name
 - run codegen
 
I think KISS is a nice thing in general, but I can't help but feeling we are optimizing out user experience for very little? There's no attack surface, it's not a complex command. It simply writes a default config with the relevant service values.

If we disagree on what's a nice user experience here, then the only way to move forward would be to walk around and ask people in the office, but that is just taking way too much time. I know I'll be using this command a bit, so I'd like it to stay. 

> It demystifies by not having the additional layer in between modifying files in the repo (the "magical" layer)

I find it hard to agree that this is magic. It's a command that automates the repetitive task of manually making a default config file.

> but instead being able to explain what needs to be done and how things work

We have to do that anyways? Just because they copy the config it doesn't mean they'll understand how the system works.

henrist:
I find us overengineering this. We create config files all the time for various things (tsconfig, package.json, eslint, etc etc) without the need for scripts to do it for us. My hope was that removing the code means we have less things to maintain, also making it simpler to do changes. I'll just close this, as I'm not interested in keep arguing over this. If you see anything useful here (like the doc), feel free to copy it into your PR.

PR #822, opened 2022-06-10
feat(annotations): remove linked resource fields

greenbech:
This PR removes the linked resource fields for AnnotationsV2 since these fields has been removed in the API (the same functionality can be achieved by using the `data` field).

### DIFF
--- a/packages/playground/src/api/annotations/types.ts
+++ b/packages/playground/src/api/annotations/types.ts
@@ -1,17 +1,14 @@
 // Copyright 2022 Cognite AS
 
 import {
-  CogniteExternalId,
   CogniteInternalId,
   FilterQuery,
   IdEither,
   InternalId,
   SetField,
-  SinglePatch,
 } from '@cognite/sdk';
 
 export type AnnotatedResourceType = 'file';
-export type LinkedResourceType = 'file' | 'asset';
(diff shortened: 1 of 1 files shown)


packages/playground/src/api/annotations/types.ts
```diff
@@ -1,17 +1,14 @@
...
-  CogniteExternalId,
   CogniteInternalId,
   FilterQuery,
   IdEither,
   InternalId,
   SetField,
-  SinglePatch,
 } from '@cognite/sdk';
 
 export type AnnotatedResourceType = 'file';
```
cyriaquebrousse, line 11: Looks like this is outdated, see https://github.com/cognitedata/context-api/blob/master/src/api/annotations_v2/schemas.py#L21L24

cyriaquebrousse (approved):
Looks good, just one comment about the annotated resource type list being outdated, but that can be kept for another PR if you want.
The tests did not use the linked resource functionality, so all good there.

packages/playground/src/api/annotations/types.ts
greenbech: Good point. I would prefer if we did this in a separate PR to keep this change as atomic as possible

harish0489-cognite (approved):

PR #825, opened 2022-06-22
feat: add anders as codeowner to codegen

cognite-ornellas:
adding anders as codeowner to codegen


harish0489-cognite (approved):

andersfylling:
Nice! I expect that there will be some moments where I'll need to approve general changes to the stable and playground package too. Can I be added there as well? Then I can help out with approving updates there (whenever new types are generated)

cognite-ornellas:
> Nice! I expect that there will be some moments where I'll need to approve general changes to the stable and playground package too. Can I be added there as well? Then I can help out with approving updates there (whenever new types are generated)

sure @andersfylling , just did it
your help is welcome

PR #832, opened 2022-06-29
fix: propose fix of the getToken already initialized

rhuanbarreto:


vegardok:
This changes the semantics of the SDK constructor, creating a SDK instance have not had this effect before. You should call `authenticate` to ensure that the SDK is authenticated.

PR #833, opened 2022-07-06
fix: added "threedmodel" as possible AnnnotatedResourceType

Savokr:
Initially I was planning to add a PR with all 3d related types but after investigation and talk with @peetcremer1 it was just a change of `AnnotatedResourceType` to include recently added 3D models support. Because in current implementation of Point Cloud annotations we have to use `//@ts-expect-error` on object that is of `AnnotationCreate` type because of missing `threedmodel` type.

P.S. Reveal team would really love to have proper types for `data` and `annotationType` fields. As far as I remember we have some npm packages in Cognite that utilise type generation for typescript, so it should be relatively easy to create something similar but I am not really knowledgable in this topic.


andersfylling:
@Savokr we do have a PR in review for some type generation; https://github.com/cognitedata/cognite-sdk-js/pull/801

polomani:
@Savokr is this released? Doesn't seem to be on docs portal

Screenshot 2022-09-09 at 09 13 14

Savokr:
> @Savokr is this released? Doesn't seem to be on docs portal
> 
> Screenshot 2022-09-09 at 09 13 14

@polomani Yes, it is released in v1 but seems that something (I suspect openAPI spec) is not updated correctly.

PR #839, opened 2022-07-15
chore: update default behavior of yarn

andersfylling:
The `yarn` or `yarn install` command should not mutate the lock file, simply install the dependencies.


harish0489-cognite (approved):
Lgtm

cognite-ornellas (approved):

PR #841, opened 2022-07-18
chore: improved documentation

digaomilleniun:


cognite-ornellas (approved):

polomani:
@cognite-ornellas @digaomilleniun 
please do not merge PRs that break in CI. We have tests for a reason, they check for code sanity.
All of these code snippets are not valid and break CI because `nodes3DApi` field doesn't exist and is not consumable
```
client.nodes3DApi.list(...)
```

Now everyone that contributes to the SDK has a really bad experience understanding why CI is partially broken.

I'd suggest reverting this PR.

PR #855, opened 2022-08-09
feat: add vision api

marvrez:
Jira task: [VIS-982](https://cognitedata.atlassian.net/browse/VIS-982).

Adds JS SDK support for the new extract endpoint in the Vision API.

Example usage

```js
const client = CognitePlaygroundClient(...)
// start the extract job
const job = await client.vision.extract(['PeopleDetection'], [{ fileId: 123 }]);
// will wait for job to complete since we set waitForCompletion=true
const { items } = await client.vision.getExtractJob(job.jobId, true);
// do something with the result
console.log(items[0].predictions.peoplePredictions[0].boundingbox)
```


andersfylling:
I'm noticing you chose to not use the type generation, did you come across an issue?

marvrez:
> I'm noticing you chose to not use the type generation, did you come across an issue?

So, we did actually use type generation. It's just that I ended up simplifying most of the types that it ended up deviating too much from what was originally generated.

There weren't really any issues. Just that the generated types was 1-to-1 with the openapi schemas, which are currently too verbose. IMO, i think we can switch over to fully using type generation once we have simplified our service contract schemas.

andersfylling:
Interesting. One of the main feature here is that if you see issues with the types, you should update open api instead of the code. This does eliminate concerns around naming conflicts, keeping everything up to date (unless some unexpected state happen), etc.

Is it possible to fix the open api spec and then come back to this PR (as that is the expected future workflow)? That way we get you into code generation right away. If you meet any issues in code generation, you can always ping me and I'll fix it up - the current phase is to see what we might have missed and what we need to implement to accommodate team requirements.

marvrez:
Thanks for the suggestion @andersfylling. I added back the codegen stuff in https://github.com/cognitedata/cognite-sdk-js/pull/855/commits/0379ff8110154666c2900fcba2caa9f00c12f3c1 after refactoring the service contracts in https://github.com/cognitedata/service-contracts/pull/1791. The suggested workflow was surprisingly pain-free and easy :)  

There was a bug that I had to fix though, which was that the type names for the exports in `exports.gen.ts` wasn't consistent with the generated type names in `types.gen.ts` (for example, the generated type `AnnotationsCogniteAnnotationTypesImagesAssetLink` in types.gen.ts would result in export of `Annotations.cognite__annotation_types__images__AssetLink` in exports.gen.ts).

I added a fix in https://github.com/cognitedata/cognite-sdk-js/pull/855/commits/e7b588489aeee7290ec69885718f0d90e2f59e8d which uses the output from `generateApi` to figure out which modules were generated and return back the name of those instead. This should make it so that we will always export the exact same type names as the ones generated in `types.gen.ts`. You may want to take a look at that commit.

miladcognite (commented):
Great job! LGTM, but I let @andersfylling approve this one so he also get a chance to review the latest changes.

miladcognite (approved):

cognite-ornellas (approved):

andersfylling:
@marvrez That's a nice adjustment!

PR #860, opened 2022-08-18
docs: improved example docs

digaomilleniun:


packages/stable/src/api/annotations/annotationsApi.ts
```diff
@@ -45,13 +47,21 @@ export class AnnotationsAPI extends BaseResourceAPI<AnnotationModel> {
    * [Retrieve](https://docs.cognite.com/api/playground/#operation/annotationsByids)
    * or [Get an](https://docs.cognite.com/api/playground/#operation/annotationsGet)
    * annotation
+   *
+   * const response = await client.annotationsApi.retrieve(createdAnnotationIds);
```
polomani:
```suggestion
   * const response = await client.annotations.retrieve(createdAnnotationIds);
```
all of these are invalid..
have you tried running those in some test project?

polomani: we cannot have code snippets that are syntactically incorrect

polomani:
have you tried
```
yarn lint --fix
```
?

packages/stable/src/api/annotations/annotationsApi.ts
```diff
@@ -36,6 +36,8 @@ export class AnnotationsAPI extends BaseResourceAPI<AnnotationModel> {
 
   /**
    * [Suggest annotations](https://docs.cognite.com/api/playground/#operation/annotationsSuggest)
+   * 
+   * const created = await client.annotations.create(annotations);
```
polomani: what's the purpose of this example if we don't show the actual input object?
we run TS compiler here, so it will fail as the variable don't exist

digaomilleniun:
I will open another one with a new branch

PR #862, opened 2022-08-23
chore(codegen): use abstract syntax tree in post processing

andersfylling:
## Description

> Note that this affects the code format (for generated files), but given it's generated code, I don't see this as a significant problem.

After types have been generated, they don't utilize types from the core package. CursorAndAsyncIterator is one case, which leaves us without auto pagination. 

By introducing ast transformers we can compensate for these lacking issues within any openapi=>typescript generator/library we chose to utilize.

I've added a simple transformer that looks for any declaration with `items` and `nextCursor`, and replaces them with a CursorAndAsyncIterator heritage clause.

## Example

Before:
```ts
/**
 * Response for allUniqueValues aggregate.
 * @example {"items":[{"count":4,"values":["hello"]},{"count":33,"values":["world"]}]}
 */
export interface DocumentsAggregateAllUniqueValuesResponse {
  items: DocumentsAggregateAllUniqueValuesItem[];

  /** The cursor to get the next page of results (if available). */
  nextCursor?: string;
}
```

After:
```ts
import { CursorAndAsyncIterator } from '@cognite/sdk-core';

/**
 * Response for allUniqueValues aggregate.
 * @example {"items":[{"count":4,"values":["hello"]},{"count":33,"values":["world"]}]}
 */
export interface DocumentsAggregateAllUniqueValuesResponse extends CursorAndAsyncIterator<DocumentsAggregateAllUniqueValuesItem> {}
```

###  ~Unrelated issue~

> Reverted this part, after updating the CursorAndAsyncIterator match rule

~Additionally (_why all the github actions are failing_): I'm having a issue with the search endpoint for the documents service, I added a github comment for the specific line. If anyone has suggestions for how to fix it, please have a look.~


packages/stable/src/api/documents/documentsApi.ts
```diff
@@ -60,32 +60,18 @@ export class DocumentsAPI extends BaseResourceAPI<Document> {
    * });
    * ```
    */
-  public search = (
-    query: DocumentSearchRequest
-  ): Promise<DocumentSearchResponse> => {
-    return this.searchDocuments<DocumentSearchResponse>(query);
+  public search = (request: DocumentSearchRequest): DocumentSearchResponse => {
+    return this.searchEndpoint(request);
```
andersfylling: Currently doesn't work as expected, any feedback on how I can go about resolving this?

Seems the search endpoint is using the generic argument for BaseResourceAPI, while I require something else.

andersfylling:
Perhaps CursorAndAsyncIterator should be used when only `items` and `nextCursor` are present.

packages/stable/src/api/documents/documentsApi.ts
andersfylling: reverted

andersfylling:
I had to add another transformer which simply sorts declarations by name, as well as their properties. I'm surprised by the large diff, but it seems that the old sorting logic was lacking, and has now been improved.

andersfylling:
closed in favour of:
 - https://github.com/cognitedata/cognite-sdk-js/pull/907
 - https://github.com/cognitedata/cognite-sdk-js/pull/908

PR #873, opened 2022-09-09
feat: support beta features in vision extract

marvrez:


andersfylling:
beta features should not be in the stable, but in the beta package.

marvrez:
> beta features should not be in the stable, but in the beta package.

It's a bit different for the Vision Extract API since it is deployed in V1, but some features are available when using the beta flag. Hence the `console.warn`.

packages/stable/src/__tests__/api/visionApi.int.spec.ts
```diff
@@ -35,6 +40,22 @@ describe('Vision API', () => {
...
+  test('extract using beta feature', async () => {
+    // Check that the beta flag is correctly used.
+    // Unfortunately, 'cdf-version' is not listed in
+    // access-control-allow-headers, which means we cannot explicitly check for
+    // that the 'cdf-version' entry is set to 'beta'. However, what we instead
+    // can do is to check that the API does not return 400 when a beta feature
+    // is sent.
+    const metadata = client.getMetadata(extractBetaJob);
+    expect(metadata).toBeDefined();
+    expect(metadata?.status).toEqual(200);
```
miladcognite, line 54: Consider also checking that `console.warn` is called

miladcognite (approved):
One minor suggestion, otherwise LGTM :rocket:

packages/stable/src/__tests__/api/visionApi.int.spec.ts
marvrez: https://github.com/cognitedata/cognite-sdk-js/pull/873/commits/7ad169c5717d1f62089433328223864de8f41ad6

PR #876, opened 2022-09-14
feat: allow overriding retry validator at request level

cemreyavuz:
[DX-907](https://cognitedata.atlassian.net/browse/DX-907?atlOrigin=eyJpIjoiNzU0MmY3ZTZiZjNkNDYyY2E3NTZkMGIzYmUyNzExZGIiLCJwIjoiamlyYS1zbGFjay1pbnQifQ)

In brief, the motivation is allowing to override retry behavior at the request level. Slack thread can be checked for more background.

In the PR, I extended the http request options with a field called `retryValidator`. It can be either a function or a boolean with the value `false`. If it is `false`, we don't retry at all. If it is a function, we call it to decide whether retry or not. Else, we use the default retry validator.

There are a few ways we can go for the solution. I went with the generic solution to include extra options for requests inside the `BasicHttpClient`. However, I wouldn't say I'm too happy with it. We need to pass the extra props for retry validator basically everywhere.

Instead of extending `BasicCogniteClient` to add retry functionality, we can implement this inside that class with an option to disable globally (which is actually the idea with `RetryableHttpClient` in the end as well). This might be one alternative.

### DIFF
--- a/packages/core/src/httpClient/retryValidator.ts
+++ b/packages/core/src/httpClient/retryValidator.ts
@@ -16,7 +16,7 @@ export type RetryValidator = (
 /** @hidden */
 export type EndpointList = { [key in HttpMethod]?: string[] };
 
-const DEFAULT_MAX_RETRY_ATTEMPTS = 5;
+export const DEFAULT_MAX_RETRY_ATTEMPTS = 5;
 
 export const createRetryValidator = (
   endpointsToRetry: EndpointList,
--- a/packages/core/src/httpClient/retryValidator.ts
+++ b/packages/core/src/httpClient/retryValidator.ts
@@ -16,11 +16,11 @@ export type RetryValidator = (
 /** @hidden */
 export type EndpointList = { [key in HttpMethod]?: string[] };
 
-export const DEFAULT_MAX_RETRY_ATTEMPTS = 5;
+export const MAX_RETRY_ATTEMPTS = 5;
 
 export const createRetryValidator = (
   endpointsToRetry: EndpointList,
-  maxRetries: number = DEFAULT_MAX_RETRY_ATTEMPTS
+  maxRetries: number = MAX_RETRY_ATTEMPTS
 ): RetryValidator => {
   const universalRetryValidator = createUniversalRetryValidator(maxRetries);
   return (
@@ -46,7 +46,7 @@ export const createRetryValidator = (
 };
 
 export const createUniversalRetryValidator =
-  (maxRetries: number = DEFAULT_MAX_RETRY_ATTEMPTS): RetryValidator =>
+  (maxRetries: number = MAX_RETRY_ATTEMPTS): RetryValidator =>
   (request, response, retryCount) => {
     if (retryCount >= maxRetries) {
       return false;
--- a/packages/core/src/httpClient/retryValidator.ts
+++ b/packages/core/src/httpClient/retryValidator.ts
@@ -3,11 +3,12 @@
 import inRange from 'lodash/inRange';
 import some from 'lodash/some';
 
-import { HttpMethod, HttpRequest, HttpResponse } from './basicHttpClient';
+import { HttpMethod, HttpResponse } from './basicHttpClient';
+import { RetryableHttpRequest } from './retryableHttpClient';
 
 /** @hidden */
 export type RetryValidator = (
-  request: HttpRequest,
+  request: RetryableHttpRequest,
   response: HttpResponse<any>,
   retryCount: number
 ) => boolean;
@@ -23,7 +24,7 @@ export const createRetryValidator = (
 ): RetryValidator => {
   const universalRetryValidator = createUniversalRetryValidator(maxRetries);
   return (
-    request: HttpRequest,
+    request: RetryableHttpRequest,
     response: HttpResponse<any>,
     retryCount: number
   ) => {
--- a/packages/core/src/httpClient/retryableHttpClient.ts
+++ b/packages/core/src/httpClient/retryableHttpClient.ts
@@ -63,9 +63,9 @@ export class RetryableHttpClient extends BasicHttpClient {
 
   protected async postRequest<T>(
     response: HttpResponse<T>,
-    _: RetryableHttpRequest // eslint-disable-line
+    request: RetryableHttpRequest
   ): Promise<HttpResponse<T>> {
-    return super.postRequest<T>(response, _);
+    return super.postRequest<T>(response, request);
   }
 
   protected async rawRequest<ResponseType>(
--- a/packages/core/src/httpClient/retryableHttpClient.ts
+++ b/packages/core/src/httpClient/retryableHttpClient.ts
@@ -1,6 +1,11 @@
 // Copyright 2020 Cognite AS
 import { sleepPromise } from '../utils';
-import { BasicHttpClient, HttpRequest, HttpResponse } from './basicHttpClient';
+import {
+  BasicHttpClient,
+  HttpRequest,
+  HttpRequestOptions,
+  HttpResponse,
+} from './basicHttpClient';
 import { RetryValidator } from './retryValidator';
 
 export class RetryableHttpClient extends BasicHttpClient {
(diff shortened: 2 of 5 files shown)


packages/core/src/httpClient/basicHttpClient.ts
```diff
@@ -253,12 +270,12 @@ export function headersWithDefaultField(
   return headers;
 }
 
-export interface HttpRequest extends HttpRequestOptions {
+export type HttpRequest<ExtraOptions = any> = {
   path: string;
   method: HttpMethod;
-}
+} & HttpRequestOptions<ExtraOptions>;
```
cemreyavuz: An interface cannot extend another interface with unknown members - so now `HttpRequest` is a type.

cemreyavuz:
Integration test probably fails because of the actual api

packages/core/src/httpClient/retryableHttpClient.ts
```diff
@@ -16,13 +21,21 @@ export class RetryableHttpClient extends BasicHttpClient {
...
-      const shouldRetry = this.retryValidator(request, response, retryCount);
+
+      const retryValidator =
+        typeof request.retryValidator === 'function'
+          ? request.retryValidator
+          : this.retryValidator;
+
+      const shouldRetry =
+        request.retryValidator !== false &&
+        retryValidator(request, response, retryCount);
```
cemreyavuz: This is basically the whole change, rest is just type updates

polomani:
@cemreyavuz what's the use case? what do you actually want to achieve?

cemreyavuz:
Forgot to add the slack thread link: https://cognitedata.slack.com/archives/CG10VQPFX/p1663074669728439. In the most basic sense, I want to override the retry behavior for some requests. I want to stop retrying for some, I want to change the retry behavior for some other (like, retry 3 times instead of 5, or change the back-off mechanism).

BugGambit:
What about adding an optional `retryValidator?: RetryValidator` field here: https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/baseCogniteClient.ts#L30-L50
And modify the `getRetryValidator` method [here](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/baseCogniteClient.ts#L328-L330) to check if the user passed a custom retryValidator?
Wouldn't that solve the problem without propagating changes down to `BasicHttpClient`?

cemreyavuz:
> What about adding an optional retryValidator?: RetryValidator field here: https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/baseCogniteClient.ts#L30-L50
And modify the getRetryValidator method [here](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/core/src/baseCogniteClient.ts#L328-L330) to check if the user passed a custom retryValidator?
Wouldn't that solve the problem without propagating changes down to BasicHttpClient?

I think we should also add that option "in addition" to this one - that was actually going to be my next PR.

I agree that it is a bit complex to propagate these changes down to `BasicHttpClient` - but I think we should still give the option to stop retrying per request. We don't have to override the function itself, so it can be a boolean. But even that wouldn't simplify this code necessarily.

You might ask why overriding the global retry validator is not enough. The fusion specific reason is that there are specific needs for each subapp for retries. The main motivation for doing this at request level is not retrying rather "unimportant" requests to unblock other "important" requests. One example from transformations sub app: we make tons of requests to raw api for fetching the tables in databases for tooltips on sidebar. When the number of requests gets bigger, we start getting 429s for raw api requests. However, we actually make some other requests to raw api in the same page that is more critical for UI to function. So when we handle these at request level, either we can stop retrying lower priority requests - or we can higher the back-off period to let higher priority requests go through first.

Also another problem with having one retry validator on global level is that we can only manage that in sdk-singleton. Subapps wouldn't be able to change the behavior at all. So, subapp developers would go into sdk-singleton and update the validator as it fits their use cases. And, there is still prioritization problem specific to each subapp. So, overall I still think we should have an option to either stop or override retrying at request level.

packages/core/src/httpClient/basicHttpClient.ts
BugGambit: Can we remove the `any` part? Then it should be extendable.
`ExtraOptions = any` is scary. We don't know how it looks like (not even if it is an object). We should control what are the extra options someone can pass in here.

BugGambit (changes requested):
I left a small comment.

I get the point that a global retry validator is not sufficient for every use case.
But I still believe we should not modify `BasicHttpClient`. That is just an abstraction for Axios/fetch etc., and shouldn't care about retries. If we do that, we mix the abstraction levels.
I suggest to modify `RetryableHttpClient` where we override the http-methods (`get`, `post`, `put`, `delete`, `patch`) from `BasicHttpClient` providing an additional option to have a custom retry validator.

What do you think?

packages/core/src/httpClient/basicHttpClient.ts
cemreyavuz: I agree the comment, but removed that part directly so no longer applies. Instead of making `HttpRequest` generic, extending it in `retryableHttpClient.ts` file also works

cemreyavuz:
> I left a small comment.
> 
> I get the point that a global retry validator is not sufficient for every use case.
> But I still believe we should not modify BasicHttpClient. That is just an abstraction for Axios/fetch etc., and shouldn't care about retries. If we do that, we mix the abstraction levels.
> I suggest to modify RetryableHttpClient where we override the http-methods (get, post, put, delete, patch) from BasicHttpClient providing an additional option to have a custom retry validator.
> 
> What do you think?

I agree on not modifying `BasicHttpClient`, that's why I went the way of extending type with `ExtraOptions` for requests . But doing all of these changes inside `RetryableHttpClient` also works - actually it's even better imo. We have more control over it this way and it's less error prone for the future.

~(I'll also add a few tests for this new functionality - but wanted to get the review for the actual code in the meanwhile)~

cemreyavuz:
@BugGambit we can also add a test to check if default max retry attemps number is used - but the same problem applies here as well: https://github.com/cognitedata/cognite-sdk-js/pull/878#discussion_r977314903. I'm open for suggestions if there is a way to make it work

packages/core/src/httpClient/retryableHttpClient.ts
```diff
@@ -15,14 +20,71 @@ export class RetryableHttpClient extends BasicHttpClient {
...
       const response = await super.rawRequest<ResponseType>(request);
-      const shouldRetry = this.retryValidator(request, response, retryCount);
+
+      const retryValidator =
+        typeof request.retryValidator === 'function'
+          ? request.retryValidator
+          : this.retryValidator;
+
+      const shouldRetry =
+        retryCount < DEFAULT_MAX_RETRY_ATTEMPTS &&
```
BugGambit: Do we want to make "make number of retries" a decision up to the validators?
We are passing `retryCount` to the retryValidators.
If yes, we don't need to export `DEFAULT_MAX_RETRY_ATTEMPTS`

packages/core/src/httpClient/retryableHttpClient.ts
```diff
@@ -31,4 +93,18 @@ export class RetryableHttpClient extends BasicHttpClient {
...
 }
+
+export type RetryableHttpRequest = HttpRequest &
+  HttpRequestRetryValidatorOptions;
+
+export type RetryableHttpRequestOptions = HttpRequestOptions &
+  HttpRequestRetryValidatorOptions;
+
+type HttpRequestRetryValidatorOptions = {
+  retryValidator?: false | RetryValidator;
```
BugGambit, line 109: So the behaviour is as follows:
- `null`/`undefined` -> default retrier
- `false` -> no retries at all
- `function` -> custom retrier
?

BugGambit (commented):
Nice. I liked this approach better. I left a few comments.

packages/core/src/httpClient/retryableHttpClient.ts
cemreyavuz: Ref to Henry's comment in slack:

> Henry Henriques
>   [9 days ago](https://cognitedata.slack.com/archives/CG10VQPFX/p1663145234640469?thread_ts=1663074669.728439&cid=CG10VQPFX)
> I don't think we should allow for custom number of tries (fear of people setting it to silly high values). But I think it's reasonable to have the option of disabling it completely.

I'm not strongly opinionated on forcing a max value - if someone wants to retry in 10 times instead of 5 times - they still can do that. So we can just remove that part if you also agree @BugGambit

packages/core/src/httpClient/retryableHttpClient.ts
cemreyavuz: Yes, the motivation was flexibility - but of course that comes with a small cost on cognitive complexity.

We can even go one step further and add the option of numbers for retry counts in the future as well - sth like this: https://tanstack.com/query/v4/docs/guides/query-retries. I didn't want to have it as I was forcing a max retry count - but if we won't have it, we can have that option as well. In that case the behaviour would be:

- undefined-> default retry validator
- false -> no retry
- function -> custom retry validator
- number -> just check retry count and retry if it doesn't exceed the argument

packages/core/src/httpClient/retryableHttpClient.ts
BugGambit: Ok, I hadn't seen @hhenriques's comment. Let's keep the max number of retries.

packages/core/src/httpClient/retryValidator.ts
```diff
@@ -3,27 +3,28 @@
...
+  request: RetryableHttpRequest,
   response: HttpResponse<any>,
   retryCount: number
 ) => boolean;
 
 /** @hidden */
 export type EndpointList = { [key in HttpMethod]?: string[] };
 
-const DEFAULT_MAX_RETRY_ATTEMPTS = 5;
+export const DEFAULT_MAX_RETRY_ATTEMPTS = 5;
```
BugGambit: Should we rename it to `MAX_RETRY_ATTEMPTS` as it can be overridden?

packages/core/src/httpClient/retryableHttpClient.ts
```diff
@@ -15,14 +20,71 @@ export class RetryableHttpClient extends BasicHttpClient {
...
+
+  protected async preRequest(
+    request: RetryableHttpRequest
+  ): Promise<RetryableHttpRequest> {
+    return super.preRequest(request);
+  }
+
+  protected async postRequest<T>(
+    response: HttpResponse<T>,
+    _: RetryableHttpRequest // eslint-disable-line
```
BugGambit: What's the problem here? And why not 
```suggestion
    request: RetryableHttpRequest
```

packages/core/src/httpClient/retryableHttpClient.ts
```diff
@@ -15,14 +20,71 @@ export class RetryableHttpClient extends BasicHttpClient {
...
+    request: RetryableHttpRequest
   ): Promise<HttpResponse<ResponseType>> {
     let retryCount = 0;
     // eslint-disable-next-line no-constant-condition
     while (true) {
       const response = await super.rawRequest<ResponseType>(request);
-      const shouldRetry = this.retryValidator(request, response, retryCount);
+
+      const retryValidator =
+        typeof request.retryValidator === 'function'
```
BugGambit: Use https://lodash.com/docs/4.17.15#isFunction ?

packages/core/src/httpClient/retryValidator.ts
cemreyavuz: Yes, makes more sense 👍🏻

packages/core/src/httpClient/retryableHttpClient.ts
cemreyavuz: I copied it from `BasicHttpClient`, but no reason to keep it that way 👍🏻

cemreyavuz:
~@BugGambit should I also bump the version, or does ci/cd do it automatically?~ (i think it does, so I'm not changing)

BugGambit (approved):
LGTM 🚀

BugGambit:
> ~@BugGambit should I also bump the version, or does ci/cd do it automatically?~ (i think it does, so I'm not changing)

Correct. It will bump it automatically based on the commit messages.

PR #877, opened 2022-09-20
Feat/units-conversion-sample

cognite-ornellas:


cognite-ornellas:
this PR was just about testing something

PR #878, opened 2022-09-21
feat: update 401 handler logic

cemreyavuz:


packages/core/src/__tests__/cogniteClient.unit.spec.ts
```diff
@@ -131,6 +131,70 @@ describe('CogniteClient', () => {
...
       });
+
+      test('getToken should be called once for parallel 401s', async () => {
+        nock(mockBaseUrl).get('/test').thrice().reply(401);
+        nock(mockBaseUrl).get('/test').thrice().reply(200);
+
+        const mockGetToken = jest.fn(
+          () =>
+            new Promise<string>((resolve) => {
+              setTimeout(() => {
```
BugGambit: Just a heads up: https://jestjs.io/docs/timer-mocks

cemreyavuz: I didn't know this existed, so TIL thanks 🙏🏻

cemreyavuz: Tried a bit, but couldn't make it work. Seems like it doesn't work well with promises: only way I made it work was calling the `jest.useFakeTimers()` etc. inside the promise, but then the code started to act like synchronous - and I cannot test what I wanted. Probably it is something around event loops (depending on how jest implemented it) - but feels like it's gonna be too much effort to make it work properly. And even I make it work, a future change in 401 handler might break it again.

If your concern is the timeout duration, we can just lower it to 50-100ms. Or, if you have a suggestion how we can use fake timers in this test, that also works.

cemreyavuz: Lowered it to 100ms - we can lower it even further but I think it is acceptable. What do you think?

packages/core/src/httpClient/basicHttpClient.ts
```diff
@@ -215,7 +215,7 @@ export class BasicHttpClient {
   protected async request<ResponseType>(request: HttpRequest) {
     const mutatedRequest = await this.preRequest(request);
     const rawResponse = await this.rawRequest<ResponseType>(mutatedRequest);
-    return this.postRequest(rawResponse, request);
+    return this.postRequest(rawResponse, mutatedRequest);
```
cemreyavuz, line 218: One of the two main goals for this PR is comparing new token with the one we used in request. All of the previous implementations were around it, but they were either using response headers (response doesn't have to have authorization header) or using a global variable (race conditions).

cemreyavuz: I don't know why we weren't passing `mutatedRequest` to `this.postRequest` before but to me, it looks fine

packages/core/src/baseCogniteClient.ts
```diff
@@ -75,19 +75,21 @@ export default class BaseCogniteClient {
...
 
   private readonly credentialsAuth?: CredentialsAuth;
 
+  /**
+   * To prevent calling `getToken` method multiple times in parallel, we set
+   * the promise that `getToken` returns to a variable and check its existence
+   * on function calls for `authenticate`.
+   */
+  private tokenPromise?: Promise<string | undefined>;
+  private isTokenPromiseFulfilled?: boolean;
```
cemreyavuz, line 91: Either we need to create our own promise class with statuses (I attempted sth like that for sdk-singleton a while back: https://github.com/cognitedata/cdf-hub-sdk-singleton/pull/74/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R6-R11) or we need to keep a boolean for that status. Right now we don't care about anything except `fulfilled` status - so this should be sufficient.

packages/core/src/baseCogniteClient.ts
```diff
@@ -168,7 +170,12 @@ export default class BaseCogniteClient {
   private authenticateGetToken: () => Promise<string | undefined> =
     async () => {
       try {
-        const token = await this.getToken();
+        if (!this.tokenPromise || this.isTokenPromiseFulfilled) {
+          this.isTokenPromiseFulfilled = false;
+          this.tokenPromise = this.getToken();
+        }
+        const token = await this.tokenPromise;
+        this.isTokenPromiseFulfilled = true;
```
cemreyavuz, line 178: Parallel 401s causes a new token request each. If there is a promise pending, we can just use that instead.

packages/core/src/baseCogniteClient.ts
```diff
@@ -203,11 +210,16 @@ export default class BaseCogniteClient {
...
       );
 
-    httpClient.set401ResponseHandler(async (_, retry, reject) => {
+    httpClient.set401ResponseHandler(async (_, request, retry, reject) => {
       try {
-        const newToken = await this.authenticate();
-        if (newToken && newToken !== this.previousToken) {
-          this.previousToken = newToken;
+        const requestToken = this.retrieveTokenValueFromHeader(request.headers);
+        const currentToken = await this.tokenPromise;
```
cemreyavuz, line 216: We might have fetched a new token before the request is completed - checking if the current token in the sdk is different from the one in original request

cemreyavuz:
Similar to https://github.com/cognitedata/cognite-sdk-js/pull/876, integration test fails for some reason.

packages/core/src/__tests__/cogniteClient.unit.spec.ts
BugGambit: Yes, that's fine. At least you are now aware of the feature 🙂

packages/core/src/__tests__/cogniteClient.unit.spec.ts
```diff
@@ -131,6 +131,102 @@ describe('CogniteClient', () => {
...
+                resolve('test-token');
+              }, 100);
+            })
```
BugGambit:
```suggestion
          async () => {
            await sleepPromise(100);
            return 'test-token';
          }
```

packages/core/src/__tests__/cogniteClient.unit.spec.ts
```diff
@@ -131,6 +131,102 @@ describe('CogniteClient', () => {
...
+        nock(mockBaseUrl).get('/test').twice().reply(200);
+        nock(mockBaseUrl).get('/test-with-delay').reply(200);
+
+        const mockGetToken = jest.fn(
+          () =>
+            new Promise<string>((resolve) => {
+              setTimeout(() => {
+                resolve('test-token');
+              }, 100);
+            })
```
BugGambit:
```suggestion
          async () => {
            await sleepPromise(100);
            return 'test-token';
          }
```

packages/core/src/__tests__/cogniteClient.unit.spec.ts
```diff
@@ -131,6 +131,102 @@ describe('CogniteClient', () => {
...
+        nock(mockBaseUrl).get('/test').thrice().reply(401);
+        nock(mockBaseUrl).get('/test').thrice().reply(200);
+        nock(mockBaseUrl).get('/test').reply(401);
+        nock(mockBaseUrl).get('/test').reply(200);
+
+        const mockGetToken = jest.fn(
+          () =>
+            new Promise<string>((resolve) => {
+              setTimeout(() => {
+                resolve(`test-token-${Math.floor(Math.random() * 1000)}`);
```
BugGambit: Can we make this part deterministic? Ex. `test-token-${tokenCount++}`.
And utilize the `sleepPromise` function.

BugGambit (changes requested):
Some minor comments

BugGambit:
FYI, the failing test was fixed here: https://github.com/cognitedata/cognite-sdk-js/pull/893/commits/fbc77ed33839a6e4d1d6d61026f56f148f4af70d
So updating this branch will fix it.

BugGambit (approved):
LGTM

PR #893, opened 2022-09-22
fix: ensure response ordering matches request when chunking a request

BugGambit:
See this [Slack thread](https://cognitedata.slack.com/archives/C6KNJCEEA/p1663773050949699) for context.

We had a bug when performing a request to the API and the SDK decides to chunk the request into smaller requests. The problem is that we were not guaranteed that the response order matched the input order.

### DIFF
--- a/packages/stable/src/__tests__/api/events.int.spec.ts
+++ b/packages/stable/src/__tests__/api/events.int.spec.ts
@@ -129,15 +129,16 @@ describe('Events integration test', () => {
     test('descending', async () => {
       await client.events.list({ sort: { endTime: SortOrder.DESC } });
     });
-    test('multiple props not supported', async () => {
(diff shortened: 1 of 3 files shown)


packages/stable/src/__tests__/api/events.int.spec.ts
```diff
@@ -129,15 +129,16 @@ describe('Events integration test', () => {
     test('descending', async () => {
       await client.events.list({ sort: { endTime: SortOrder.DESC } });
     });
-    test('multiple props not supported', async () => {
+    test('multiple props supported', async () => {
```
BugGambit, line 132: The events API added support for multiple fields. So the test is not valid anymore.

polomani (approved):
LGTM 🎉

PR #896, opened 2022-09-26
fix: override auth header while retrying

cemreyavuz:
Latest change to this line caused a regression: https://github.com/cognitedata/cognite-sdk-js/blob/9797d780e1e20705e7ffaa6767e0e271e6360239/packages/core/src/httpClient/basicHttpClient.ts#L218

Updated tokens are not used on retries for 401s now.

Because now we are passing `mutatedRequest` to `postRequest`, not all the default headers are applied properly. For us, the important one is the authorization header. Here we keep using the old authorization header on retries of because of this ordering: https://github.com/cognitedata/cognite-sdk-js/blob/9797d780e1e20705e7ffaa6767e0e271e6360239/packages/core/src/httpClient/basicHttpClient.ts#L172-L177

I'm not sure if we should change this to other way - then we wouldn't be able to override default headers. So I went with adding another parameter to `postRequest` for keeping both the original request and the mutated version. Open to suggestions to clean it up.

### DIFF
--- a/packages/core/src/httpClient/basicHttpClient.ts
+++ b/packages/core/src/httpClient/basicHttpClient.ts
@@ -160,7 +160,8 @@ export class BasicHttpClient {
 
   protected async postRequest<T>(
     response: HttpResponse<T>,
-    _: HttpRequest // eslint-disable-line
+    _: HttpRequest, // eslint-disable-line
+    __: HttpRequest // eslint-disable-line
   ): Promise<HttpResponse<T>> {
     const requestIsOk = BasicHttpClient.validateStatusCode(response.status);
     if (!requestIsOk) {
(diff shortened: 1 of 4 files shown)


packages/core/src/httpClient/basicHttpClient.ts
```diff
@@ -215,7 +216,7 @@ export class BasicHttpClient {
   protected async request<ResponseType>(request: HttpRequest) {
     const mutatedRequest = await this.preRequest(request);
     const rawResponse = await this.rawRequest<ResponseType>(mutatedRequest);
-    return this.postRequest(rawResponse, mutatedRequest);
+    return this.postRequest(rawResponse, request, mutatedRequest);
```
cemreyavuz, line 219: We need a way to differentiate between default headers and original headers in the request. Filtering out the default headers will not work because they can be overridden in the original request.

We need the mutated request to compare the authorization header used in the request, so maybe we can pass that instead of the whole request.

BugGambit (approved):
LGTM. As far as I can see, the SDK interface for the user would not change due to this PR.

BugGambit:
@cemreyavuz Did the PR fix the problem?

cemreyavuz:
@BugGambit yes the problem is fixed with this version - also I have used it for the whole day and i didn't see any problem: 401s, retrys, etc. Everything looked fine 👍🏻

PR #897, opened 2022-10-03
fix: move `@types/geojson` to dependencies

samir-ajdarpasic:
The `geojson` types are missing when we install the sdk package in other places. This should fix it.

### DIFF
--- a/packages/stable/package.json
+++ b/packages/stable/package.json
@@ -25,9 +25,7 @@
   "dependencies": {
     "@cognite/sdk-core": "^4.8.1",
     "geojson": "^0.5.0",
-    "lodash": "^4.17.11"
-  },
-  "devDependencies": {
+    "lodash": "^4.17.11",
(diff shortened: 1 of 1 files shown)


andersfylling (approved):
I see this is a dependency in the stable package: [packages/stable/src/api/geospatial/types.ts](https://github.com/cognitedata/cognite-sdk-js/blob/4e207aa41eab84a40968b9aed826ac29da7700d1/packages/stable/src/api/geospatial/types.ts)

samir-ajdarpasic:
> I see this is a dependency in the stable package: [packages/stable/src/api/geospatial/types.ts](https://github.com/cognitedata/cognite-sdk-js/blob/4e207aa41eab84a40968b9aed826ac29da7700d1/packages/stable/src/api/geospatial/types.ts)

@andersfylling Yes, but it was a "devDependency" in package.json

BugGambit (approved):

PR #899, opened 2022-10-18
feat: geospatial compute endpoint

BenChand:
* Partially stabilize the new geospatial compute endpoint

### DIFF
--- a/packages/stable/src/api/geospatial/computeAPI.ts
+++ b/packages/stable/src/api/geospatial/computeAPI.ts
@@ -1,7 +1,7 @@
 import { BaseResourceAPI } from '@cognite/sdk-core';
 import {
-  GeospatialComputeRequest,
-  GeospatialComputeResponse
+  GeospatialJsonComputeOutput,
+  GeospatialComputeResponse,
 } from './types';
 
 export class ComputeAPI extends BaseResourceAPI<GeospatialComputeResponse> {
@@ -10,12 +10,12 @@ export class ComputeAPI extends BaseResourceAPI<GeospatialComputeResponse> {
    *
    */
   public compute = async (
-    request: GeospatialComputeRequest
+    request: GeospatialJsonComputeOutput
   ): Promise<GeospatialComputeResponse> => {
     const path = this.url();
     const response = await this.post<GeospatialComputeResponse>(path, {
-      data: request
+      data: request,
     });
-    return response.data
+    return response.data;
   };
 }
--- a/packages/stable/src/api/geospatial/computeAPI.ts
+++ b/packages/stable/src/api/geospatial/computeAPI.ts
@@ -1,0 +1,21 @@
+import { BaseResourceAPI } from '@cognite/sdk-core';
+import {
+  GeospatialComputeRequest,
+  GeospatialComputeResponse
+} from './types';
+
+export class ComputeAPI extends BaseResourceAPI<GeospatialComputeResponse> {
+  /**
+   * [Compute custom json output structures or well known binary format responses based on calculation or selection of feature properties or direct values given in the request.](https://docs.cognite.com/api/v1/#operation/compute)
+   *
+   */
+  public compute = async (
+    request: GeospatialComputeRequest
+  ): Promise<GeospatialComputeResponse> => {
+    const path = this.url();
+    const response = await this.post<GeospatialComputeResponse>(path, {
+      data: request
+    });
+    return response.data
+  };
+}
--- a/packages/stable/src/api/geospatial/computeAPI.ts
+++ b/packages/stable/src/api/geospatial/computeAPI.ts
@@ -8,6 +8,21 @@ export class ComputeAPI extends BaseResourceAPI<GeospatialComputedItemList> {
   /**
    * [Compute custom json output structures or well known binary format responses based on calculation or selection of feature properties or direct values given in the request.](https://docs.cognite.com/api/v1/#operation/compute)
    *
+   * ```js
+   *
+   * const response = await client.geospatial.compute.compute({
+   *   output: {
+   *     "output": {
+   *       stTransform: {
+   *         geometry: {
+   *           ewkt: "SRID=4326;POLYGON((0 0,10 0,10 10,0 10,0 0))"
+   *         },
+   *         srid: 4326,
+   *       }
+   *     }
+   *   }
+   * });
+   * ```
    */
   public compute = async (
     request: GeospatialJsonComputeOutput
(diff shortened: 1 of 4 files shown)


packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
   srid: number;
 }
+
+export type GeospatialComputeFunction =
+  | {
+      stTransform: GeospatialGeometryTransformComputeFunction;
+    };
+
+export interface GeospatialGeometryTransformComputeFunction {
+  geospatialGeometryComputeFunction: GeospatialGeometryComputeFunction;
```
saagar-cognite: suggestion(blocking): I think the key should be just `geometry`.

BenChand: Oh you're right, will update

packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
+  srid: number;
+}
+
+export type GeospatialGeometryComputeFunction =
+  | {
+      geometryValue: GeospatialGeometryValueComputeFunction;
```
saagar-cognite: suggestion(blocking): the key here should be `ewkt`

packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
+  | {
+      stTransform: GeospatialGeometryTransformComputeFunction;
+    };
+
+export interface GeospatialGeometryTransformComputeFunction {
+  geospatialGeometryComputeFunction: GeospatialGeometryComputeFunction;
+  srid: number;
+}
+
+export type GeospatialGeometryComputeFunction =
```
saagar-cognite: suggestion(blocking): `GeospatialGeometryComputeFunction` should be just extending `GeospatialGeometryValueComputeFunction`.
Something like

```suggestion
      export interface GeospatialGeometryComputeFunction extends GeospatialGeometryValueComputeFunction {
      }
```

packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
+  | {
+      geometryValue: GeospatialGeometryValueComputeFunction;
+    };
+
+export interface GeospatialGeometryValueComputeFunction {
```
saagar-cognite:
```suggestion
export interface GeospatialExtendedWellKnownText {
```

saagar-cognite: should replace `GeospatialGemetryValueComputeFunction` in above functions as well.

packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
+
+export interface GeospatialComputedItem {
+  item: Map<string, any>;
+}
+
+export interface GeospatialComputedItemList {
+  items: GeospatialComputedItem[];
+}
+
+export interface GeospatialComputeRequest {
```
saagar-cognite:
```suggestion
export interface GeospatialComputeRequest extends GeospatialJsonComputeOutput {
}
```

saagar-cognite:
```
interface GeospatialJsonComputeOutput {
output: Record<string, GeospatialComputeFunction>
}
```

saagar-cognite: to make it in synergy with the docs

packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
+  output: Map<string, GeospatialGeometryComputeFunction>
```
saagar-cognite: suggestion(blocking): `Map` should be replaced with `Record`

packages/stable/src/api/geospatial/types.ts
```diff
@@ -269,3 +269,38 @@ export interface GeospatialCRSResponse
...
+}
+
+export interface GeospatialComputeResponse {
+  items: GeospatialComputedItemList;
```
saagar-cognite: suggestion(blocking): we generally do not need `items` in our types as that adapter is taken care by the base sdk.
Need to figure out which endpoint method we've to use for compute.
We've options like `listEndpoint`, `retrieveEndpoint`, etc.

packages/stable/src/api/geospatial/computeAPI.ts
```diff
@@ -0,0 +1,21 @@
...
+  GeospatialComputeRequest,
+  GeospatialComputeResponse
+} from './types';
+
+export class ComputeAPI extends BaseResourceAPI<GeospatialComputeResponse> {
+  /**
+   * [Compute custom json output structures or well known binary format responses based on calculation or selection of feature properties or direct values given in the request.](https://docs.cognite.com/api/v1/#operation/compute)
+   *
+   */
+  public compute = async (
```
saagar-cognite, line 27: suggestion(blocking): could you please check the implementation of `search` or `retrieve` features and see if we can use an existing `endpoint` method for not having to use list in our types and having the response already adapter to our need with `items` already extracted.
This is pretty standard in the sdk.

BenChand: I looked through the various methods in the `baseResourceApi.ts` file and couldn't find an existing endpoint method that supported:

* No suffix to url
* No requirement of the query being a filter type
* Doesn't require the response to be a cursor list

Might be missing something, but seems to me like the custom response interface might be needed here (?)

packages/stable/src/api/geospatial/types.ts
BenChand: I think this has to be the other way around right? Because the `GeospatialGeometryValueComputeFunction` is just one function in the `GeospatialGeometryComputeFunction` group? Or at least it will be when the rest of the compute endpoint is stabilized

packages/stable/src/api/geospatial/types.ts
BenChand: That has been replaced with an interface

23andreas (approved):

bdimitrijoski (commented):
@BenChand can you fix the tests please?

BenChand:
> @BenChand can you fix the tests please?

Looks like some unrelated tests are failing, not sure there's much I can change to fix those!

```
FAIL packages/stable/src/__tests__/api/visionApi.int.spec.ts (6.298 s)
  ● Vision API › retrieve extract job › waitForCompletion=false
```

PR #901, opened 2022-10-25
feat: add units conversion for js sdk

harish0489-cognite:


samples/nodejs/legacy-auth-typescript/quickstart.ts
```diff
@@ -19,20 +19,21 @@ async function quickstart() {
...
   });
 
   console.log(
-    `\nThe SDK client is now configured to talk with Cognite project "${
-      project
-    }"\n`
+    `\nThe SDK client is now configured to talk with Cognite project "${project}"\n`
   );
 
   // get maximum 5 root assets
```
cognite-ornellas, line 29: there should not be changes to this class

packages/stable/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "@cognite/sdk-core": "^4.6.6",
     "geojson": "^0.5.0",
+    "@cognite-ornellas/units": "^1",
```
cognite-ornellas: this is wrong... should be @cognite/units

packages/stable/src/__tests__/api/datapoints.int.spec.ts
```diff
@@ -44,6 +51,10 @@ describe('Datapoints integration test', () => {
       start: '2d-ago',
       end: new Date(),
     });
+
+    console.log('Unconverted retrieve values: ');
```
cognite-ornellas: please remove the comments

samples/nodejs/legacy-auth-typescript/quickstart.ts
```diff
@@ -19,20 +19,21 @@ async function quickstart() {
...
     project,
     apiKeyMode: true,
-    getToken: () => Promise.resolve(apiKey)
+    getToken: () => Promise.resolve(apiKey),
   });
 
   console.log(
-    `\nThe SDK client is now configured to talk with Cognite project "${
-      project
-    }"\n`
```
cognite-ornellas, line 28: there should not be changes in this file I guess

samples/typescript/units/.gitignore
```diff
@@ -0,0 +1 @@
+quickstart.js
```
cognite-ornellas: why ignore quickstart.js?

samples/typescript/units/README.md
```diff
@@ -0,0 +1,117 @@
...
+## How to get logged by yourself?
+
+1. Add the Cognite SDK and @azure/msal-node to your project with yarn or npm.
+
+```sh
+yarn add @cognite/sdk@6.1.1
+yarn add @azure/msal-node
+```
+
+```sh
```
cognite-ornellas: this document is about authwrapper... it need to be about units conversion

samples/typescript/units/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "private": true,
+  "license": "Apache-2.0",
+  "scripts": {
+    "tsc": "tsc",
+    "start": "yarn tsc && node build/quickstart.js"
+  },
+  "dependencies": {
+    "@cognite/auth-wrapper": "^1",
+    "@cognite/sdk": "^7",
+    "@cognite/units": "^1.0.2"
```
cognite-ornellas, line 13: maybe it a good idea to change to "@cognite/units": "^1"

cognite-ornellas (changes requested):

packages/stable/package.json
harish0489-cognite: @cognite/units packaged added

samples/typescript/units/package.json
harish0489-cognite: I guess this came in from semantic version format.

packages/stable/src/__tests__/api/datapoints.int.spec.ts
harish0489-cognite: removed

andersfylling:
I'm not familiar with the case, but I was wondering what the benefits are of keeping this within the endpoint method vs. just doing the transformation after retrieving the results?

cognite-ornellas:
Anders, I believe that the benefit of it is encapsulating the logic of what
is needed to do the the unit conversion for timeseries endooint. If you
check the class called dataPointsUnitConverter you will notice that there
are some code that is being abstracted. So the point of it is enabling more
with writing less code. The more you can do without needing to deal with
too many details, the better.

Without this PR's code you would need to find the library, add it into your
project and write most of code that is found in it (with is not so litle
code) and, ofcourse deal with mistakes :)

Em qui., 27 de out. de 2022 23:11, Anders Øen Fylling <
***@***.***> escreveu:

andersfylling:
I'm all for having utility package or just simply helper functions (as an addition to a simplistic SDK - thinking about long term maintenance) :+1: 

What I was wondering about is why we want to adjust the endpoint methods, instead of just running transformations afterwards?
It also clarifies, in code, what is a server response and client side mutated data.

It also looks to me that the try catch discards any errors, so it's not possible for the user to verify programmatically if the transformation was successful or not. However, if we just make the transformation tools available, instead of as an argument option, the user can control error handling themselves.

Just throwing in my two cents here, I might be missing something ^^,

cognite-ornellas:
I understand and appreciate you opinion Anders.
About the try-catch, I think it should not supress the error. There is
option that will allow to just warn when conversion fails, but its not the
default behaviour.
About segregating what is a server response or not, I dont think that an
sdk should be limited to reproducing exactly the same behavior of the API.
Other than that, I think it makes sense to make the use of the API easier
and more intuitive.

Em sex., 28 de out. de 2022 00:24, Anders Øen Fylling <
***@***.***> escreveu:

andersfylling:
@cognitedata/development-experiences could you take a decision on this?

BugGambit:
I will close this PR. The original creators are not around anymore, and no push from PM to release this. Feel free to re-open a PR if you disagree.

PR #905, opened 2022-10-26
feat(annotations): add annotation payload types

haakonflatval-cognite:
This adds payload types for annotations, generated using the codegen setup.

In order to only create the payload types, and not e.g. request/response types, I have changed the codegen source to allow for a filter for "relevant" types, which it will use as root of recursion when determining OpenAPI definitions to translate.

The reason I only wanted to create payload types, is that request/response types contain timestamps with the `EpochTimeStamp` type, which is just a 64-bit int. Codegen translates these to `number`s, while the current types use JavaScript `Date` objects for dates, so suddenly using these new types would introduce a breaking change, not to mention the renames of all the query types.

The previous type used for annotation payloads was just `object`, which makes me believe we can consider the changes here non-breaking.

I have observed that the newly-generated types partially overlap with the auto-generated types for the vision API. There is currently no way to de-duplicate these types with codegen to my knowledge, so I have left it the way it is.


packages/codegen/src/codegen.ts
```diff
@@ -289,12 +322,17 @@ export class CodeGen {
    * @returns names of generated types
    */
   public generateTypes = async (
-    openApiDoc: OpenApiDocument
+    openApiDoc: OpenApiDocument,
+    relevantReferenceNames: string[]
```
andersfylling: Can you make this optional?

packages/codegen/src/__tests__/codegen.uint.spec.ts
```diff
@@ -110,7 +110,7 @@ describe('code generation', () => {
         'LimitList',
       ];
 
-      const typeNames = await gen.generateTypes(basicSnapshot);
+      const typeNames = await gen.generateTypes(basicSnapshot, []);
```
andersfylling: change could be removed if type is optional, see other comment. Same for other places.

andersfylling (approved):
Unfortunate with the Date type, but I think this is an acceptable way to handle it until we have a proper fix.

andersfylling:
Could you expand on this? Which types exactly, and are these types sharing the same reference in open api or are they duplicate in the open api spec as well?
> I have observed that the newly-generated types partially overlap with the auto-generated types for the vision API. There is currently no way to de-duplicate these types with codegen to my knowledge, so I have left it the way it is.

haakonflatval-cognite:
> Could you expand on this? Which types exactly, and are these types sharing the same reference in open api or are they duplicate in the open api spec as well?

They share the same reference in OpenAPI. VisionAPI use many definitions that belong to the annotations API because annotations are used for storing e.g. information about tags and detections in images. For instance, the vision type `VisionExtractPredictions` references `AnnotationsTextRegion` and a few other annotation types,, that are automatically generated with the vision types. Conveniently, these duplicate types were removed automatically from the `exports.gen.ts` file when they were generated in the Annotations API.

marvrez (commented):
LGTV (Looks good to vision). Regarding the redundant types, we can address that in a separate PR :)

marvrez (approved):

PR #907, opened 2022-10-28
fix: add codegen support for CusrorAndAsyncIterator 

andersfylling:
### Why do we need sorting?
Deterministic code output. Not sorting the produced code will cause horrible diffs, and make it hard to verify if code was correctly generated on changes.

### Why switch to AST based?

We need the AST handling for other things, such as auto pagination, etc. It also ensures that whatever library we decide to use for `open api` to `ts` conversion won't affect ordering of generated code. The previous sorting version is also incorrect, as it is not in alphabetical order and will in the future suddenly fail as it's not deterministic.

### The diff is not as large as it seems
The diff looks large, at 1,886 additions and 2,189 deletions, but is mostly due to changes in formatted code. Some extra new lines were dropped when converting the abstract syntax tree to typescript. This has no effect on behavior. I just did not invest time in programmatically call the linter after generating code, as the output is more than pretty enough for it's purpose.

In reality the diff should be more like; `92 additions and 19 deletions.`


bdimitrijoski (commented):
@andersfylling I am not familiar with this code, in fact, I am seeing this for the first time, but what exactly are you trying to do here? If I understand correctly, you are generating TS types from AST but from what exactly? From GraphQL, JSON?
My question here is can't we just reuse something that is already out there instead of creating a custom one?

andersfylling:
We poll the open api spec, generated by our service contract repo and generating typescript from it.

However, the library that does this conversion is a bit limited and we need additional transformation - only way to do this without some regex fun is by converting the code to AST and manipulating the tree. Then we generate it back to typescript.

One issue from TS => AST => TS conversion is that the AST nodes lose ordering. So we've applied a sorting step to make sure that the outputted code is always in order - producing clear and easy to read git diff's.

andersfylling:
There is a library that creates AST from openapi, but we would still have to deal with sorting ourselves.

andersfylling:
Note that, this PR was originally a part of https://github.com/cognitedata/cognite-sdk-js/pull/908 but I split it up to make reviewing a little easier and feature focused.

We are adding this AST transformer layer so we can support features like auto pagination on generated typescript types.

23andreas (approved):

andersfylling:
Adjusting the title to make it more obvious in releases what has changed.

miladcognite (approved):
Changes in vision types LGTM

polomani:
@andersfylling I know generation is fun and all, but have you considered the time spend for developing this and also the cost of future maintenance? In 2019 the team owning the SDK invested some time and concluded that SDK generation gives more headache than it provides a value:
- https://cognitedata.atlassian.net/wiki/spaces/DEVX/pages/1005715913/2019-07-30+SDK+generation+decision+meeting
- based on (https://cognitedata.atlassian.net/wiki/spaces/DEVX/pages/1005912434/SDK+Generation)


Imho this code becomes quite cumbersome here, and it doesn't take much time to write a TS interface with proper types  and descriptions. (also future proof) Everything else is already a part of SDK-core anyway

tommythorsen (approved):

queCognite (approved):

andersfylling:
@polomani you raise good points, so I talked a little more with @bdimitrijoski and we decided we will merge this. However, we won't automate it fully. Teams will have this as an opt-in tool.

> I know generation is fun and all, but have you considered the time spend for developing this...

It's been a learning experience for sure, but it's also been a win for me. Maintaining our relevant code (around 2,000loc for just the documents service) I fixed several bugs thanks to codegen (typos, outdated code, etc.) - it also added nice comments for our users. It also allows other teams that aren't currently in the SDK to onboard much quicker.

Development time is a complex topic here, I think this could have been done much faster. Priorities where elsewhere, for both contributors of codegen and reviewers - so there's months of dead time that makes it seems we've spent more time on this than we really did.

> Imho this code becomes quite cumbersome here, and it doesn't take much time to write a TS interface with proper types and descriptions.

I and @bdimitrijoski agree with you. I sent you a slack thread for more insight. I'm not really interested in pushing more work on people since concerns are being raised now, so we've instead decided that if need be, we will just pull the plug on this if maintenance or something else becomes an issue.


#### From the link you sent
> The tools we tested didn’t create good type names, correct descriptions, lack of union type support, single-language solutions etc.

 - Generated type names now reflect the same names as in the Cognite documentation, I consider this to be a good thing.
 - Descriptions look correct to me (code comments, I figure).
 - Union type seems to be correct as well. They just haven't added support for "never" in a field type, but it doesn't seem to be used(?) anyways.

PR #914, opened 2022-11-01
chore(workflow): add no verify for lerna publish

saagar-cognite:


BugGambit (approved):
Let's try even though the option is deprecated

PR #916, opened 2022-11-09
chore: disable SSL check for no auth mode

ghost:


BugGambit:
@dlcognite do you have a use-case in mind?

ghost:
Yes this is for our local mock server right now every time we instantiate a client for testing (where we skip auth) it would throw out this error making debugging ci/cd super hard

bkuzma (approved):

BugGambit (approved):
LGTM

PR #932, opened 2022-12-13
feat(alpha): monitoring task listing and model change

VerstraeteBert:
- Listing endpoint: only allow querying by TimeseriesId, not by TimeseriesExternalId
- Create endpoint: embed ModelExternalId and parameters into Model field

### DIFF
--- a/packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
+++ b/packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
@@ -1,6 +1,10 @@
 // Copyright 2022 Cognite AS
 
-import { Channel, MonitoringTaskThresholdModelCreate } from 'alpha/src/types';
+import {
+  Channel,
+  ModelExternalId,
+  MonitoringTaskThresholdModelCreate,
+} from 'alpha/src/types';
 import CogniteClientAlpha from '../../cogniteClient';
 import {
   CLIENT_ID,
@@ -21,8 +25,8 @@ describe('monitoring tasks api', () => {
   const monitoringTaskExternalId = `test_mt_${ts}`;
   const channelExternalId = `test_channel_mt_${ts}`;
   const sessionsApi = `/api/v1/projects/${TEST_PROJECT}/sessions`;
-  const testMtModel = {
-    externalId: 'threshold',
+  const testMtModel: MonitoringTaskThresholdModelCreate = {
+    externalId: ModelExternalId.THRESHOLD,
     timeseriesExternalId: 'test-functions',
     threshold: 50.1,
     granularity: '1m',
@@ -64,7 +68,7 @@ describe('monitoring tasks api', () => {
         interval: testMtInterval,
         nonce: sessionsRes?.data?.items[0]?.nonce,
         overlap: testMtOverlap,
-        model: testMtModel as MonitoringTaskThresholdModelCreate,
+        model: testMtModel,
       },
     ]);
     expect(response.length).toBe(1);
--- a/packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
+++ b/packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
@@ -34,6 +34,7 @@ describe('monitoring tasks api', () => {
   const testMtOverlap = 1000 * 60;
   const testMtInterval = 5 * 60 * 1000;
   const expectedResponseModel = {
+    externalId: "threshold",
     timeseriesId: 4944699311094690,
     granularity: testMtModel.granularity,
     threshold: testMtModel.threshold,
--- a/packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
+++ b/packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
@@ -1,6 +1,6 @@
 // Copyright 2022 Cognite AS
 
-import { Channel } from 'alpha/src/types';
+import { Channel, MonitoringTaskThresholdModelCreate } from 'alpha/src/types';
 import CogniteClientAlpha from '../../cogniteClient';
 import {
   CLIENT_ID,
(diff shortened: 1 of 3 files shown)


packages/alpha/src/types.ts
```diff
@@ -17,47 +17,64 @@ import {
...
+  constructor(
+    timeseriesExternalId: CogniteExternalId,
+    threshold: number,
+    granularity?: string
+  ) {
+    this.timeseriesExternalId = timeseriesExternalId;
+    this.granularity = granularity;
+    this.threshold = threshold;
+  }
+}
```
polomani, line 31: we don't do classes here, this is overly complex for the SDK :)
```
export interface MonitoringTaskThresholdModelCreate {
  externalId: "threshold";
  timeseriesExternalId: CogniteExternalId;
  threshold: number;
  granularity?: string;
```
should be enough

packages/alpha/src/types.ts
```diff
@@ -17,47 +17,64 @@ import {
 
 export * from '@cognite/sdk';
 
-export interface MonitoringTaskParametersCreate {
-  timeseriesExternalId: CogniteExternalId;
-  granularity?: string;
-  threshold: number;
+enum ModelExternalId {
+  Threshold = 'threshold',
+}
```
polomani: please do not use enums they're quite painful for the consumer.

`paramName: "string_val1" | "string_val1"` is usually fine

packages/alpha/src/types.ts
VerstraeteBert: Tried this earlier. Interfaces don't seem to play well with literals, when constructing this and using it usually complains that (similar to) ` externalId: type "threshold" expected, got type string. `. Can mitigate this by not duck typing anymore and explicitly casting the object MonitoringTaskThresholdModelCreate (see new commit), which is not ideal either.

VerstraeteBert: Now users both have to declare explicitely which Model they're using in the cast and redundantly in the object iself with the externalId. By using the class and hidden enum this is not necessary.

packages/alpha/src/types.ts
VerstraeteBert: Enums were hidden for users, only used internally.

packages/alpha/src/types.ts
polomani: classes give different sorts of headaches, trust me on this one. js SDK has a long history of going from interfaces to classes and then back to interfaces.

If you wish to improve the developer experience you can define a type + const.
Example: 

```
export type ContextJobStatus = 'QUEUED' | 'RUNNING' | 'COMPLETED' | 'FAILED';

export const ContextJobStatus = {
  QUEUED: 'QUEUED' as ContextJobStatus,
  RUNNING: 'RUNNING' as ContextJobStatus,
  COMPLETED: 'COMPLETED' as ContextJobStatus,
  FAILED: 'FAILED' as ContextJobStatus,
};
```

packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -21,14 +21,20 @@ describe('monitoring tasks api', () => {
   const monitoringTaskExternalId = `test_mt_${ts}`;
   const channelExternalId = `test_channel_mt_${ts}`;
   const sessionsApi = `/api/v1/projects/${TEST_PROJECT}/sessions`;
-  const testMtParams = {
+  const testMtModel = {
```
polomani:
```suggestion
  const testMtModel: MonitoringTaskThresholdModelCreate
```
to get a proper type inference, as you declare this outside of the request

polomani: similar to `model: testMtModel as MonitoringTaskThresholdModelCreate` but better

polomani:
`yarn lint:fix` + remove the comments

packages/alpha/src/types.ts
```diff
@@ -17,47 +17,53 @@ import {
 
 export * from '@cognite/sdk';
 
-export interface MonitoringTaskParametersCreate {
+export type ModelExternalId = 'threshold';
+
+export const ModelExternalId = {
```
```suggestion
export const MonitoringTaskModelExternalId = {
```
prepend API name for users to get a nice autocomplete

packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -1,6 +1,10 @@
 // Copyright 2022 Cognite AS
 
-import { Channel } from 'alpha/src/types';
+import {
+  Channel,
+  ModelExternalId,
+  MonitoringTaskThresholdModelCreate,
+} from 'alpha/src/types';
```
polomani:
```suggestion
} from '../../types';
```

polomani: this is the import issue you have

packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -21,14 +25,20 @@ describe('monitoring tasks api', () => {
   const monitoringTaskExternalId = `test_mt_${ts}`;
   const channelExternalId = `test_channel_mt_${ts}`;
   const sessionsApi = `/api/v1/projects/${TEST_PROJECT}/sessions`;
-  const testMtParams = {
+  const testMtModel: MonitoringTaskThresholdModelCreate = {
+    externalId: ModelExternalId.THRESHOLD,
+    timeseriesExternalId: 'test-functions',
```
polomani:
```suggestion
    timeseriesExternalId: 'test_functions',
```

polomani (approved):

packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -1,6 +1,10 @@
 // Copyright 2022 Cognite AS
 
-import { Channel } from 'alpha/src/types';
+import {
+  Channel,
+  ModelExternalId,
```
polomani:
```suggestion
  MonitoringTaskModelExternalId,
```

packages/alpha/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -21,14 +25,20 @@ describe('monitoring tasks api', () => {
   const monitoringTaskExternalId = `test_mt_${ts}`;
   const channelExternalId = `test_channel_mt_${ts}`;
   const sessionsApi = `/api/v1/projects/${TEST_PROJECT}/sessions`;
-  const testMtParams = {
-    threshold: 50.1,
+  const testMtModel: MonitoringTaskThresholdModelCreate = {
+    externalId: ModelExternalId.THRESHOLD,
```
polomani:
```suggestion
    externalId: MonitoringTaskModelExternalId.THRESHOLD,
```

hamdielhamdi-cognite (approved):

sou1est (approved):

PR #937, opened 2023-01-12
feat(documents): add new document search functionality

tommythorsen:
https://cognitedata.atlassian.net/browse/SBS-5737


packages/stable/src/api/documents/types.gen.ts
```diff
@@ -356,8 +356,38 @@ export type DocumentsAggregateCountRequest = {
...
+/**
+ * Find all metadata property names
+ */
+export type DocumentsAggregateUniquePropertiesRequest = {
+    search?: {
+        query: string;
+    };
+} & DocumentSearchFilter & {
+    aggregate?: "uniqueProperties";
+    properties?: {
```
andersfylling: Interesting. Seems there's a codegen bug as these fields are mandatory.

andersfylling (approved):

packages/stable/src/api/documents/types.gen.ts
andersfylling: This is fine though. Adjusting this to required fields later won't cause a breaking change, so I'm clicking approve.

tommythorsen: The object this is based on is quite complex, with a big `allOf` with multiple inline objects. I am not surprised it is difficult for the autogenerator to deal with.

Perhaps it would work better if the `required` clause in the spec was inside the same object where the required properties are. I'll test that

tommythorsen: yup, that works

tommythorsen: I have updated the service contract here: https://github.com/cognitedata/service-contracts/pull/2020

andersfylling (changes requested):
Remember to update the title and get the latest changes from our openapi

tommythorsen:
> Remember to update the title and get the latest changes from our openapi

Done

andersfylling (approved):
LGTM

PR #936, opened 2023-01-09
chore(stable): replace apiKey with msal-node in setupLoggedInClient

sou1est:

PR #938, opened 2023-01-17
feat: translation and scale for 3D revisions [release]

haakonflatval-cognite:
This implements the new scale- and translation feature of the 3D API in the SDK.

API doc PR: https://github.com/cognitedata/service-contracts/pull/2026


packages/stable/src/types.ts
```diff
@@ -1990,6 +1990,16 @@ export interface Revision3D {
...
+  /**
+   * Global translation to be applied to the entire model.
+   * The translation is expressed in meters.
+   */
+  translation?: Tuple3<number>;
+  /**
+   * Global scale to be applied to the entire model.
+   * The scale is expressed as scale along the X, Y and Z axes.
+   */
+  scale?: Tuple3<number>;
```
andersfylling, line 2002: Just out of curiousity: is there any reason we have an alias for an array? Is there some type constraint somewhere in the SDK or is it just a hint for users?

andersfylling (approved):
LGTM!

A fix for the tests is in progress: https://github.com/cognitedata/cognite-sdk-js/pull/936

packages/stable/src/types.ts
haakonflatval-cognite: I just took the same type as `rotation`, so I didn't think about it :sweat_smile: But I expected it to be defined as `[T, T, T]`, and not `T[]`. Not sure why it is like this :thinking:

haakonflatval-cognite:
Thanks for review - I'll have to wait to merge this until the `service-contracts` PR linked in the description is done

PR #949, opened 2023-02-02
chore: enable dependabot

23andreas:


borajimin (approved):

andersfylling:
We are moving to renovate in general, could be interesting to just set that up instead?

andersfylling (approved):

23andreas:
> We are moving to renovate in general, could be interesting to just set that up instead?

good point, I've messaged security asking them to set it up.

andersfylling:
@23andreas just in case I miss out on the configuration PR. It's nice to only run updates during weekends, and only do so for releases thats been out for a few days. Helps reducing spam for everyone that gets automatically tagged.

PR #982, opened 2023-02-09
docs: replace invalid mimetype image/jpg with image/jpeg in examples

larsmoa:
`image/jpg` is not a valid mimetype, `image/jpeg` is the correct spelling. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types


BugGambit (approved):
LGTM

andersfylling:
Note that changing comments in the snapshot file has no effect, and perhaps we should adjust the open api spec examples as well (different repo), but I'll leave that to OP.

andersfylling (approved):

larsmoa:
> Note that changing comments in the snapshot file has no effect, and perhaps we should adjust the open api spec examples as well (different repo), but I'll leave that to OP.

What repo is this, @andersfylling ?

andersfylling:
@larsmoa sent you a message with more insight. See here for local documentation: https://github.com/cognitedata/cognite-sdk-js/blob/master/CONTRIBUTING.md#code-generation

PR #984, opened 2023-02-10
fix: add yarn resolutions, unfreeze yarn.lock for dev

stefanmaretic:
There'a likelihood some previously resolved CVEs might pop-up again, but we'll fix them afterward.

There's also a likelihood `loader-utils` might cause an issue because there's 2 major versions we need to support.

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,1 +1,0 @@
---install.frozen-lockfile true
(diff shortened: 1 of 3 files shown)


BugGambit (approved):
LGTM

PR #986, opened 2023-02-13
fix: add marked in resolution

borajimin:


andersfylling (approved):

andersfylling:
Just out of curiosity, this doesn't seem to add any more granularity than the markdown link checker dependency does. Is there a reason we need to add this?

https://github.com/tcort/markdown-link-extractor/blob/master/package.json

borajimin:
@andersfylling Hi :) I'm trying to mitigate this security vulnerability: https://github.com/cognitedata/cognite-sdk-js/security/dependabot/74

My first approach to directly update marked package didn't work properly. So, I'm testing some scenarios 😬

andersfylling:
Oh! nice!

What if you just update the version for markdown-link-check to be at least 3.10.3? seems to be missing a few minor versions

borajimin:
I can try that :)

borajimin:
image
looks like the problem comes from "typedoc" I can try updating it and see if I don't break it

andersfylling (approved):

PR #997, opened 2023-03-09
fix: fix sample apps on newer node versions

bdemirel:
https://hub.cognite.com/developer-and-user-community-134/cognitedata-cognite-sdk-js-1616

User was trying to install and start this sample inside the monorepo with node 16.19.1 on an M1 Mac, and it didn't work. I have updated necessary packages to hopefully solve all these issues. I have tried:
1- node@16 with yarn, inside the monorepo
2- node@18 with yarn, inside the monorepo
3- node@16 with yarn, copied sample outside
4- node@18 with yarn, copied sample outside
5- node@16 with npm, copied sample outside
6- node@18 with npm, copied sample outside

I have also tried npm inside the monorepo, but because npm doesn't read from `yarn.lock`, it causes a lot of problems.
I'd appreciate if you can try this on an M1 Mac as well, to make sure everything works fine; I have some doubts.
Mainly because even though I saw errors trying to run it first, they were different from what the user has posted.

**UPDATE:**
- Fixed all samples. They should all be working.
- Except the legacy auth ones. They got  the :axe: .


bdimitrijoski (approved):

package.json
```diff
@@ -79,6 +80,7 @@
     "yargs": "^17.5.1"
   },
   "resolutions": {
+    "@types/prettier": "2.6.0",
```
bdemirel, line 83: 2.6.1 and above requires typescript@^4.2.0. I don't know why this wasn't an issue before, but now `yarn test-snippets` was failing because of this.

package.json
```diff
@@ -56,6 +56,7 @@
     "@types/yargs": "^17.0.10",
     "@typescript-eslint/eslint-plugin": "^3.4.0",
     "@typescript-eslint/parser": "^3.4.0",
+    "axios": "^0.27.2",
```
bdemirel, line 59: Our devDependency `swagger-typescript-api-nextgen` requires this, even though it doesn't say it :upside_down_face:

bdemirel:
CI thinks no packages have changed, so skips all tests. But since I've changed `yarn.lock`, I've run all of them locally. I don't see any problems.

anjalirai2626:
Dropping an additional comment i.e. tested locally with node `v16.19.1`, looks good!

Screenshot 2023-03-29 at 14 36 19

PR #998, opened 2023-03-17
chore: remove dead links

bdemirel:
Playground API documentation has been removed in an effort to deprecate it. 
YouTube thumbnail link is broken and I believe the video itself is private.


andersfylling (approved):
Approved to unblock PRs :partying_face: 

Btw, I think we can just remove the playground package at this point. I don't see the reason to have undocumented code. May just confuse users and we aren't planning on future realses for playground packages(?).

PR #1002, opened 2023-03-30
chore: update README auth & testing sections

sou1est:


bdemirel (approved):
Looks good to me!
But it's kinda weird that the tests fail now. They fail because of changes in my PR, why would they not fail then?

bdemirel:
> Looks good to me! But it's kinda weird that the tests fail now. They fail because of changes in my PR, why would they not fail then?

Ahh, nvm, they fail now because the links are to the master branch, and the master branch links got dead after my changes. Of course haha! :sweat_smile:

PR #1005, opened 2023-04-04
feat(core): add optional retryValidator to override the default one

franrull:
The current implementation has a max number of retries hardcoded at 5 and there is no way to modify this for the whole CDF Client instance. This little change should allow for such override.


packages/core/src/baseCogniteClient.ts
```diff
@@ -201,7 +202,7 @@ export default class BaseCogniteClient {
   ) {
     const httpClient = new CDFHttpClient(
       getBaseUrl(baseUrl),
-      this.getRetryValidator()
+      options.retryValidator || this.getRetryValidator()
```
cemreyavuz: I think adding this logic here might be a bit confusing: `getRetryValidator` fn would return a different fn while we use `options.retryValidator` for `CDFHttpClient`.

Instead, either `getRetryValidator` fn should return `options.retryValidator` with a fallback to `createUniversalRetryValidator` - or we can add a private variable (ex: `retryValidator`) inside the class and initialize it in constructor with the same logic (and `getRetryValidator` would return `this.retryValidator`).

franrull: I like the second option, let me make the change.

franrull: @cemreyavuz can you review again?

cemreyavuz (approved):
LGTM

BugGambit (approved):
LGTM

PR #1007, opened 2023-04-18
chore(stable): fix CVE issue

sou1est:


andersfylling:
Could you provide some more insight into this problem you've solved? Did you run a script that we can add in github actions?

andersfylling:
I'm not sure I understand this change. What is this empty key?
image

Also note that, with the workspaces here the idea is to define all dependencies in the root, and let lerna package and link everything. At least, that's my mental model of how this is supposed to work, but I think cdf UX could confirm/deny this.

sou1est:
> I'm not sure I understand this change. What is this empty key? image
> 
> Also note that, with the workspaces here the idea is to define all dependencies in the root, and let lerna package and link everything. At least, that's my mental model of how this is supposed to work, but I think cdf UX could confirm/deny this.

I just ran `npm audit fix`. But I was surprised that some of the packages use `package-lock.json` while the root of the project relies on yarn workspaces. All of the changes are generated so not sure of empty keys over there

sou1est:
@andersfylling somehow both `core` and `stable` packages ended up with package-lock.json files instead of looking up at the root level `yarn.lock`. Looking at the git history it happened somewhat year ago.

sou1est:
@andersfylling Now I think removing package.lock json should actually be fine, and it should as well fix CVEs i hope

andersfylling:
> @andersfylling Now I think removing package.lock json should actually be fine, and it should as well fix CVEs i hope

I agree. That sounds like the right thing to do.

bkuzma (approved):

PR #1009, opened 2023-04-27
feat(project): project update user profiles

birkmose:
Update project configuration to support updating user profiles settings


sou1est:
@birkmose could you also remove unused var, since they are treated as errors  
src/__tests__/api/login.int.spec.ts:4:10 - error TS6133: 'setupClient' is declared but its value is never read.
 import { setupClient, setupLoggedInClient } from '../testUtils';
 Also there is vision api test case which became flaky very recently, don't know exactly why but last week just restarting CI job multiple times helped

sou1est (approved):

PR #1010, opened 2023-05-15
feat: annotation reverse-lookup [release]

haakonflatval-cognite:
Add support for the new annotations reverse lookup endpoint. Documentation PR: https://github.com/cognitedata/service-contracts/pull/2199


packages/stable/src/api/annotations/annotationsApi.ts
```diff
@@ -140,4 +143,32 @@ export class AnnotationsAPI extends BaseResourceAPI<AnnotationModel> {
...
+   *     data: {
+   *       { id: 123 }
+   *     }
+   *   }
+   * };
+   *
+   * const resourceIdsResponse = this._client.annotations.reverseLookup(assetQuerydata);
+   */
+  public reverseLookup = (
+    filter: AnnotationReverseLookupRequest
```
andersfylling, line 165: is filter optional?

olacognite (commented):
Looks like black magic to me, but seems right!

packages/stable/src/api/annotations/annotationsApi.ts
haakonflatval-cognite: No, I would expect the user to supply this for the request to make sense. I think we should wait with merging this until the API doc is ready, so that it's clear what the API looks like :P

olacognite: https://github.com/cognitedata/service-contracts/pull/2199 API doc PR here

haakonflatval-cognite:
Re-requesting review since the API Doc PR has been merged

haakonflatval-cognite:
@andersfylling Could you have another look at this?

andersfylling:
Looks like the service-contracts says this is in beta, but you've published it in stable. I think these should match.

https://github.com/cognitedata/service-contracts/blob/bf4b7514f921c74978d3c2cce3b7db90b664c8e4/versions/v1/annotations.preproc.yml#LL205C16-L205C16

andersfylling:
Names in service contracts doesn't match the names in the js sdk, so I just need a bit of time to get around that

packages/stable/src/types.ts
```diff
@@ -3306,3 +3306,17 @@ export interface AnnotationFilterProps {
...
+    FilterQuery {}
+
+export interface AnnotationReverseLookupFilter {
+  filter: AnnotationReverseLookupFilterProps;
+}
+
+export interface AnnotationReverseLookupFilterProps {
+  annotatedResourceType: AnnotatedResourceType;
+  annotationType: AnnotationType;
+  data?: Record<string, any>;
```
andersfylling, line 3325: missing: "creatingApp", `creatingAppVersion`, `creatingUser`, `status`.

packages/stable/src/api/annotations/annotationsApi.ts
```diff
@@ -140,4 +143,32 @@ export class AnnotationsAPI extends BaseResourceAPI<AnnotationModel> {
...
+  ): CursorAndAsyncIterator<AnnotationsAssetRef> => {
```
andersfylling: Missing `annotatedResourceType` in response?

packages/stable/src/api/annotations/annotationsApi.ts
```diff
@@ -12,6 +13,8 @@ import {
   AnnotationSuggest,
   AnnotationFilterRequest,
   AnnotationModel,
+  AnnotationReverseLookupRequest,
+  AnnotationsAssetRef,
```
andersfylling, line 19: can you link me the definition, I'm struggling to find it for some reason  :thinking:

haakonflatval-cognite: Sure - `AnnotationAssetRef` is among the generated types: https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/api/annotations/types.gen.ts

andersfylling: oh! haha

andersfylling:
Any reason we are using stable and not beta package here? Should our docs be updated instead?

haakonflatval-cognite:
> Looks like the service-contracts says this is in beta, but you've published it in stable. I think these should match.
> 
> https://github.com/cognitedata/service-contracts/blob/bf4b7514f921c74978d3c2cce3b7db90b664c8e4/versions/v1/annotations.preproc.yml#LL205C16-L205C16

I don't understand how I'm supposed to implement this in Beta - I can't see any other endpoint being implemented in the `sdk-beta` package. Is there a guide for this, or should I just try to copy the structure from the stable package?

packages/stable/src/api/annotations/annotationsApi.ts
haakonflatval-cognite: This part of the endpoint breaks with the convention used in the rest of the SDK, so I had to use more lower-level components.. But I think it's done now, although I'd rather have it the way it was, to be honest

packages/stable/src/types.ts
haakonflatval-cognite: Fixed :+1:

haakonflatval-cognite:
@andersfylling Thanks for reviewing - I've addressed the comment. But I may need some guidance to understand how to put this in beta.

andersfylling:
> ... or should I just try to copy the structure from the stable package?

Yes, more or less. I think creating:

packages/beta/src/api/annotations/annotationsApi.ts
```ts
export class BetaAnnotationsAPI extends AnnotationsAPI {


  /**
   * [Reverse lookup](https://developer.cognite.com/api/v1-beta/#tag/Annotations/operation/annotationsReverseLookup)
   * const assetQueryData = {
   *   limit: -1,
   *   filter: {
   *     annotatedResourceType: 'file',
   *     annotationType: 'images.AssetLink',
   *     data: {
   *       { id: 123 }
   *     }
   *   }
   * };
   *
   * const resourceIdsResponse = this._client.annotations.reverseLookup(assetQuerydata);
   */
  public reverseLookup = (
    filter: AnnotationReverseLookupRequest
  ): CursorAndAsyncIterator<AnnotationsAssetRef> => {
    const path = this.url(`reverselookup`);
    return this.cursorBasedEndpoint(
      (params) =>
        this.post<CursorResponse<AnnotationsAssetRef[]>>(path, {
          data: params,
        }),
      filter
    );
  };
}
```

and then move the types `AnnotationReverseLookupRequest`, `AnnotationReverseLookupFilter`, `AnnotationReverseLookupFilterProps` over to the beta package, would suffice.

Codegen (as in the *.gen.ts files) isn't supported for alpha, beta, so this shouldn't be generate conflicting types.


Perhaps adding a quick unit test to verify this would be nice.

packages/stable/src/api/annotations/annotationsApi.ts
andersfylling: ah, right! If you decide that users can get by without that field, i won't require that you add it. it's certainly difficult when you also want to keep the cursor and async iterator

It seems the annotatedResourceType is specified in the request structure, so perhaps it's just a helpful field, but not strictly important?

haakonflatval-cognite:
> Perhaps adding a quick unit test to verify this would be nice.

@andersfylling I've made an attempt at moving the endpoint and the test to beta. I now get the error (in the test) that `reverseLookup` cannot be found on the `client.annotations` object, which still refers to the stable version of the API. Am I expected to merge the stable and beta `annotations` objects somehow?

packages/stable/src/api/annotations/annotationsApi.ts
haakonflatval-cognite: Yes, to be honest I don't understand why that field was added in the response, since it will, to my knowledge, always be the same as the one in the request :sweat: So yes, maybe I'll just revert :thinking: Makes it much easier to maintain

andersfylling:
See slack message.

I'm realizing, a bit late, that it doesn't seem that we have any intuitive way of extending the beta package. At least in my eyes. @BugGambit is there a proper way to do this?

andersfylling (approved):

andersfylling:
did you figure out how to deal with the beta header?

haakonflatval-cognite:
> did you figure out how to deal with the beta header?

No, I haven't investigated it further.. I have been struggling with getting my one unit test to pass :see_no_evil: I have reached out to the Matchmaker team to look into it. 

By the way - the endpoint is currently in use in Data Explorer next-release, and will be going public within a week. (It's used for asset search in 360 image annotations). The Cognite Client in Data Explorer is not set up to use beta headers, but there seems to be an error in the annotations API which makes the endpoint available also without a beta header. I've told them not to fix the issue before we have figured out how to support beta headers in the client properly :sweat_smile:

PR #1011, opened 2023-06-09
feat: change default autopage limit to infinity

haakonflatval-cognite:
In my opinion this is a much more sensible limit, as the point of autopage is to easily retrieve all results from a query where the query max limit is exceeded. This would at least have saved me quite some time of debugging, and likely others as well.

BREAKING CHANGE: Changes default limit in `autoPageToArray`

### DIFF
--- a/packages/core/src/autoPagination.ts
+++ b/packages/core/src/autoPagination.ts
@@ -81,7 +81,7 @@ function makeAutoPagingEach<T>(
 
 function makeAutoPagingToArray<T>(autoPagingEach: AutoPagingEach<T>) {
   return async function autoPagingToArray(options?: AutoPagingToArrayOptions) {
-    let limit = (options && options.limit) || 25;
+    let limit = (options && options.limit) || Infinity;
     if (limit === -1) {
       limit = Infinity;
     }


BugGambit (commented):
Since it is a breaking change, then we should bump major version.
Not sure if I like that Infinity is the default. The user can accidentally download a lot of data and. trigger many requests to our API without knowing about it.

bdimitrijoski:
Why Infinity? Can you describe the problem and why this change? Can't you set this optionally in your sdk client?

haakonflatval-cognite:
> Why Infinity? Can you describe the problem and why this change? Can't you set this optionally in your sdk client?

The reason I want infinity, is that the main use case (to my knowledge) of `autoPageToArray` is to conveniently get all results for a query, although each request is typically capped at 1000 result items. In my case, it lead to a lot of unnecessary time in debugging. I had already set the limit to 1000 in the query, but forgot to set the limit in `autoPageToArray`. Of course this can be solved through setting the limit when calling this function, but I think this change makes using our SDK much more intuitive, as this is the limit one would want for almost all cases.

haakonflatval-cognite:
> Since it is a breaking change, then we should bump major version. Not sure if I like that Infinity is the default. The user can accidentally download a lot of data and. trigger many requests to our API without knowing about it.

I agree. This is a breaking change.

It is a valid concern that some users may download a lot of data unintentionally. However, I think it is much more likely that a user of this function will end up with far fewer results than they wanted, and that it makes more sense optimizing for that case.

haakonflatval-cognite:
If it's a better alternative, I'd be content with making `options.limit` a required argument.

haakonflatval-cognite:
> If it's a better alternative, I'd be content with making options.limit a required argument.

@bdimitrijoski @BugGambit Thoughts on this? I think this could be a good middle ground with the least probability for causing bugs, although it will probably require migration in many applications

BugGambit:
@haakonflatval-cognite I think making `limit` required could help. It's not required in the API, so it will deviate from the API.
It would then make sense to accept the values `-1` and `Infinity` to represent "give me everything".

PR #1017, opened 2023-09-18
feat: update openapi spec to bump types fo docs, annotations and vision

danlevings:
Our openapi spec has changed some key values, so the regex needed to be updated
Then rerun the commands to do codegen to update the types


packages/codegen/src/codegen.ts
```diff
@@ -28,7 +28,7 @@ export type StringFilter = (str: string) => boolean;
 export const passThroughFilter: StringFilter = (): boolean => true;
 
 const createServiceNameRegex = (name: string): RegExp => {
-  return new RegExp(`^/api/.+/projects/{project}/${name}($|/)`);
+  return new RegExp(`^/${name}($|/)`);
```
andersfylling, line 31: nice!

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -272,7 +302,7 @@ export interface DocumentGeoJsonGeometry {
      * Coordinates of the geometry.
      * @example [10.74609,59.91273]
      */
-    coordinates?: any[];
```
andersfylling, line 275: Looks like this is a breaking change. but then again, you should never be able to put anything but numbers in there I believe. Perhaps a patch is valid for this case.

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -377,11 +429,10 @@ export interface DocumentsAggregateCardinalityValuesItem {
  */
 export type DocumentsAggregateCardinalityValuesRequest = DocumentSearchInAggregate & DocumentSearchFilter & {
     aggregate: "cardinalityValues";
+    aggregateFilter?: DocumentAggregateFilter;
     properties: {
         property: DocumentFilterProperty;
     }[];
-    limit?: number;
-    cursor?: string;
```
andersfylling, line 384: @tommythorsen where these removed in our service?

andersfylling:
I think changes related to vision and documents are backwards compatible, waiting on tommy to confirm. But annotations look a little suspicious, I wonder if this will break any frontend code? Perhaps we have to do a major release for this.

cc; @olacognite @marvrez

andersfylling:
~Seems regex is failing for the old playground snapshot, where there's still complete uri's. I wonder if it's enough to update the playground snapshot file by replacing `/api/.+/projects/{project}/` with `/`. Just to keep the code simple.~

Let's disable code generation in playground and keep the generated types there. Playground won't have any future changes anyways. Just delete `packages/playground/src/.cognite-openapi-snapshot.json` and `packages/playground/src/codegen.json`, and adjust the github workflow to not run for playground. I think that should be enough.

Edit: did not mean to close ^^,

packages/stable/src/api/documents/types.gen.ts
tommythorsen: They were in the contract, but were not supported by the API, so I removed them from the contract

danlevings: So this is an expected change, and ok to remove?

packages/stable/src/api/documents/types.gen.ts
danlevings: Interesting one. I'd be fine to take the risk here. Like you say - anything but a number would fail anyway

markdownlint.json
```diff
@@ -1,24 +1,29 @@
...
     },
     {
-      "pattern": "^https:\/\/localhost.+"
+      "pattern": "^https://localhost.+"
     },
     {
       "pattern": "^mailto:.+"
+    },
+    {
+      "pattern": "^https://doc.cognitedata.com+"
```
danlevings: https://cognitedata.slack.com/archives/CN9QYQRSR/p1695039928231879
This is a bit sad. The URL initially returns a 404 now, before loading the correct page. I can't think of a reasonable fix for this right now, aside from ignoring these :(

danlevings:
Note: 

> I think changes related to vision and documents are backwards compatible, waiting on tommy to confirm. But annotations look a little suspicious, I wonder if this will break any frontend code? Perhaps we have to do a major release for this.
> 
> cc; @olacognite @marvrez

Looking at the changes, I think the safest bet would be to make this a major breaking change regardless

markdownlint.json
danlevings: nevermind!

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
```diff
@@ -126,32 +126,7 @@ describe('Documents unit test', () => {
...
-                      ],
-                      [
-                        [30.0, 20.0],
-                        [20.0, 15.0],
-                        [20.0, 25.0],
-                        [30.0, 20.0],
-                      ],
-                    ],
-                  ],
+                  coordinates: [10, 20],
```
danlevings: The types no longer support this format. Is this expected? @tommythorsen

tommythorsen: No, this is not quite right. We probably screwed up in this PR https://github.com/cognitedata/service-contracts/pull/2208. This is probably blocking this PR, but could you just comment out the test instead of changing it?

packages/codegen/src/__tests__/codegen.uint.spec.ts
```diff
@@ -95,7 +94,7 @@ describe('code generation', () => {
           }, {});
       };
 
-      const base = '/api/v1/projects/{project}';
```
danlevings, line 98: the generated URLs from OpenAPI has changed. Assuming this is expected, have adjusted the base URLs of codegen to account for this

packages/stable/src/__tests__/api/visionApi.int.spec.ts
```diff
@@ -13,7 +13,7 @@ describe('Vision API', () => {
   let extractBetaJob: VisionExtractPostResponse;
 
   beforeAll(async () => {
-    jest.setTimeout(2 * 60 * 1000); // timeout after 2 minutes
+    jest.setTimeout(3 * 60 * 1000); // timeout after 2 minutes
```
danlevings:
```suggestion
    jest.setTimeout(3 * 60 * 1000); // timeout after 3 minutes
```

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
danlevings: Done!
It indeed blocks this PR. When do you think we could look at fixing the service contract?

packages/stable/src/api/documents/types.gen.ts
tommythorsen: yes

packages/stable/src/__tests__/api/documentsApi.unit.spec.ts
danlevings: This would currently incur a breaking change it seems. which is fine if it is, but it sounds like its unintentional. so we'd need to fix up the service contracts and redo the openapi import before
image

packages/stable/src/api/documents/types.gen.ts
danlevings: This is infact a breaking change according to the tests.
I think we could just revert this back for now and avoid any breaking changes in this PR

packages/stable/src/codegen.json
```diff
@@ -1,5 +1,5 @@
 {
   "snapshot": {
-    "version": "v1"
+    "version": "20230101"
```
danlevings, line 3: Looks like we no longer update v1.json, we now use the new versioning strategy

packages/stable/src/__tests__/api/logout.int.spec.ts
```diff
@@ -1,31 +0,0 @@
-// Copyright 2020 Cognite AS
```
stefanmaretic: Why is this file removed?

danlevings: the logout endpoint isn't a thing in our api anymore.

stefanmaretic (approved):

andersfylling (approved):
Awesome! Thank you for updating this!

Remember to add the `BREAKING CHANGE: description` as you are merging this, otherwise we will have to backport a revert.

packages/stable/src/api/vision/types.gen.ts
```diff
@@ -56,16 +66,46 @@ export interface AnnotationsCogniteAnnotationTypesImagesAssetLink {
...
+}
+/**
+ * The numerical value of something
+ */
+export interface AnnotationsNumerical {
+    /** The description of a primitive */
+    description?: string;
+    type: "numerical";
+    /** The numerical value */
+    value: number | number;
```
marvrez, line 100: `number | number` -- interesting 😅 . Since TSC is not complaining, then I guess this is fine

marvrez (approved):
LGTM!

packages/stable/src/api/vision/types.gen.ts
andersfylling: Maybe you want to update your openapi spec:
image

PR #1019, opened 2023-09-18
feat: user profiles api

danlevings:
Introduce the user profiles API


packages/stable/src/api/userProfiles/profilesApi.ts
```diff
@@ -0,0 +1,63 @@
...
+   * [List all user profiles](https://api-docs.cognite.com/20230101/tag/User-profiles/paths/~1profiles/get)
+   *
+   * ```js
+   * const response = await client.profiles.list({ limit: 1000, nextCursor: 'abc' });
+   * ```
+   */
+  public list = (scope: {
+    limit: number;
+    cursor: string;
+  }): CursorAndAsyncIterator<UserProfileItem> => {
```
andersfylling: is `UserProfilesListResponse` a better fit?

danlevings: yes! thanks!

andersfylling:
I would recommend using the generated request and response objects in your endpoint implementations. Then the endpoints are also updated whenever your request/response objects are updated as well.

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  describe('client.profiles.list', () => {
+    test('fetch list of users', async () => {
+      const profile = await client.profiles.list();
+      expect(profile).toBeDefined();
+    });
+
+    test('fetch list of users with limit = 2 + cursors', async () => {
+      const profile = await client.profiles.list({ limit: 2 });
```
stefanmaretic: I guess this should be `profiles`?

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+      expect(profile.items.length).toBe(2);
+
+      const nextProfiles = await client.profiles.list({
+        limit: 2,
+        cursor: profile.nextCursor,
+      });
+      expect(nextProfiles.items.length).toBe(2);
```
stefanmaretic: Should we test here for `expect(nextProfiles).not.toEqual(profiles)`?

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+  });
```
stefanmaretic: Is this just the exact same test as the one above?

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  describe('client.profiles.search', () => {
+    test('fetch list of users', async () => {
+      const me = await client.profiles.me();
+
+      const searchForMe = await client.profiles.search({
+        search: { name: me.displayName || '' },
+      });
+      console.log(searchForMe);
```
stefanmaretic:
```suggestion
```

packages/stable/src/api/userProfiles/profilesApi.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright 2022 Cognite AS
```
stefanmaretic: Not sure whether the year is relevant here, but 2023? 😬

stefanmaretic (commented):
I don't think any of my comments are blockers, but would be nice to resolve them before merging.

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
gulkaktus: fixed!

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
gulkaktus: fixed!

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
gulkaktus: removed!

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
gulkaktus: Removed

packages/stable/src/api/userProfiles/profilesApi.ts
gulkaktus: fixed!

gulkaktus:
Tests are failing because user ids are currently not generated for service accounts. This is expected to be fixed around October 1st - 5th

packages/stable/src/api/userProfiles/profilesApi.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  public list = (scope?: {
+    limit?: number;
+    cursor?: string;
+  }): UserProfilesListResponse => {
+    return super.listEndpoint(this.callListEndpointWithGet, scope);
+  };
+
+  /**
+   * [Retrieve one or more user profiles by ID](https://api-docs.cognite.com/20230101/tag/User-profiles/paths/~1profiles~1byids/post)
```
haakonvt: I dont read js, but just a heads up that byids doesn't return in order. Here's how the Python SDK manually sorts results before returning:
https://github.com/cognitedata/cognite-sdk-python/blob/f4c00fe6fcae3fc9d821e3e9ca5444ac0e11d9a9/cognite/client/_api/user_profiles.py#L81

gulkaktus: Thanks for the heads up! Should probably be updated at some point, but I'll leave this as is since it's how we're currently treating all the retrieve methods.

packages/stable/src/__tests__/api/userProfiles.nit.spec.ts
polomani, line 1: .unit.spec.ts (misspelled)

gulkaktus: good catch, fixed!

packages/stable/src/__tests__/api/userProfiles.unit.spec.ts
polomani, line 1: looking at this again, those are integration tests, not unit, right?

packages/stable/src/api/userProfiles/profilesApi.ts
```diff
@@ -0,0 +1,64 @@
...
+   * const response = await client.profiles.retrieve([{ userIdentifier: 'abcd' }]);
+   * ```
+   */
+  public retrieve = async (
+    userIdentifiers: UserIdentifier[]
+  ): Promise<UserProfileItem[]> => {
+    return super.retrieveEndpoint(userIdentifiers);
+  };
+  /**
+   * [Search user profiles](https://api-docs.cognite.com/20230101/tag/User-profiles/paths/~1profiles~1search/post)
```
polomani: is there a chance to make better links for this, e.g. by giving an operation name?
e.g. sessions: `https://api-docs.cognite.com/20230101/tag/Sessions/operation/listSessions`
compare to `https://api-docs.cognite.com/20230101/tag/User-profiles/paths/~1profiles~1search/post`

polomani: the reason I ask, is that they won't appear as code snippets in the portal otherwise, example:

Screenshot 2023-11-28 at 16 08 15

gulkaktus: Good point, I guess I should modify this file? https://github.com/cognitedata/service-contracts/blob/master/versions/v1/user_profiles.yml

The current link is being referenced by others, e.g. the Python SDK. What's the best way of communicating to everyone that the link will break? @polomani

packages/stable/src/__tests__/api/userProfiles.unit.spec.ts
gulkaktus: Yes, fixed!

packages/stable/src/api/userProfiles/profilesApi.ts
polomani: it might not break, as there is a chance that both will work no? All other resources use operationids anyway, it's better to be consistent.
Also it should be possible to update the link in python sdk

gulkaktus: Made a PR: https://github.com/cognitedata/service-contracts/pull/2388
Awaiting review from the API team

gulkaktus: @polomani above PR merged and urls updated 😃

packages/stable/src/api/userProfiles/profilesApi.ts
```diff
@@ -0,0 +1,64 @@
...
+   * ```js
+   * const response = await client.profiles.list({ limit: 1000, nextCursor: 'abc' });
+   * ```
+   */
+  public list = (scope?: {
+    limit?: number;
+    cursor?: string;
+  }): UserProfilesListResponse => {
+    return super.listEndpoint(this.callListEndpointWithGet, scope);
+  };
```
polomani: you could reuse the common type
```suggestion
  public list = (scope?: FilterQuery): UserProfilesListResponse => {
    return this.listEndpoint(this.callListEndpointWithGet, scope);
  };
```

packages/stable/src/api/userProfiles/profilesApi.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const response = await this.get<UserProfileItem>(path);
+    return this.addToMapAndReturn(response.data, response);
+  };
+
+  /**
+   * [List all user profiles](https://developer.cognite.com/api#tag/User-profiles/operation/listUserProfiles)
+   *
+   * ```js
+   * const response = await client.profiles.list({ limit: 1000, nextCursor: 'abc' });
```
polomani: this snippet is incorrect
Screenshot 2023-11-30 at 16 28 59

packages/stable/src/api/userProfiles/profilesApi.ts
gulkaktus: ah nice

packages/stable/src/api/userProfiles/profilesApi.ts
gulkaktus: Fixed!

packages/stable/src/__tests__/api/userProfiles.int.spec.ts
```diff
@@ -0,0 +1,47 @@
+// Copyright 2023 Cognite AS
+
+import CogniteClient from '../../cogniteClient';
+import { setupLoggedInClient } from '../testUtils';
+
+describe('User Profiles unit test', () => {
```
polomani:
```suggestion
describe('User Profiles', () => {
```

packages/stable/src/exports.gen.ts
```diff
@@ -116,6 +116,20 @@ export {
...
+  CursorQueryParameter,
+  IdentityType,
+  IdentityTypeFilter,
+  UserIdentifier,
+  UserProfileItem,
+  UserProfilesByIdsRequest,
+  UserProfilesByIdsResponse,
+  UserProfilesErrorResponse,
+  UserProfilesListResponse,
+  UserProfilesNotFoundResponse,
```
polomani, line 130: it seems like some of these types aren't needed/used, @andersfylling is there a setting to omit them somehow?

andersfylling: Hm, I don't think we added support for that. Question here is why aren't they using an ErrorResponse type in the open api spec?

polomani (approved):
LGTM, please squash when merging (and keep "feat: user profiles api" as the commit name)

PR #1027, opened 2023-09-29
feat(v1): add units catalog and timeseries unit conversion

polomani:
- units (list, retrieve)
- unit systems (list)
- timeseries APIs

I also had to fix broken code snippets in other places (e.g. Annotations) as those were not captured before


packages/beta/src/api/units/unitsApi.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright 2022 Cognite AS
```
polomani:
```suggestion
// Copyright 2023 Cognite AS
```

packages/beta/src/api/units/unitSystemsApi.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2022 Cognite AS
```
polomani:
```suggestion
// Copyright 2023 Cognite AS
```

packages/stable/src/__tests__/api/annotationsApi.int.spec.ts
```diff
@@ -226,7 +226,7 @@ describe('Annotations API', () => {
     expect(items).toHaveLength(1);
   });
 
-  test('reverse lookup annotation', async () => {
+  test.skip('reverse lookup annotation', async () => {
```
polomani, line 229: this often fails, we ignore it in python SDK as well

lpereiracgn (approved):

PR #1034, opened 2023-11-06
feat(monthly aggregate): retrieve datapoints where granularity set as month

anjalirai2626:
[Monthly aggregates through JS SDK](https://cognitedata.atlassian.net/browse/UX-1720)

### DIFF
--- a/packages/stable/src/api/dataPoints/dataPointsApi.ts
+++ b/packages/stable/src/api/dataPoints/dataPointsApi.ts
@@ -60,7 +60,7 @@ export class DataPointsAPI extends BaseResourceAPI<
    */
   public retrieveDatapointMonthlyAggregates = async (
     query: DatapointsMonthlyGranularityMultiQuery
-  ): Promise<DatapointAggregates[] | Datapoints[]> => {
+  ): Promise<DatapointAggregates[]> => {
     // Find the start and end dates from the query
     const startDate = query.start;
     const endDate = query.end;
@@ -84,7 +84,6 @@ export class DataPointsAPI extends BaseResourceAPI<
       });
 
       // Call the API for each month in parallel and save it in a variable
-      // eslint-disable-next-line @typescript-eslint/no-unused-vars
       const results = await Promise.all(promises);
 
       // Merge the datapoints into a single array
@@ -107,7 +106,7 @@ export class DataPointsAPI extends BaseResourceAPI<
       return [firstItem];
     }
 
-    return this.retrieveDatapointsEndpoint(query);
+    return this.retrieveDatapointsEndpoint<DatapointAggregates[]>(query);
   };
 
   /**
@@ -150,11 +149,13 @@ export class DataPointsAPI extends BaseResourceAPI<
     return {};
   }
 
-  private async retrieveDatapointsEndpoint(query: DatapointsMultiQuery) {
+  private async retrieveDatapointsEndpoint<
+    T extends DatapointAggregates[] | Datapoints[] =
+      | DatapointAggregates[]
+      | Datapoints[]
+  >(query: DatapointsMultiQuery) {
     const path = this.listPostUrl;
-    const response = await this.post<
-      ItemsWrapper<DatapointAggregates[] | Datapoints[]>
-    >(path, {
+    const response = await this.post<ItemsWrapper<T>>(path, {
       data: query,
     });
     return this.addToMapAndReturn(response.data.items, response);
--- a/packages/stable/src/api/dataPoints/dataPointsApi.ts
+++ b/packages/stable/src/api/dataPoints/dataPointsApi.ts
@@ -56,10 +56,10 @@ export class DataPointsAPI extends BaseResourceAPI<
    * [Retrieve data points](https://doc.cognitedata.com/api/v1/#operation/getMultiTimeSeriesDatapoints)
    *
    * ```js
-   * const data = await client.datapoints.retrieveMonthlyGranularity({ items: [{ id: 123 }] });
+   * const monthlyAggregatesData = await client.datapoints.retrieveDatapointMonthlyAggregates({ items: [{ id: 123 }] });
    * ```
    */
-  public retrieveMonthlyGranularity = async (
+  public retrieveDatapointMonthlyAggregates = async (
     query: DatapointsMonthlyGranularityMultiQuery
   ): Promise<DatapointAggregates[] | Datapoints[]> => {
     // Find the start and end dates from the query
@@ -93,7 +93,7 @@ export class DataPointsAPI extends BaseResourceAPI<
 
       for (const result of results) {
         // Check if the array contains data before pushing it to the merged datapoints array
-        if (Array.isArray(result) && result.length === 1) {
+        if (result?.length && result[0]?.datapoints) {
           const datapoints = result[0].datapoints;
           // Merge the datapoints into the result array
           mergedDatapoints.push(...datapoints);
--- a/packages/stable/src/api/dataPoints/dataPointsApi.ts
+++ b/packages/stable/src/api/dataPoints/dataPointsApi.ts
@@ -53,7 +53,7 @@ export class DataPointsAPI extends BaseResourceAPI<
   };
 
   /**
-   * [Retrieve data points](https://doc.cognitedata.com/api/v1/#operation/getMultiTimeSeriesDatapoints)
+   * [Retrieve data points](https://doc.cognitedata.com/api/v1/#operation/getMultiTimeSeriesDatapointsMonthly)
    *
    * ```js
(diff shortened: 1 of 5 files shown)


bdimitrijoski (commented):
Looks ok to me

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +51,58 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+      const results = Promise.all(promises);
+
+      // Format the data in Promise<DatapointAggregates[] | Datapoints[]>
+      return results.then((data) => {
+        // Flatten the array of arrays
+        const flattenedData = data.flat();
+
+        // Return the data
+        return flattenedData;
+      });
```
BugGambit: Prefer `async/await`?

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -121,4 +174,36 @@ export class DataPointsAPI extends BaseResourceAPI<
   }
 }
 
+interface MonthInfo {
+  start_date: Timestamp;
+  end_date: Timestamp;
+  number_of_days: number;
```
BugGambit: Use camelCase

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +51,58 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+    if (startDate && endDate) {
+      // Get the months between the start and end dates
+      const months = getMonthsBetweenDates(startDate, endDate);
+
+      // Create a array of promises for each month
+      const promises = months.map(month => {
+        // Create a new query for each month
+        const newQuery = {
+          ...query,
+          start: month.start_date,
```
bdimitrijoski: I think you will need to pass the aggregate function here as well

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -121,4 +174,36 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+    const lastDay = new Date(year, month, 0);
+
+    result.push({
+      start_date: firstDay.getTime(),
+      end_date: lastDay.getTime(),
+      number_of_days: lastDay.getDate() - firstDay.getDate() + 1,
+    });
+
+    // Move to the next month
+    currentMonth.setMonth(currentMonth.getMonth() + 1);
```
BugGambit, line 214: This works between years? e.g. Dec 2022 will give you Jan 2023?

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -121,4 +174,36 @@ export class DataPointsAPI extends BaseResourceAPI<
   }
 }
 
+interface MonthInfo {
+  start_date: Timestamp;
+  end_date: Timestamp;
+  number_of_days: number;
+}
+
+const getMonthsBetweenDates = (startDate: Timestamp | string, endDate: Timestamp | string): MonthInfo[] => {
```
bdimitrijoski: Maybe unit tests for this file would be nice

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +51,58 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+   */
+  public retrieveMonthlyGranularity = (
+    query: Omit<DatapointsMultiQuery, "granularity">
+  ): Promise<DatapointAggregates[] | Datapoints[]> => {
+
+    // Find the start and end dates from the query
+    const startDate = query.start;
+    const endDate = query.end;
+
+    if (startDate && endDate) {
```
BugGambit, line 68: What if one of them is undefined?

packages/stable/src/api/dataPoints/dataPointsApi.ts
anjalirai2626: It's expected to be present in the initial [query](https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/types.ts#L800) though it is optional. Do you mean it needs to be mandatory?

packages/stable/src/api/dataPoints/dataPointsApi.ts
anjalirai2626: It seems working, I'll write a test case for it

packages/stable/src/api/dataPoints/dataPointsApi.ts
anjalirai2626: handled in default/API [here](https://github.com/cognitedata/cognite-sdk-js/pull/1034/files#diff-1a7fec289c75d6888825dadca5e4fff169305d1c7459d4ccbd1fe5250b389d87R111)

packages/stable/src/api/dataPoints/dataPointsApi.ts
anjalirai2626: works as expected, tested [here](https://github.com/cognitedata/cognite-sdk-js/pull/1034/files#diff-04fc6dab9e7c7cdac19f22741d90516fa63d141ad8a1a57d5e6a18ae3cd976bcR181)

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +52,65 @@ export class DataPointsAPI extends BaseResourceAPI<
     return this.retrieveDatapointsEndpoint(query);
   };
 
+  /**
+   * [Retrieve data points](https://doc.cognitedata.com/api/v1/#operation/getMultiTimeSeriesDatapoints)
+   *
+   * ```js
+   * const data = await client.datapoints.retrieve({ items: [{ id: 123 }] });
```
polomani:
```suggestion
   * const data = await client.datapoints.retrieveMonthlyGranularity({ items: [{ id: 123 }] });
```
typedoc is wrong

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +52,65 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+   * ```
+   */
+  public retrieveMonthlyGranularity = async (
+    query: DatapointsMonthlyGranularityMultiQuery
+  ): Promise<DatapointAggregates[] | Datapoints[]> => {
```
polomani:
```suggestion
  ): Promise<DatapointAggregates[]> => {
```
presumably?

anjalirai2626: Yes, I'm always formatting the return data to the `DatapointAggregates` type however I think we can keep it this way because the default return type could be one of both [here](https://github.com/cognitedata/cognite-sdk-js/pull/1034/files#diff-1a7fec289c75d6888825dadca5e4fff169305d1c7459d4ccbd1fe5250b389d87R111) otherwise I'll need to custom return type in `retrieveDatapointsEndpoint` for monthly aggregation

samir-ajdarpasic (approved):

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +52,65 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+      // Call the API for each month in parallel and save it in a variable
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      const results = await Promise.all(promises);
+
+      // Merge the datapoints into a single array
+      const mergedDatapoints = [];
+
+      for (const result of results) {
+        // Check if the array contains data before pushing it to the merged datapoints array
+        if (Array.isArray(result) && result.length === 1) {
```
BugGambit: Is there a case when result isn't an array?

anjalirai2626: good point, it will always be one of these `DatapointAggregates[] | Datapoints[]` it could be just `result?.length`

packages/stable/src/api/dataPoints/dataPointsApi.ts
polomani: you can cast the type before returning it (since you know it anyway)

packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -50,6 +52,64 @@ export class DataPointsAPI extends BaseResourceAPI<
...
+          end: month.endDate,
+          granularity: `${month.numberOfDays}d`,
+        };
+
+        // Return a promise for each month
+        return this.retrieveDatapointsEndpoint(newQuery);
+      });
+
+      // Call the API for each month in parallel and save it in a variable
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
polomani:
```suggestion
```
it works without it, no?

PR #1040, opened 2023-11-23
fix: use UTC timezone regardless of client time zone

muradsater:
Always do queries in UTC timezone regardless of input timezone. This avoid getting the monthly aggregates shifted to the previous day in the case of time zones which is after UTC.

In Oslo 2023-01-01:00:00:00 is 2022-12-31T23:00:00 in UTC. (which gave aggregates from 31/12->30/1 instead of 1/1->31/1)


anjalirai2626 (approved):
Lgtm!

PR #1041, opened 2023-11-24
feat: add options field to simulator steps and name to simulator DM

lpereiracgn:


packages/alpha/src/types.ts
```diff
@@ -108,6 +115,7 @@ export interface SimulatorIntegration {
 export interface SimulatorIntegrationCreate {
   externalId?: CogniteExternalId;
   simulatorExternalId?: CogniteExternalId;
+  name?: string;
```
```suggestion
```
SimulatorIntegrationCreate doesn't have a name,SimulatorCreate does

polomani (approved):

PR #1042, opened 2023-11-27
feat: monitoring tasks upsert

23andreas:
- Adds monitoring tasks upsert - [docs](https://pr-60.tasks-api.preview.cogniteapp.com/#tag/Monitoring/operation/upsert_monitoring_tasks_api_v1_projects__project__monitoringtasks_upsert_post)
- Updates MonitoringTaskCreate interface, make `interval` and `overlap` optional to match api 

These changes are primarily in the beta package, but I have [added a new endpoint to core](https://github.com/cognitedata/cognite-sdk-js/pull/1042/files#diff-393f42bfefc6d77b843d57c6cec2e9b4f2315f679029b72b71c2bb9c5fb026f7) and a [new type to stable](https://github.com/cognitedata/cognite-sdk-js/pull/1042/files#diff-a1cb34e0ba200ad9a54dfae19192d251fe8573729eea371b18dacdbf2850ab44R2383)

### DIFF
--- a/packages/beta/src/__tests__/api/monitoringTasksApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/monitoringTasksApi.int.spec.ts
@@ -16,7 +16,7 @@ describe('monitoring tasks api', () => {
   const client: CogniteClient = setupLoggedInClient();
   const ts = Date.now();
   const monitoringTaskExternalId = `test_mt_${ts}`;
-  const monitoringTaskName = `test_mt_${ts}`;
+  let monitoringTaskName = `test_mt_${ts}`;
   const channelExternalId = `test_channel_mt_${ts}`;
   const sessionsApi = `/api/v1/projects/${process.env.COGNITE_PROJECT}/sessions`;
   const testMtModel: MonitoringTaskDoubleThresholdModelCreate = {
@@ -92,6 +92,33 @@ describe('monitoring tasks api', () => {
     expect(response[0].externalId).toBe(monitoringTaskExternalId);
     done();
   }, 10000);
+
+  test('upsert monitoring task', async () => {
+    const sessionsRes = await client.post<SessionsResponse>(sessionsApi, {
+      data: {
+        items: [
+          {
+            clientId: process.env.COGNITE_CLIENT_ID,
+            clientSecret: process.env.COGNITE_CLIENT_SECRET,
+          },
+        ],
+      },
+    });
+
+    monitoringTaskName = monitoringTaskName + '_updated';
+    const response = await client.monitoringTasks.upsert([
+      {
+        externalId: monitoringTaskExternalId,
+        name: monitoringTaskName,
+        channelId: channel.id,
+        model: testMtModel,
+        nonce: sessionsRes?.data?.items[0]?.nonce,
+      }
+    ]);
+
+    expect(response.length).toBe(1);
+    expect(response[name]).toBe(monitoringTaskName);
+  });
 
   test('list all monitoring tasks', async () => {
     const response = await client.monitoringTasks.list({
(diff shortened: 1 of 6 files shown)


tuanng-cognite (approved):
LGTM, Assuming that most stuffs are generated.

23andreas:
> LGTM, Assuming that most stuffs are generated.

generated? Is there a way to generate types etc from the api?

tuanng-cognite:
> > LGTM, Assuming that most stuffs are generated.
> 
> generated? Is there a way to generate types etc from the api?

I mixed it with other sdk. There are no tools to generate SDK code for this. Sorry

packages/beta/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -93,6 +93,33 @@ describe('monitoring tasks api', () => {
...
+        items: [
+          {
+            clientId: process.env.COGNITE_CLIENT_ID,
+            clientSecret: process.env.COGNITE_CLIENT_SECRET,
+          },
+        ],
+      },
+    });
+
+    monitoringTaskName = monitoringTaskName + '_updated';
```
saagar-cognite: suggestion(blocking): could we please use const declaration for updated name just like 
`const updatedName = channelExternalId + '_updated';` ?

23andreas: https://github.com/cognitedata/cognite-sdk-js/pull/1042/commits/4a094746b52a0ab74aefc82dcae9ec0b59bcdf1d

packages/stable/src/types.ts
```diff
@@ -2380,6 +2380,7 @@ export type SingleCogniteCapability =
   | { templateInstancesAcl: AclTemplateInstances };
 
 export type SinglePatch<T> = { set: T } | { setNull: boolean };
+export type SinglePatchRequired<T> = { set: T };
```
polomani, line 2383: This is a very specific type to alerts api, define it inline in the Channel patch
```
    parentId?: { set: number };
```

23andreas: It's not visible in the changes, but we already have SinglePatchRequiredString defined just below it 
https://github.com/cognitedata/cognite-sdk-js/blob/f408bfbf83efd9268d94e3d9b962cd64e50d35c4/packages/stable/src/types.ts#L2387

So I thought it made sense to have it there for the more generic SinglePatch as well. 
Wdyt? Should I still move it inline?

polomani: right, in that case it's fine

packages/beta/src/__tests__/api/monitoringTasksApi.int.spec.ts
```diff
@@ -93,6 +94,32 @@ describe('monitoring tasks api', () => {
...
+        channelId: channel.id,
+        model: testMtModel,
+        nonce: sessionsRes?.data?.items[0]?.nonce,
+      },
+    ]);
+
+    expect(response.length).toBe(1);
+    expect(response[name]).toBe(monitoringTaskNameUpdated);
+  });
+
```
nabati, line 124: Didn't dive in to this, just for my own curiosity, how are these tests run, are they run against an actual backend or are the responses mocked?

nabati (commented):
LGTM, but difficult for me to assess in isolation

packages/beta/src/__tests__/api/monitoringTasksApi.int.spec.ts
23andreas: They run against actual backend I think

PR #1045, opened 2023-12-18
fix: return all time series when asking for multiple time series of monthly aggregates

muradsater:
The monthly aggregates would only return one time series (the first in the result set) regardless of the number of time series asked for in the query.

It appeared that the queries were correct, but the result parsing only used the first time series in the result.


anjalirai2626 (approved):
LGTM, thanks!

PR #1046, opened 2023-12-22
fix: set empty array for no results

muradsater:


packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -114,7 +114,11 @@ export class DataPointsAPI extends BaseResourceAPI<
 
       for (const key in mergedTimeseries) {
         const item = mergedTimeseries[key];
-        item.datapoints = mergedDatapoints[key];
+        if (key in mergedDatapoints) {
+          item.datapoints = mergedDatapoints[key];
+        } else {
+          item.datapoints = [];
+        }
```
samir-ajdarpasic, line 121: @muradsater as a quick fix this is ok.
I do believe that we should refactor this function a bit in the future for easier readability and performance.

samir-ajdarpasic (approved):

PR #1047, opened 2024-01-11
feat: instances/search and instances/list

gulkaktus:
Add `models/instances/search` and `models/instances/list` endpoints to `stable`.


packages/stable/src/.cognite-openapi-snapshot.json
```diff
@@ -57975,18 +57975,6 @@
                       }
                     ]
                   }
-                },
```
gulkaktus: Should not remove this. Was only done to make type generation script run

packages/stable/src/exports.gen.ts
```diff
@@ -117,6 +117,129 @@ export {
   Label,
   LabelList
 } from './api/documents/types.gen';
+export {
```
andersfylling: This file is owned by the codegen package. So whenever someone runs codegen, these imports will be lost.

Simply moving these out of the exports.gen.ts file will resolve the potential future issue.

andersfylling (changes requested):

packages/stable/src/exports.gen.ts
gulkaktus: Thanks for pointing that out! Moved the types to the types.ts file instead.

andersfylling (approved):
lgtm.

If you make any big chances, besides fixing the test cases, then please request another review.

PR #1049, opened 2024-01-15
fix(docs): added missing baseUrl in auth code to comply with demo code

skepticalcat:
Client instantiation needs baseUrl set - its correctly set in the demo code, but missing in the example docs. Added.

### DIFF
--- a/guides/authentication.md
+++ b/guides/authentication.md
@@ -125,6 +125,7 @@ async function quickstart() {
   const client = new CogniteClient({
     appId: 'Cognite SDK samples',
     project,
+    baseUrl: "https://api.cognitedata.com",
     getToken: () =>
       pca
         .acquireTokenByClientCredential({


guides/authentication.md
```diff
@@ -125,6 +125,7 @@ async function quickstart() {
   const client = new CogniteClient({
     appId: 'Cognite SDK samples',
     project,
+    baseUrl: "https://api.cognitedata.com",
```
andersfylling, line 128: there is a baseUrl variable further up, perhaps reusing that would avoid any unexpected behavior if anyone runs this =)

skepticalcat: You mean the one defined in the **OIDC authentication using code authorization w pkce** snippet?

andersfylling: whops, seems i scrolled a bit too quickly through this

andersfylling (approved):

PR #1050, opened 2024-01-16
type generator error

gulkaktus:
## The problem
The [instances/query](https://api-docs.cognite.com/20230101/tag/Instances/operation/queryContent) endpoint returns a QueryResponse that looks like this:

```
type QueryResponse = {
   items: Record<string, NodeOrEdge>;
   nextCursor: Record<String, String>
}

```

This is non-standard and deviates from the service contract design doc, and causes the type generator to fail
image (8)

Specifically, it seems like it's expecting that the `QueryResponse` node is of type `arrayTypeNode`, but it's considered a `typeReferenceNode`
image

## What we've tried so far
- Create an exception for cases where we have a `typeReferenceNode`. Intuitively, it seems like it should be possible to extract the `items` and `nextCursor` nodes, but we're not sure how. I'm also not sure if this type of `nextCursor`, which is `Record<String, String>` is supported.

## Our current workaround
- Generate the types while omitting `nextCursor`. Add it back manually, then exclude `instances` from further type generation.


packages/codegen/src/ast_transformer/cursor_and_async_iterator.ts
```diff
@@ -65,7 +103,11 @@ const cursorAndAsyncIteratorTransformer: ts.TransformerFactory<
         ts.isInterfaceDeclaration(node) &&
         interfaceDeclarationHasRequiredMembers(node)
       ) {
-        const itemsType = getItemsType(node);
+        console.log(node.name.escapedText);
+        const itemsType = getItemsType(
```
gulkaktus, line 107: Only console log for the offending node

PR #1051, opened 2024-01-17
feat(instances): added retrieve and delete to InstancesAPI class

EliasBjorne:
added retrieve and delete to InstancesAPI class as part of adding data model instances to the cognite js sdk


andeplane:
Should we close this?

PR #1058, opened 2024-02-07
feat: introduce datamodels endpoints + export types correctly

danlevings:
Introduce the datamodels endpoints listed here 
https://api-docs.cognite.com/20230101/tag/Data-models/operation/createDataModels

also ensure we're exporting types correctly from root.


packages/stable/src/api/models/codegen.skip.json
```diff
@@ -0,0 +1,10 @@
+{
+  "_": "Due to issue with codegen duplicating types, we skip this for now.",
```
andersfylling, line 2: Can you expand on this issue?

danlevings: Sure! you can replicate this by renaming it from 'codegen.skip.json' to 'codegen.json' and running `yarn codegen --package=stable generate-types`

The `exports.gen.ts` tries to export these types, but they don't exist inside of the `types.gen` file.
It seems like the code generation is correctly identifying 'shared types', but theyre not correctly being removed from their respective types.gen file and into the common one. e.g. in this case, these types exist in `api/datamodels/types.gen.ts`, when they should be moved into `types.gen`
image

The code generation portion would take someone unfamiliar with the code a while to dig into. I tried, it was a bit tricky to follow. so we decided to generate the types once.

does that make sense?

gulkaktus (approved):
LGTM!

packages/stable/src/api/models/codegen.skip.json
andersfylling: I think that is good way to get around that issue. I'm all for the tool simply helping you to reduce time and typos, otherwise the value is lost.

Ofcourse it should be fixed, but I'll need to get back to this a bit later.

PR #1066, opened 2024-03-07
feat: update simulation run and filter types

bkuzma:
updates `SimulationRun` and `SimulationRunFilter`.

### DIFF
--- a/packages/alpha/src/types.ts
+++ b/packages/alpha/src/types.ts
@@ -145,8 +145,11 @@ export interface SimulationRunFilter {
   modelName?: string;
   routineName?: string;
   status?: SimulationRunStatus;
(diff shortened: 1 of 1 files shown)


packages/alpha/src/types.ts
```diff
@@ -220,6 +225,7 @@ export interface SimulationRun {
   eventId?: CogniteInternalId;
   runType: SimulationRunType;
   userId?: string;
+  logId?: number;
```
```suggestion
  logId?: CogniteInternalId;
```

polomani (approved):
looks good, just a small comment

packages/alpha/src/types.ts
bkuzma: fixed 👍

PR #1070, opened 2024-03-15
feat: update alerts filter

neringaalt:
Adding status, closed and startTime fields for AlertFilter

### DIFF
--- a/packages/beta/src/types.ts
+++ b/packages/beta/src/types.ts
@@ -7,6 +7,7 @@ import {
   SinglePatchRequiredString,
   SinglePatchString,
   Timestamp,
+  DateRange,
 } from '@cognite/sdk';
 import {
   CogniteExternalId,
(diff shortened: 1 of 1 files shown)


packages/beta/src/types.ts
```diff
@@ -145,6 +158,9 @@ export interface MonitoringTaskFilterQuery extends FilterQuery {
 export interface AlertFilter {
   channelIds?: CogniteInternalId[];
   channelExternalIds?: CogniteExternalId[];
+  closed?: boolean;
+  startTime?: StartTime;
```
polomani:
```suggestion
  startTime?: DateRange;
```

packages/beta/src/types.ts
```diff
@@ -57,6 +57,19 @@ export const MonitoringTaskModelExternalId = {
...
+  FIRING: 'FIRING' as AlertStatus,
+  PENDING: 'PENDING' as AlertStatus,
+};
+
+export interface StartTime {
+  min?: number;
+  max?: number;
+
+};
+
```
polomani: no need of this type, use DateRange;
```suggestion
```

neringaalt: ok, thanks! :D just trying my best here :D

tuanng-cognite (approved):
Let's ship this

PR #1071, opened 2024-03-20
feat(datapoints): add support for data quality (beta)

23andreas:
Add support for data quality status codes to beta package

https://developer.cognite.com/dev/concepts/resource_types/timeseries#data-point-quality-status-codes


packages/beta/src/types.ts
```diff
@@ -322,3 +330,72 @@ export interface Subscription {
...
+   * denotes that the API should return bad data points.
+   * Because the API treats uncertain data points as bad by default,
+   * this parameter includes both uncertain and bad data points.
+   */
+  ignoreBadDatapoints?: boolean;
+}
+
+export interface DatapointStatus {
+  code: number;
+  symbol: string;
```
23andreas, line 395: Symbol can be a combination of [Category](https://developer.cognite.com/dev/concepts/reference/quality_codes/#subcategories)
and [modifier flags.](https://developer.cognite.com/dev/concepts/reference/quality_codes/#modifier-flags) 
For example `BadBrowseNameInvalid, StructureChanged, Low`. 

I initially tried typing it using [template literals](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html)
but this scenario was too advanced for that, so went with string. 

Code is the number you get by adding all the category and modifier codes together. 
The example above would result in 2153775104 + 32768 + (1024) + 256 = 2153809152. 
[More details here](https://developer.cognite.com/dev/concepts/reference/quality_codes/#status-code-structure)

packages/beta/src/types.ts
```diff
@@ -322,3 +330,72 @@ export interface Subscription {
...
+}
+
+export interface DatapointAggregate extends DatapointAggregateStable {
+  countUncertain?: number;
+  countBad?: number;
```
polomani, line 406: this is pretty clean! nice

packages/stable/src/__tests__/api/datamodels.int.spec.ts
```diff
@@ -48,8 +48,17 @@ describe('Data models integration test', () => {
...
     client = setupLoggedInClient();
+
+    const spaces = await client.spaces
+      .list({ limit: 1000 })
+      .autoPagingToArray();
+    const oldSpaces = spaces.filter(
+      // Filter spaces last updated more than 24 hours ago
+      (space) => space.lastUpdatedTime < Date.now() - 24 * 60 * 60 * 1000
+    );
+    await client.spaces.delete(oldSpaces.map((space) => space.space));
```
BugGambit: Move to a function to avoid duplicate code? It's currently repeated three times

23andreas: Good suggestion 
will do that

packages/beta/src/__tests__/api/dataPoints.int.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+        id: timeserie.id,
+        datapoints: datapoints,
+      },
+    ]);
+  });
+  afterAll(async () => {
+    await client.timeseries.delete([{ id: timeserie.id }]);
+  });
+
+  test('retrieve aggregates with correct data point status counts', async () => {
```
23andreas, line 93: Test #1 for status codes, rest of the tests are copied from stable

packages/beta/src/__tests__/api/dataPoints.int.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+    expect(response[0].datapoints[3].countGood).toBe(1);
+    expect(response[0].datapoints[3].countUncertain).toBe(0);
+    expect(response[0].datapoints[3].countBad).toBe(1);
+
+    expect(response[0].datapoints[4].countGood).toBe(1);
+    expect(response[0].datapoints[4].countUncertain).toBe(1);
+    expect(response[0].datapoints[4].countBad).toBe(0);
+  });
+
+  test('retrieve datapoints with status code', async () => {
```
23andreas, line 124: test #2

packages/beta/src/__tests__/api/dataPoints.int.spec.ts
```diff
@@ -0,0 +1,611 @@
...
+};
+
+describe('Datapoints integration test', () => {
+  const client = setupLoggedInClient();
+
+  let timeserie: Timeseries;
+  const timestampFrom = Date.now() - 60 * 60 * 48 * 1000;
+  const timestampTill = Date.now() - 60 * 60 * 24 * 1000;
+  const datapoints = [
+    {
```
polomani: maybe we should not copy-paste all the existing tests? only add a new one with the statuses here?

packages/beta/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -0,0 +1,107 @@
...
+    return this.addToMapAndReturn(response.data.items, response);
+  }
+
+  /**
+   *
+   * ```js
+   * const monthlyAggregatesData = await client.datapoints.retrieveDatapointMonthlyAggregates({ items: [{ id: 123 }] });
+   * ```
+   */
+  public retrieveDatapointMonthlyAggregates = async (
```
polomani, line 39: instead of copying the implementation here, could you just make a 

```
protected doRetrieveDatapointMonthlyAggregates = async <T extends DatapointAggregates[] = DatapointAggregates[]> (
```

then inside
```
public retrieveDatapointMonthlyAggregates
```
you would just call
```
doRetrieveDatapointMonthlyAggregates(...)
```
same here in beta, you'd just call
```
doRetrieveDatapointMonthlyAggregates<DatapointAggregates[]>
```

with your new type as a param

polomani (approved):
LGTM

PR #1085, opened 2024-04-11
feat: introduce remaining instance endpoints

danlevings:
Introduce the remaining instance endpoints to the SDK. 
- Retrieve
- upsert
- Delete
- Aggregate 
- Query
- Sync

and adjust tests to use these functins


packages/stable/src/api/instances/instancesApi.ts
```diff
@@ -93,4 +101,208 @@ export class InstancesAPI extends BaseResourceAPI<NodeOrEdge> {
...
+   *        externalId: describable2.externalId,
+   *        space: testSpace.space,
+   *      },
+   *    ],
+   *   })
+   * ```
+   */
+  public delete = async (params: ListOfSpaceExternalIdsRequestWithTyping) => {
+    await this.delete(params);
+  };
```
BugGambit: FYI, in the other SDK methods for upserting, deleting, etc., we support deleting more instances than the hard API limit. We do this by chunking the array into `limit`-sizes and, in parallel, deleting each chunk.
I noticed this when working on the FDM SDK that we don't do this.
E.g. here is the delete method for assets supporting deleting >1000 assets in a single SDK call: https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/api/assets/assetsApi.ts#L121-L127

BugGambit (approved):
LGTM,
But also clean this up? https://github.com/cognitedata/cognite-sdk-js/blob/master/packages/stable/src/__tests__/testUtils.ts#L115-L125

gulkaktus (approved):

PR #1086, opened 2024-04-12
fix: model revision create boundary condition type

bkuzma:
When creating a model revision, the `boundaryConditions` property expects an array of objects with only one key called `key`.


polomani:
we are getting rid of this type soon

PR #1088, opened 2024-04-24
feat(files-beta): add multi part upload sdk and tests (BND3D-4012)

G0nza1es:
**Introduce the new multi part file uploads to the Beta SDK:**

API Docs can be seen on here : [multi-part-upload-api](https://api-docs.cognite.com/20230101/tag/Files/operation/initMultiPartUpload)

Jira task: https://cognitedata.atlassian.net/browse/BND3D-4012?atlOrigin=eyJpIjoiNmYzNDk1ZmI0ZmIxNGZhYWFhZjhmZDgwOTAxMTdmODgiLCJwIjoiaiJ9

We introduce this api in Beta as a new funtionality and not extending current files capabilities. Once we promote it to _Stable_, we will eventually deprecate current `Upload ` and only use this capability as you can upload a content as a 1 chunk and this is a superset of what we have currently

**There are some limits in the API:** 

Maximum `250` parts can be requested
Upload chunk cannot have `mime/content-type`
If there are multiple chunks are requested , the last chunk cannot be smaller than 5MB : this validation is not yet implemented

**How this is tested**

- We have unit and integration tests . Integration tests ran only locally .

- We have `yalc ` this beta sdk and crated a react component within `3d-management` subapp  and created a Beta-Sdk context and we tested with binary files ranging from 1 MB to 10GB. We also tested sequantial and  concurrent uploads . Concurrent uploads with 10 concurrent threads uploads 10 GB file up to 7x faster than current GCS uploader component.


packages/beta/package.json
```diff
@@ -7,6 +7,7 @@
   "main": "dist/index.js",
   "types": "dist/src/index.d.ts",
   "version": "5.7.4",
+  "version": "5.8.0",
```
polomani: do not modify version number manually (!).
lerna does this automatically based on commit message

packages/beta/src/__tests__/api/filesMultiPart.int.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+      );
+      for (let i = 0; i < numberOfParts; i++) {
+        try {
+          await response.uploadPart(i, fileChunks[i]);
+          console.log(`Uploaded part ${i + 1} successfully.`);
+        } catch (error) {
+          console.error(`Error uploading part ${i + 1}:`, error);
+        }
+      }
+      await sleepPromise(2000);
```
msusag, line 89: Remove before merge :)

packages/beta/src/__tests__/api/filesMultiPart.int.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+      const [retrievedFile] = await client.files.retrieve([
+        { id: response.multiPartFileUploadResponse.id },
+      ]);
+      expect(retrievedFile.uploaded).toBeTruthy();
```
msusag: The check for upload=true may have to be repeated, as this is happening as part separate async processing within files API service and the specific cloud environment storage buckets.

After upload is done, it normally takes less than a second before upload is set to true. 
But in certain cases it can take longer time.

To avoid making this a flaky test, the check for upload=true should be retried a number of times.
If it is still not true after for example 2 minutes, then it should be ok to fail it.

packages/beta/src/__tests__/api/filesMultiPart.int.spec.ts
```diff
@@ -0,0 +1,296 @@
...
+        async (fileChunk, i) => await response.uploadPart(i, fileChunk)
+      )
+    );
+
+    expect(response.getNotCompletedParts()).toEqual([]);
+    console.log(
+      `Uploaded file  ${response.multiPartFileUploadResponse.id} successfully.`
+    );
+    await sleepPromise(2000);
+    const [retrievedFile] = await client.files.retrieve([
```
msusag: Repeat also here, to handle cases where upload=true takes a long time.

packages/beta/src/types.ts
```diff
@@ -322,3 +323,11 @@ export interface Subscription {
   subscriberExternalId?: CogniteExternalId;
   metadata?: Metadata;
 }
+export interface MultiPartFileUploadResponse extends FileInfo {
```
msusag, line 350: Add newline between each interface declaration, similar to others above?

samir-ajdarpasic (approved):
LGTM

packages/beta/src/__tests__/api/filesMultiPart.uint.spec.ts
```diff
@@ -0,0 +1,407 @@
+import nock from 'nock';
```
polomani: what is uint? is it a `unit` test?

G0nza1es: yes . it was a typo. fixed 👍🏽

packages/beta/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 - add sorting for alerts ([#1097](https://github.com/cognitedata/cognite-sdk-js/issues/1097)) ([1a08cca](https://github.com/cognitedata/cognite-sdk-js/commit/1a08ccaf7a2d0d16e057bcd8819774c3b5a513dd))
 
+- **files:** add support for multi part file upload ([#1088] closes (https://github.com/cognitedata/cognite-sdk-js/issues/1088))
+
```
polomani: please don't touch the changelog, it's autogenerated from commit messages

packages/beta/src/__tests__/api/filesMultiPart.int.spec.ts
G0nza1es: asserting with retries now

packages/beta/src/__tests__/api/filesMultiPart.int.spec.ts
G0nza1es: done above

packages/beta/CHANGELOG.md
G0nza1es: 👍🏽

polomani (approved):
can't judge the implementation code, but structurally this looks much better than the first edition

PR #1097, opened 2024-04-30
feat: add sorting for alerts

VebjornG:
Make sorting alerts available in the frontend. We are going to fix the tests for creating a subscriber and subscription, as well as listing them, in a different PR.
[Ticket](https://cognitedata.atlassian.net/browse/AH-3002?atlOrigin=eyJpIjoiMTEyNjNhZmMzM2Y4NDJkNDllMDZhOTQyZjczNzI0MDUiLCJwIjoiaiJ9)

### DIFF
--- a/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
@@ -109,8 +109,8 @@ describe('alerts api', () => {
           externalId: email,
         },
       ]);
-    } catch (error: AxiosError) {
-      expect(error.response?.status).toBe(400);
+    } catch (error) {
+      expect((error as AxiosError).response?.status).toBe(400);
     }
   });
 
--- a/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
@@ -2,6 +2,7 @@
 
 import CogniteClient from '../../cogniteClient';
 import { setupLoggedInClient } from '../testUtils';
+import { AxiosError } from 'axios';
 
 describe('alerts api', () => {
   const client: CogniteClient = setupLoggedInClient();
@@ -101,13 +102,16 @@ describe('alerts api', () => {
   });
 
   test('create subscribers', async () => {
-    const response = await client.alerts.createSubscribers([
-      {
-        email,
-        externalId: email,
-      },
-    ]);
-    expect(response.length).toBe(1);
+    try {
+      await client.alerts.createSubscribers([
+        {
+          email,
+          externalId: email,
+        },
+      ]);
+    } catch (error: AxiosError) {
+      expect(error.response?.status).toBe(400);
+    }
   });
 
   test('list subscribers', async () => {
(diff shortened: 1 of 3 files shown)


tuanng-cognite (approved):
Look good! Do we test it somehow?

packages/beta/src/api/alerts/alertsApi.ts
```diff
@@ -50,10 +52,14 @@ export class AlertsAPI extends BaseResourceAPI<Alert> {
     return this.createEndpoint(items);
   };
 
-  public list = async (filter?: AlertFilterQuery) => {
-    return this.listEndpoint<AlertFilterQuery>(
+  public list = async (filter?: AlertFilterQuery, sort?: AlertSortQuery) => {
```
polomani: why not just use a single query property for this same as others?
```
export interface AlertFilterQuery extends FilterQuery {
  filter?: AlertFilter;
  sort?: WhateverSortTypeHere
}
```

neringaalt:
@VebjornG can you disable failing  subscriber tests? We will fix them in another PR :)

packages/beta/src/api/alerts/alertsApi.ts
VebjornG: That's what i tried initially, but that didn't work for me because listEndpoint() expected two parameters, and not three, so i combined sort and filter so that i avoided changing listEndpoint()

packages/beta/src/api/alerts/alertsApi.ts
```diff
@@ -70,6 +76,10 @@ export class AlertsAPI extends BaseResourceAPI<Alert> {
     return this.channelsApi.list(filter);
   };
 
+  public sortAlerts = async (sort: AlertSortQuery) => {
+    return this.list(undefined, sort);
+  };
```
polomani: what is the meaning of this one? sort without the filter, or?

polomani (dismissed):
Let's give this a bit more attention, the implementation is a bit misaligned with the other APIs in the SDK.
Also, I am not sure what is the point of `sortAlerts` function.

packages/beta/src/api/alerts/alertsApi.ts
VebjornG: Is such a function unnecessary if sorting is to be added? @polomani 
If so I'm happy to remove it. I tried to follow the structure of the file.

polomani: sort and list are normally the same endpoint, hence it should be a single method as well.
I pushed a change

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -101,13 +102,16 @@ describe('alerts api', () => {
...
-    expect(response.length).toBe(1);
+    try {
+      await client.alerts.createSubscribers([
+        {
+          email,
+          externalId: email,
+        },
+      ]);
+    } catch (error) {
+      expect((error as AxiosError).response?.status).toBe(400);
```
polomani: the SDK never throws axios error directly. you should use CogniteError

PR #1098, opened 2024-04-30
feat: add simulation run data API

bkuzma:
add methods for `Simulation Data By Run Id` endpoint.

`https://{cluster}.cognitedata.com/api/v1/projects/{project}/simulators/runs/data/list`

### DIFF
--- a/packages/alpha/src/types.ts
+++ b/packages/alpha/src/types.ts
@@ -247,12 +247,12 @@ export interface SimulationRunDataOutput {
   referenceId: CogniteInternalId;
   value: string | number | string[] | number[];
   valueType: SimulationRunDataValueType;
-  unit: {
+  unit?: {
     name: string;
     externalId?: CogniteExternalId;
   };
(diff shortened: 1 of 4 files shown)


packages/alpha/src/types.ts
```diff
@@ -230,6 +230,45 @@ export interface SimulationRun {
...
+};
+
+export interface SimulationRunDataOutput {
+  referenceId: CogniteInternalId;
+  value: string | number | string[] | number[];
+  valueType: SimulationRunDataValueType;
+  unit: {
+    name: string;
+    externalId?: CogniteExternalId;
+  };
```
polomani: I think the unit is optional
```suggestion
  unit?: {
    name: string;
    externalId?: CogniteExternalId;
  };
```

bkuzma: Good catch, fixed that and another one.

polomani (approved):

PR #1102, opened 2024-05-02
fix: simulation run data api path

bkuzma:
Fixes the path for the simulation run data endpoint.

Renames `validationEndTime` to `runTime`.


polomani (approved):
good

PR #1108, opened 2024-06-10
fix(stable/models): pass 'inlineViews' as params instead of request body for data models retrieve

deep-cognite:
According to the documentation, the 'inlineViews' has to be passed as a query param: https://api-docs.cognite.com/20230101/tag/Data-models/operation/byExternalIdsDataModels

### DIFF
--- a/packages/stable/src/__tests__/api/datamodels.int.spec.ts
+++ b/packages/stable/src/__tests__/api/datamodels.int.spec.ts
@@ -123,6 +123,23 @@ describe('Data models integration test', () => {
     expect(datamodel.items[0].name).toEqual(datamodelCreationDefinition.name);
   });
 
+  it('should include views when retrieving datamodels', async () => {
+    const datamodel = await client.dataModels.retrieve(
+      [
+        {
+          space: TEST_SPACE_NAME,
+          externalId: datamodelCreationDefinition.externalId,
+        },
+      ],
+      {
+        inlineViews: true,
+      }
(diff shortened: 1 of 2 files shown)


packages/stable/src/__tests__/api/datamodels.int.spec.ts
```diff
@@ -121,6 +125,31 @@ describe('Data models integration test', () => {
     ]);
     expect(datamodel.items.length).toBe(1);
     expect(datamodel.items[0].name).toEqual(datamodelCreationDefinition.name);
+    expect(datamodel.items[0].views).toBeDefined();
```
danlevings, line 128: i thought without `inlineViews: true`, this would be undefined? or should we check that it always returns an empty array?

deep-cognite: Nope, without inline views you still get the view reference (i.e., type, space, external id and version) - see screenshot:
image

"inline views" is just a way to get the view data in the same response (reducing the amount of round trips we have to do to the backend) - see screenshot: 
image

danlevings: aha - so we could add here that e.g. properties should be undefined?

deep-cognite: added

danlevings (approved):

PR #1114, opened 2024-06-18
fix(simint): fix tests and update filter for routines and models

lpereiracgn:
* Add filter for versions listing in model revisions and routine revisions
* Fix tests


packages/alpha/package.json
```diff
@@ -6,7 +6,7 @@
   "homepage": "https://github.com/cognitedata/cognite-sdk-js#readme",
   "main": "dist/index.js",
   "types": "dist/src/index.d.js",
-  "version": "0.28.1",
+  "version": "0.28.2",
```
polomani: we never change package version manually here.
version number is based on the commit title, e.g. fix() bumps a minor version, and so on

packages/alpha/src/__tests__/api/simulationRunsApi.int.spec.ts
```diff
@@ -1,21 +1,140 @@
 // Copyright 2023 Cognite AS
+
+
+// Copyright 2023 Cognite AS
+
```
polomani:
```suggestion
```
duplicated

packages/alpha/src/__tests__/api/simulatorRoutinesApi.spec.ts
```diff
@@ -82,7 +81,7 @@ describeIf('simulator routines api', () => {
         externalId: modelRevisionExternalId,
         modelExternalId,
         description: 'test sim model revision description',
-        fileId: 3747718694331206,
+        fileId: 6396395402204465,
         metadata: {},
```
polomani: could you remove this metadata field from the type? we should not use it anymore

packages/alpha/src/types.ts
```diff
@@ -235,6 +235,23 @@ export interface SimulationRunId {
   runId: CogniteInternalId;
 }
 
+interface SimulatorRoutineIO {
+  referenceId: string;
+  name: string;
```
polomani, line 240:
```suggestion
  name?: string;
```

polomani (approved):
LGTM

PR #1116, opened 2024-06-21
feat: add test for pagination cursor

VebjornG:
[Ticket](https://cognitedata.atlassian.net/browse/AH-3373?atlOrigin=eyJpIjoiYjk0NzFmYWQ2Y2I0NGVlYmIxMjFiMzhhYjRiM2ZkNDkiLCJwIjoiaiJ9)


packages/beta/src/types.ts
```diff
@@ -184,6 +184,7 @@ export interface AlertSort {
 export interface AlertFilterQuery extends FilterQuery {
   filter?: AlertFilter;
   sort?: AlertSort;
+  cursor?: string;
```
polomani: `extends FilterQuery` already gives a cursor

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -201,3 +201,27 @@ test('sort alerts', async () => {
...
+    sort: {
+      property: 'createdTime',
+      order: 'desc',
+    },
+    cursor: '',
+  });
+  expect(response.items.length).toBe(1);
+  // test for more than 1000 items
+  const response2 = await client.alerts.list({
+    limit: 1001,
```
polomani: limits is always per page on all other resources on CDF.
to get more then 1000, user would use `autoPagingToArray()`.
check how it's done in other resources, e.g.
```
const timeseries = await response.autoPagingToArray({ limit: 5000 });
```

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -201,3 +201,29 @@ test('sort alerts', async () => {
...
+    .list({
+      sort: {
+        property: 'createdTime',
+        order: 'desc',
+      },
+      cursor: response?.nextCursor,
+    })
+    .autoPagingToArray({ limit: 1001 });
+
+  expect((await alerts).length).toBeGreaterThan(0);
```
saagar-cognite: question(non-blocking): is there way to test the number of alerts to be greater than 1000 ?

saagar-cognite (dismissed):
Just 1 nitpick for better testing.

saagar-cognite:
@VebjornG I think you'll need PR title adhering to sdk release guidelines.

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
VebjornG: I suppose i could do this: `expect(alerts.length).toBeGreaterThan(1000);` @saagar-cognite

saagar-cognite: Yes. That'd make sense.
But you might have to make sure that there are more than 1000 alerts present.. So mabye create 1001 alerts and then delete them after the test would be helpful ? 🤔

VebjornG: I agree. I do so in my latest commit @saagar-cognite 👍🏻

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -201,3 +201,44 @@ test('sort alerts', async () => {
   });
   expect(response.items.length).toBeGreaterThan(0);
 });
+
+test('cursor pagination', async () => {
+  const client: CogniteClient = setupLoggedInClient();
```
saagar-cognite: suggestion(blocking): we already have a client instantiated above at starting of `describe` block.
We can avoid initiating it over here.

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -201,3 +201,44 @@ test('sort alerts', async () => {
...
+    }))
+  );
+
+  const alerts = client.alerts
+    .list({
+      sort: {
+        property: 'createdTime',
+        order: 'desc',
+      },
+      cursor: response?.nextCursor,
```
saagar-cognite: question(blocking): should this cursor be used from previous response ?
I don't think we need to pass cursor when we're autoPagingToArray.

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -3,6 +3,7 @@
 import { CogniteError } from '@cognite/sdk-core';
 import CogniteClient from '../../cogniteClient';
 import { setupLoggedInClient } from '../testUtils';
+//import { AlertCreate } from '@cognite/sdk-beta/dist';
```
saagar-cognite: nitpick: remove the unused import comment.

saagar-cognite (approved):
LGTM!

PR #1117, opened 2024-06-24
feat: add test for limit

VebjornG:
[Ticket](https://cognitedata.atlassian.net/browse/AH-3372?atlOrigin=eyJpIjoiZjJjOTczZWYyNzYyNDY4YjhkZDE0NDk4OThhM2Y1YTMiLCJwIjoiaiJ9)

### DIFF
--- a/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
@@ -205,7 +205,7 @@ describe('alerts api', () => {
     const alerts = Array.from({ length: 10 }).map((_, i) => ({
       source: 'smth',
       channelExternalId,
-      externalId: `${alertExternalId}_${i}`,
+      externalId: `test_limit_${i}`,
     }));
     await client.alerts.create(alerts);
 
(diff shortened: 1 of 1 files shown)


tuanng-cognite (approved):
Nice! I thought it did not support the limit

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -189,15 +189,21 @@ describe('alerts api', () => {
...
+      },
+    });
+    expect(response.items.length).toBeGreaterThan(0);
+  });
+
+  test('test limit', async () => {
+    const response = await client.alerts.list({
+      limit: 1,
+    });
+    expect(response.items.length).toBe(1);
```
saagar-cognite: question(blocking): How many alerts are already there ?

saagar-cognite: I think we should create 2 alerts and then check for limit: 1 and see if it returns just 1.

VebjornG: I'm creating 10 alerts in my latest commit @saagar-cognite

saagar-cognite (approved):
LGTM!
Though sorting is not tested as a feature but it should be safe to assume that its implemented correctly in backend.

saagar-cognite:
Just needs the fix of externalIds I guess.

PR #1103, opened 2024-05-07
feat(annotations): generate annotations for isolation planning app

SudilHasithaCognite:
Add IsoPlan annotations to be consistent with migration.

References:

- https://github.com/cognitedata/annotation-types/pull/136
- https://github.com/cognitedata/annotation-types/pull/157
- https://github.com/cognitedata/context-api/pull/4057
- https://github.com/cognitedata/service-contracts/pull/2505

PR #1118, opened 2024-06-29
chore: revert breaking change created by isoplan annotations

SudilHasithaCognite:
Reverts cognitedata/cognite-sdk-js#1103

- This PR do not reflect the breaking change it introduces by making fileRef optional.
- A fix is made for the breaking change in this [PR](https://github.com/cognitedata/fusion/pull/6966/files)


BugGambit:
@SudilHasithaCognite I suggest we close this PR and leave the breaking change in the SDK. The harm seems low and we will soon release a new major version of the SDK.

PR #1119, opened 2024-06-29
feat: create this to reflect the breaking change introduced by isoplan annotations

SudilHasithaCognite:
- This PR is created to replace the breaking change introduced by the PR which makes fileRef optional. https://github.com/cognitedata/cognite-sdk-js/pull/1103
- The changes are reverted in this [PR](https://github.com/cognitedata/cognite-sdk-js/pull/1118)


BugGambit:
@SudilHasithaCognite looks like the version in the package.json was manually edited? Please don't do that.

This is a breaking change, which will bump major and should be coordinated across Cognite to bulk several breaking changes together.
Is this change still a priority? I am working on releasing a new major version of the SDK and I am looking into changes we want to include

BugGambit (changes requested):
Let's not bump a major version like this and rather bulk several breaking changes. See my comment

SudilHasithaCognite:
No, you can safely discard this if necessary. We initially thought this would be
significant enough for a version bump. After this
https://github.com/cognitedata/fusion/pull/6966/files we thought of not changing it. But @cemreyavuz and @G0nza1es mentioned it will be necessary in the long run to do so.

BugGambit:
@SudilHasithaCognite I suggest we close this PR and leave the breaking change in the SDK. The harm seems low and we will soon release a new major version of the SDK.

PR #1121, opened 2024-08-02
feat(documents): add sdk support for the elements API

tommythorsen:
The document elements API is in alpha, so this commit enables code generation for the alpha package.

The elements API only returns data on a very granular level. Everything is per individual character, so we add some helper functions to combine characters into words, lines or high-level elements.


packages/alpha/src/api/documents/documentsApi.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+class ListWithBoundingBox extends LinesWithBoundingBox {
+  constructor(element: DocumentListElement) {
+    const lines = element.lines.map((line) => new LineWithBoundingBox(line));
+    super(lines);
+  }
+}
+
+class TitleWithBoundingBox extends LinesWithBoundingBox {
```
andersfylling, line 148: why not use the auto generated element types?

packages/alpha/src/api/documents/documentsApi.ts
```diff
@@ -0,0 +1,290 @@
...
+  level: number;
+
+  constructor(element: DocumentTitleElement) {
+    const level = element.level || 3;
```
andersfylling, line 152: why 3, some levels may be eg. 6? I adjusted the document-science package as some extracted titles had a higher level than 3.

tommythorsen: At the time I wrote this code, 3 was the highest level possible. I could pick a different number, but 1) I think value will in practice always be set, and 2) maybe 3 is a nice average value to use for a default?

Do you have a different number in mind?

packages/alpha/src/api/documents/documentsApi.ts
tommythorsen: These are not in the API. They are added convenience functionality in the SDK

packages/alpha/src/api/documents/documentsApi.ts
andersfylling: No, not sure I have any comments that can help here. im not sure what behavior would be expected if no value is present tbh.
maybe just remove it if it's always set?

andersfylling (approved):
lgtm!

PR #1127, opened 2024-08-16
chore: remove dependencies and fix typescript warnings

BugGambit:
Changes:
- Removing dependencies url, query-string and is-buffer. They are either trivial or have native node/browser replacements.
- Fix some typescript warnings.
- Increase timeout on an integration test (flaky).


packages/core/src/authFlows/__tests__/auth.spec.ts
```diff
@@ -246,7 +246,7 @@ describe('Cognite Auth', () => {
         });
         expect(spiedLocationAssign).toBeCalledTimes(1);
         expect(spiedLocationAssign.mock.calls[0][0]).toMatchInlineSnapshot(
-          `"https://example.com/login/redirect?errorRedirectUrl=https%3A%2F%2Ferror-redirect.com&project=my-tenant&redirectUrl=https%3A%2F%2Fredirect.com"`
+          `"https://example.com/login/redirect?project=my-tenant&redirectUrl=https%3A%2F%2Fredirect.com&errorRedirectUrl=https%3A%2F%2Ferror-redirect.com"`
```
BugGambit, line 249: They are the same, just query params changed order.

packages/stable/src/__tests__/api/documentsApi.int.spec.ts
```diff
@@ -173,23 +173,27 @@ describe('Documents integration test', () => {
...
+
+        const pdfPrefix = [0x25, 0x50, 0x44, 0x46, 0x2d]; // %PDF-
+        expect(resp.byteLength).toBeGreaterThan(pdfPrefix.length);
+        const frontSlice = resp.slice(0, pdfPrefix.length);
+        expect(frontSlice.byteLength).toStrictEqual(pdfPrefix.length);
+        const match = Buffer.from(frontSlice, 0).equals(
+          Buffer.from(pdfPrefix, 0)
+        );
+        expect(match).toBe(true);
+      },
```
BugGambit, line 194: Indent caused by added timeout

packages/core/src/authFlows/__tests__/auth.spec.ts
cyriaquebrousse: why is the change needed then? 🤔

BugGambit: Using URLSearchParams vs query-string which use different ordering rules

BugGambit: And the matching here is on string level, it doesn't know this represents an URL

packages/core/src/utils.ts
```diff
@@ -39,6 +38,21 @@ export function isBrowser() {
   );
 }
 
+function isBuffer(obj: any): boolean {
```
cyriaquebrousse, line 41: does this essentially check whether `obj` is a byte array? My JS is a bit rusty.

danlevings (approved):
sgtm, looks like just a tidy up 👌

packages/core/src/authFlows/__tests__/auth.spec.ts
cyriaquebrousse: Right!

packages/core/src/utils.ts
BugGambit: No, just a nodejs Buffer.
The reason we can't use Buffer.isBuffer is that this code must be compatible with browsers where `Buffer` isn't a concept

cyriaquebrousse (approved):

PR #1128, opened 2024-08-17
chore: upgrade to typescript 5 & vite

BugGambit:
- Change SDK to be an es5 package to an ES / module package (breaking change!)
- Upgrade Typescript from v3 to v5 (and fix all breaking changes)
- Migrate from plain rollup to Vite

PR #1131, opened 2024-08-19
chore: replace eslint/prettier with biome

BugGambit:
Part of splitting up this PR: https://github.com/cognitedata/cognite-sdk-js/pull/1128

Biome is an efficient linter and formatted.
This PR replaces eslint/prettier with [Biome](https://biomejs.dev/) and formats our files accordingly.

Jira: https://cognitedata.atlassian.net/browse/AUTH-3069


packages/core/src/httpClient/basicHttpClient.ts
```diff
@@ -218,11 +223,40 @@ export class BasicHttpClient {
...
+            default: {
+              throw new Error(
+                `Unsupported value query parameter type: ${typeof value}, ${key}: ${value}`
+              );
+            }
+          }
+        },
+        {} as Record<string, string>
+      );
+      const search = new URLSearchParams(normalizedParams).toString();
```
BugGambit, line 259: Biome was strictier in the type and we had to open up for `params` to be type of object. `URLSearchParams` can't take in a generic object, so we need to handle each type.

danlevings (approved):
it seems like the only file worth reviewing is the biome.json / package.json file? everything else is either automatic linting or deleting things that arent needed anymore

if so, its not much to review, looks fine to me!

PR #1134, opened 2024-08-22
feat(simint): add sort property to filter queries

bkuzma:
This adds sort and limit properties to filter queries that support those fields in the API.

Also fixes a typo.

### DIFF
--- a/packages/alpha/src/types.ts
+++ b/packages/alpha/src/types.ts
@@ -564,7 +564,7 @@ export interface SimulatorRoutineRevisionslFilter {
   createdTime?: DateRange;
 }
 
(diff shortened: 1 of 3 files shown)


packages/alpha/src/types.ts
```diff
@@ -73,6 +73,7 @@ export interface Simulator {
 
 export interface SimulatorFilterQuery extends FilterQuery {
   filter?: object;
+  limit?: number;
 }
```
```suggestion
export interface SimulatorFilterQuery extends FilterQuery {
  filter?: object;
}

```
filter query has cursor and limit already

bkuzma: 🤦 

fixed

polomani (approved):

PR #1135, opened 2024-08-22
chore: upgrade to typescript 5

BugGambit:
This PR upgrades to Typescript 5 from 3, and resolve breaking changes.
We also replace Rollup with Vite and change the SDKs from es5 compatibility to become es6 modules.

We are *not* merging this into `master` as we would like to bulk up multiple breaking changes before the final release. We, therefore, use the target branch `release-v10` (protected), and this PR is against that branch.

We have configured this branch to publish release candidates to npm.


.github/workflows/release.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
           GITHUB_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
-        run: yarn lerna version --conventional-commits --create-release github --no-private --yes
+        run: yarn lerna version --conventional-commits --conventional-prerelease --preid rc --create-release github --no-private --yes
```
BugGambit, line 88: In this major release, we start with releasing release candidates (rc)

scripts/extractCodeSnippets.js
```diff
@@ -10,17 +10,28 @@ function stripMarkdownCodeSnippet(rawCode) {
...
+      const operationId = match[2];
+      if (value.length !== 2) {
+        throw new Error('Unexpected length of summary');
+      }
+      if (value[1].kind !== 'code') {
+        throw new Error(`Expected code snippet, but got ${value[1].kind}`);
       }
-      const operationId = matches[2];
-      const rawCode = object.text;
+      const rawCode = value[1].text;
```
BugGambit, line 34: The typedoc output format has changed in the recent version

.github/workflows/release.yaml
BugGambit: fyi @polomani

anjalirai2626 (approved):
LGTM, there isn't any change in logic and mostly config 👍 

More eyes are welcome if needed!

PR #1136, opened 2024-08-22
test: skip pagination test temporarily

saagar-cognite:
https://cognitedata.slack.com/archives/C05A4PX656F/p1724010276336469

After some discussion on the above thread, we have decided to skip this test until we find a fix.

Reason:
The cursor pagination test is creating more than 1000 alerts which is currently very slow.
Which could be a probably reason for 504 error.

We will later either try to make smaller page size and create less alerts or do a cleanup before all tests are run to clear all alerts which could help with sorting in database.


saagar-cognite:
Closing this in favour of
https://github.com/cognitedata/cognite-sdk-js/pull/1133

PR #1138, opened 2024-08-26
test: fix test for pagination cursor

VebjornG:
In the previous test of the pagination cursor in alert-api, 1001 alerts was created. This created 504s, so we reduced that number to 50. 
[Ticket](https://cognitedata.atlassian.net/browse/AH-3708).

### DIFF
--- a/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
+++ b/packages/beta/src/__tests__/api/alertsApi.int.spec.ts
@@ -228,7 +228,7 @@ describe('alerts api', () => {
 
   vi.setConfig({ testTimeout: 30_000 });
 
-  test.skip('cursor pagination', async () => {
+  test('cursor pagination', async () => {
     // create channel for the next test
     const channelsToCreate = [
       {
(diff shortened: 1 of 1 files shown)


packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -290,9 +269,9 @@ describe('alerts api', () => {
           order: 'desc',
         },
       })
-      .autoPagingToArray({ limit: 1001 });
+      .autoPagingToArray({ limit: 50 });
 
-    expect((await alerts).length).toBeGreaterThan(1000);
+    expect((await alerts).length).toBe(50);
```
BugGambit, line 274: I don't understand how this tests pagination. Aren't we just fetching 50 alerts without pagination here?
Also `autoPagingToArray({ limit: 50 });` will not provide a `limit` query param to the API as this is logic for the SDK. If you want to send a `limit` query param to the API, then it should be here:
```js
const alerts = client.alerts.list({..., limit: 25})`
```

BugGambit (dismissed):
Also, this is not a `fix` as specified in the PR title and commit.
From the end-users point of view, this PR doesn't fix any bugs.

The proper prefix is `test`

See the contribution docs: https://github.com/cognitedata/cognite-sdk-js/blob/master/CONTRIBUTING.md#making-changes

packages/beta/src/__tests__/api/alertsApi.int.spec.ts
```diff
@@ -244,55 +245,34 @@ describe('alerts api', () => {
         property: 'createdTime',
         order: 'desc',
       },
-      cursor: '',
     });
     expect(response.items.length).toBe(1);
```
vincent-cognite: No alarm has been created at this point, so why is this expected?
Is it coming from a previous test?

vincent-cognite (approved):
looks like you are testing the pagination correctly. I am not sure I understand how the test is structured though...

PR #1139, opened 2024-08-26
fix: dummy release to override a bad one

BugGambit:
Context: https://cognitedata.slack.com/archives/C0425G1JP2A/p1724673350421259

We had a bad release from the `release-v10` branch which overwrote the `latest` tag instead of the `next` tag


cemreyavuz (approved):

anjalirai2626 (approved):
Lgtm, it has some linting issue though!

PR #1147, opened 2024-09-17
feat(simint): add and remove properties from simulators API

bkuzma:


packages/alpha/src/types.ts
```diff
@@ -116,6 +117,7 @@ export interface SimulatorIntegrationCreate {
 
 export interface SimulatorIntegrationFilter {
   simulatorExternalIds?: CogniteExternalId[];
+  active: boolean;
```
```suggestion
  active?: boolean;
```

polomani: it's a filter, so optional

bkuzma: Fixed it 👍

polomani (approved):
LGTM

PR #1148, opened 2024-09-17
refactor(api-key): remove unsupported api-key feature

BugGambit:
As part of the v10 release, we would like to clean up api-key references in the SDK as api-keys isn't supported by CDF anymore (and hasn't been supported for a long time - years).

Jira: https://cognitedata.atlassian.net/browse/AUTH-3104


akhramov (approved):

packages/core/src/constants.ts
```diff
@@ -8,9 +8,7 @@ export const DEFAULT_DOMAIN = 'cognitedata.com';
 export const BASE_URL = `https://${DEFAULT_CLUSTER}.${DEFAULT_DOMAIN}`;
 
 /** @hidden */
-export const AUTHORIZATION_HEADER = 'Authorization';
-/** @hidden */
-export const API_KEY_HEADER = 'api-key';
+export const AUTHORIZATION_HEADER = 'authorization';
```
cemreyavuz, line 11: Feels like this change should break things but i'm not sure if we add both versions of the authorization header.

BugGambit: Tests passed and the api shoudn’t care about the casing

BugGambit: Or what are you worried about will break?

cemreyavuz: > the api shoudn’t care about the casing

I'm worried about this part. If we're using the same constant everywhere in the repo (which we probably do), i'm not surprised tests are working fine. Do we have a case that we pass this header manually from the outside of the sdk?

cemreyavuz: (this is not a blocker though, just nitpicking)

BugGambit: Let me check how easy it is to undo the change. But we run integration tests so the API accepts the header

cemreyavuz: Then don't worry too much about it, we would be testing this sdk version anyway (and it would be almost instant to figure it out)

cemreyavuz: This is the only place in `fusion` repo that we manually pass the authorization header: https://github.com/cognitedata/fusion/blob/a3b007c8255cb0475878f6767e95a658ba8c34d9/libs/shared/cogidp-sdk/src/lib/client.ts#L226-L228. But we're not using js sdk there so should be good imo

PR #1150, opened 2024-09-17
refactor(auth): remove legacy auth & ADFS support

BugGambit:
As part of the v10 release we would like to remove legacy auth (the API is gone so legacy auth doesn't work) and ADFS support. ADFS is only used by Aramco and Cntxt and we would rather transition them to CogIdP.

Jira:
- https://cognitedata.atlassian.net/browse/AUTH-3108
- https://cognitedata.atlassian.net/browse/AUTH-3103


packages/core/src/baseCogniteClient.ts
```diff
@@ -126,29 +99,10 @@ export default class BaseCogniteClient {
...
-        credentials,
-        provider
-      );
-    }
-
-    if (isBrowser() && !isUsingSSL() && !options.noAuthMode) {
-      console.warn(
-        'You should use SSL (https) when you login with OAuth since CDF only allows redirecting back to an HTTPS site'
-      );
-    }
```
cemreyavuz, line 139: Wouldn't this warning be logged even if we're not using legacy auth?

packages/core/src/baseCogniteClient.ts
```diff
@@ -107,16 +75,21 @@ export default class BaseCogniteClient {
...
+      throw Error('`CogniteClient` is missing parameter `options`');
+    }
+
+    if (!isString(options.appId)) {
+      throw Error('options.appId is required and must be of type string');
+    }
+
+    if (!isString(options.project)) {
+      throw Error('options.project is required and must be of type string');
+    }
```
cemreyavuz, line 84: Feels like we can still use `verifyOptionsRequiredFields` in some form to manage this easier - but i'm also not sure how many new "required" things we would be adding to `options`.

cemreyavuz (approved):
LGTM btw, comments are not blockers

packages/core/src/baseCogniteClient.ts
BugGambit: Yes, but it was placed there due to legacy auth. No reason to restrict to https anymore.
The https restriction came from legacy auth.

packages/core/src/baseCogniteClient.ts
BugGambit: The `verifyOptionsRequiredFields` did a bit more and mostly related to auth options.
Keeping this as required for now as it's aligned with version 9.

PR #1152, opened 2024-09-18
refactor(projects): remove deprecated API endpoints

BugGambit:
As part of the v10 release, we would like to remove deprecated Project API endpoints.
These endpoints are already removed from API docs.

Over time, we want all changes to projects to go through CogIdP.

Jira: https://cognitedata.atlassian.net/browse/AUTH-3244


packages/stable/src/types.ts
```diff
@@ -1660,44 +1656,12 @@ export interface ProjectResponse {
   name: ProjectName;
   urlName: UrlName;
   oidcConfiguration?: OidcConfiguration;
-  userProfilesConfiguration: UserProfilesConfiguration;
```
cemreyavuz, line 1642: Did we remove the `userProfilesConfiguration` from project response as well?

cemreyavuz (approved):
LGTM - one question

packages/stable/src/types.ts
BugGambit: Good catch! I will undo that.

cemreyavuz (approved):

PR #1153, opened 2024-09-18
refactor(core): clean up Client constructor and deprecate getToken

BugGambit:
As part of the v10 release, we are deprecating the `getToken` method. The name is confusing, and we instead rename it to `oidcTokenProvider` which is a more accurate description.

Jira: https://cognitedata.atlassian.net/browse/AUTH-3107


cemreyavuz (approved):
RIP `getToken`

PR #1154, opened 2024-09-18
docs: clean up samples folder

BugGambit:
Many of the samples fail, are outdated or are against what we recommend.
Cleaning it up for now, and we can hopefully add better samples in the future.

Jira: https://cognitedata.atlassian.net/browse/AUTH-3106


packages/alpha/vite.config.js
```diff
@@ -2,11 +2,14 @@ import { resolve } from 'node:path';
...
 import dts from 'vite-plugin-dts';
 
+const externals = ['@cognite/sdk', '@cognite/sdk-core'];
+
 export default defineConfig({
   plugins: [
     dts({
       exclude: ['**/__tests__/**/*'],
       entryRoot: '.',
+      aliasesExclude: externals,
```
BugGambit, line 12: Without this, then the typescript declaration files will use relative paths instead of `@cognite/sdk-core`

samples/nodejs/oidc-typescript/package.json
```diff
@@ -5,11 +5,12 @@
   "license": "Apache-2.0",
   "scripts": {
     "tsc": "tsc",
+    "test": "yarn g:tsc -p ./tsconfig.json",
```
BugGambit, line 8: Add this to verify that the example has valid typescript syntax

packages/alpha/vite.config.js
cemreyavuz: Is this because they are defined as paths in tsconfig?

BugGambit: Maybe

cemreyavuz (approved):
LGTM

PR #1155, opened 2024-09-19
refactor: favour private fields over typescript private annotation

BugGambit:
Jira: https://cognitedata.atlassian.net/browse/AUTH-3261

The motivation for this change is that we have seen Cognite applications abusing private fields by using `@ts-ignore` and accessing them anyway. With this change, that will not be possible as private fields are checked during runtime vs compilation time.


tsconfig.common.json
```diff
@@ -8,15 +8,15 @@
   },
   "compilerOptions": {
     "module": "es6",
-    "target": "es5",
+    "target": "es6",
```
BugGambit, line 11: Private field are only supported in Ecmascript 2015/es6 or above.
We want SDK v10 to target es6

BugGambit:
Closing as the value is low and can be done after the v10 release.

PR #1120, opened 2024-07-02
chore(stable): generate latest types & add types for instances - 02.07.2024

deep-cognite:
* Enable codegen for 'InstancesAPI'
* Pull latest openapi snapshot for the stable package
* Affected endpoints: annotations*, vision* & instances
* Update the methods in InstanceAPI with changes to param and response type name (search, retrieve, aggregate)

NOTE (*): Due to changes done in the following commit https://github.com/cognitedata/service-contracts/commit/191ddb3b200091afa65a3adf9ba2482ff11aa7ed, the types for vision (similar to annotations) were changed from containing 'Cognite' to 'Cogmono'. Not sure what to do here.

PR #1159, opened 2024-09-24
feat: adding functions and sessions apis

bdimitrijoski:
This PR adds sessions API and partial support for functions API in the SDK.
I covered the APIs with unit tests and added code samples.

There were some problems (mostly, generated some types with `any` and later it failed on lint) with codegen when I tried to generate the types for functions so at the end I skipped codgen for this API.


packages/stable/src/exports.gen.ts
```diff
@@ -8,12 +11,12 @@ export {
...
-  AnnotationsCogniteAnnotationTypesImagesAssetLink,
-  AnnotationsCogniteAnnotationTypesImagesInstanceLink,
-  AnnotationsCogniteAnnotationTypesPrimitivesGeometry2DGeometry,
-  AnnotationsCogniteAnnotationTypesPrimitivesGeometry3DGeometry,
+  AnnotationsCogmonoAnnotationTypesDiagramsAssetLink,
+  AnnotationsCogmonoAnnotationTypesDiagramsInstanceLink,
+  AnnotationsCogmonoAnnotationTypesImagesAssetLink,
+  AnnotationsCogmonoAnnotationTypesImagesInstanceLink,
+  AnnotationsCogmonoAnnotationTypesPrimitivesGeometry2DGeometry,
+  AnnotationsCogmonoAnnotationTypesPrimitivesGeometry3DGeometry,
```
polomani, line 16: looks like a breaking change? any way possible to keep the old names?

bdimitrijoski: That's what annoys me the most, I don't have control over codegen and this is a change coming from the tool. How can I avoid this breaking change?

polomani: My advice is not to use it since you spend more time working around it than it requires writing a few type definitions.

but maybe @andersfylling or @BugGambit have a better idea

packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,152 @@
...
+   * client.functions.retrieve({ items: [{ id: '123' }] })
+   * ```
+   *
+   */
+  async retrieve(
+    scope?: IgnoreUnknownIdsField & { items: FunctionIdEither[] }
+  ): Promise<HttpResponse<FunctionListResponse>> {
+    return await this.post<FunctionListResponse>(this.byIdsUrl, {
+      data: scope,
+    });
```
polomani, line 80: check how assets/timeseries API do it, the default `retrieveEndpoint` gives a bit more (e.g. chunking) 
```suggestion
    return super.retrieveEndpoint(ids, params);
```

packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,152 @@
...
+   * ```js
+   * client.functions.list({ limit: 100 })
+   * ```
+   *
+   */
+  async list(scope?: LimitList): Promise<HttpResponse<FunctionListResponse>> {
+    return await this.post<FunctionListResponse>(this.listPostUrl, {
+      data: scope || { limit: 100 }, // default limit
+    });
+  }
```
polomani, line 49:
```suggestion
  async list(scope?: LimitList): Promise<HttpResponse<FunctionListResponse>> {
    return super.listEndpoint(this.callListEndpointWithPost, scope);
  }
```
super.listEndpoint will do a bit more for free (e.g. auto pagination)

packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,152 @@
...
+   * ```
+   *
+   */
+  async filter(
+    scope?: FunctionListScope
+  ): Promise<HttpResponse<FunctionListResponse>> {
+    return await this.post<FunctionListResponse>(this.listPostUrl, {
+      data: scope,
+    });
+  }
```
polomani, line 65: why do you expose both list and filter? this one should be enough since it has all the filtering capabilities
https://api-docs.cognite.com/20230101/tag/Functions/operation/listFunctions

packages/stable/src/exports.gen.ts
BugGambit: There is a discussion about this specific breaking change here: https://github.com/cognitedata/cognite-sdk-js/pull/1120#issuecomment-2209470020
I suggest:
1. Revert the breaking change.
2. Change the base branch to `release-v10` where we, in a few days, will release a new SDK major version where breaking changes are allowed.

packages/stable/src/api/functions/functionCallsApi.ts
```diff
@@ -0,0 +1,69 @@
...
+  FunctionCallLogResponse,
+  FunctionCallRequest,
+  FunctionCallResponse,
+  FunctionCalledResponse,
+} from './types.gen';
+
+export class FunctionCallsApi extends BaseResourceAPI<FunctionCall> {
+  /**
+   * @hidden
+   */
```
BugGambit, line 15: Why hidden?

packages/stable/src/__tests__/api/functions.unit.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+    },
+    cpu: 1,
+    memory: 1.5,
+    runtime: 'py311',
+    runtimeVersion: 'Python 3.8.13',
+    error: {
+      code: 400,
+      message: 'Could not build function.',
+    },
+  } as CogniteFunction;
```
BugGambit, line 43: Why `as CogniteFunction` vs `const mockFunction: CogniteFunction = {}`?

packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,152 @@
...
+    super(...args);
+
+    const [baseUrl, httpClient, map] = args;
+    this.functionCallsAPI = new FunctionCallsApi(baseUrl, httpClient, map);
+  }
+
+  /**
+   * @hidden
+   */
+  public get calls() {
```
BugGambit, line 33: What is this used for and why is it hidden?

packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,152 @@
...
+   * ```
+   */
+  async status(): Promise<HttpResponse<FunctionsActivationResponse>> {
+    return await this.get<FunctionsActivationResponse>(this.url('status'));
+  }
+
+  /**
+   * @hidden
+   */
+  async limits(): Promise<HttpResponse<FunctionsLimitsResponse>> {
```
BugGambit, line 112: Why is it hidden?

scripts/extractCodeSnippets.js
```diff
@@ -24,7 +24,14 @@ function findAllCodeSnippetsInJsDoc(jsDoc) {
       if (!codeSnippets.has(operationId)) {
         codeSnippets.set(operationId, []);
       }
-      codeSnippets.get(operationId).push(stripMarkdownCodeSnippet(rawCode));
+
+      try {
+        codeSnippets.get(operationId).push(stripMarkdownCodeSnippet(rawCode));
```
BugGambit, line 40: What failed?

BugGambit (commented):
Nice PR. I left some comments.
In retro, I believe the PR should have been two PRs. One for Functions and one for Sessions.

packages/stable/src/api/functions/functionsApi.ts
bdimitrijoski: Yeah, not sure, I just did both

packages/stable/src/api/functions/functionsApi.ts
bdimitrijoski: The reason why is because the API does not support pagination

packages/stable/src/api/functions/functionsApi.ts
bdimitrijoski: The functions API endpoint does not support chunking

packages/stable/src/exports.gen.ts
bdimitrijoski: @polomani how can I keep the old names?

packages/stable/src/api/functions/functionCallsApi.ts
bdimitrijoski: Snippets or samples, one of those tools was failing

packages/stable/src/__tests__/api/functions.unit.spec.ts
bdimitrijoski: Isn't doing the same?

packages/stable/src/api/functions/functionsApi.ts
bdimitrijoski: This is pointing to the function calls methods (API) example:
`functionsApi.calls.list()`

It is hidden because the snippets or samples tools were failing

scripts/extractCodeSnippets.js
bdimitrijoski: When the comment above the method is not in a format that this tool is expecting, it is failing without telling why or showing a useful error message. It took me quite some time to figure out why (tnx to @polomani )

packages/stable/src/__tests__/api/functions.unit.spec.ts
BugGambit: Hm, not completely. Both will result in `mockFunction` being recognized as the type `CogniteFunction`, but `as CogniteFunction` will ignore type errors.

```ts
type Foo = {
    bar: number;
}

const a: Foo = {} // this will fail as `bar` is missing
const b = {} as Foo; // this will NOT fail
```

packages/stable/src/api/functions/functionCallsApi.ts
BugGambit: Ok, we should look into that. `@hidden` will hide this endpoint from the documentation.
I can look into it.

packages/stable/src/api/functions/types.gen.ts
```diff
@@ -0,0 +1,445 @@
...
+  /** The number of milliseconds since 00:00:00 Thursday, 1 January 1970, Coordinated Universal Time (UTC), minus leap seconds. */
+  timestamp?: EpochTimestamp;
+}
+export interface FunctionCallLogResponse {
+  items: FunctionCallLogEntry[];
+}
+export interface FunctionCallRequest {
+  // biome-ignore lint/suspicious/noExplicitAny: couldn't find correct type definition.
+  data?: any;
+  // biome-ignore lint/suspicious/noExplicitAny: couldn't find correct type definition.
```
BugGambit, line 236: Are these comments auto-generated 🤔 ?

BugGambit: E.g. `nonce` here should be a string

scripts/extractCodeSnippets.js
```diff
@@ -24,7 +24,14 @@ function findAllCodeSnippetsInJsDoc(jsDoc) {
...
+      } catch (e) {
+        console.error(
+          `Error in operation ${operationId} with code snippet: ${rawCode}, value: ${value}`
+        );
+      }
```
BugGambit:
```suggestion
      } catch (e) {
        console.error(
          `Error in operation ${operationId} with code snippet: ${rawCode}, value: ${value}`
        );
        throw e;
      }
```
I would throw the exception here, to not ignore the error. Now, we are giving more context about the problem, making it easier to debug.

packages/stable/src/api/functions/functionCallsApi.ts
BugGambit: I just tested and this works:
```ts
  /**
   * [List function calls](https://api-docs.cognite.com/20230101/tag/Function-calls/operation/postFunctionsCall)
   *
   * ```js
   * client.functions.call(123, { nonce: 'abc' } )
   * ```
   */
```
🤷

packages/stable/src/api/functions/types.gen.ts
bdimitrijoski: Heh, no, since I turned off the auto-generated types for this API I  can update the types, thanks :)

bdimitrijoski:
Closing this PR as priorities changed and I unfortunately I have to work on something else and don't have time to finish this PR

PR #1163, opened 2024-10-07
chore(lerna): aggressive patch bump

SudilHasithaCognite:
### Addressing Lerna [aggressive versioning](https://github.com/cognitedata/cognite-sdk-js/blob/release-v10/CONTRIBUTING.md#:~:text=We%20restrict%20new,of%20conventional%20commits)

> We restrict new npm releases to `[release]`-tagged commits because lerna is
> quite aggressive in its versioning. Changes to any file not ignored by Lerna will
> cause a PATCH bump. Markdown files and tests are ignored, but changing anything else,
> like a comment in a source file, will trigger a new version,
> irrespective of conventional commits.

## References

1. Lerna documentation: https://github.com/lerna/lerna/tree/main/libs/commands/version#--no-git-tag-version

## Modifications made

1. Create a dummy NPM package to verify the [release](https://github.com/SudilHasithaCognite/My_Dummy_JS_Versioning_Playground/actions/runs/11932615738/job/33257938054) logic to NPM registry.
2. Added aRC tag removal script.
3. Verify the [expected behavior](https://www.npmjs.com/package/@my-dummy-js-versioning-playground/header?activeTab=versions). 
4. Updated the CONTRIBUTIONS.md


.github/workflows/release.yaml
```diff
@@ -85,9 +85,16 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
           GITHUB_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
-        run: yarn lerna version --conventional-commits --conventional-prerelease --preid rc --create-release github --no-private --yes
-
+        run: |
+          if [[ "${{contains(github.event.commits[0].message,'[release]')}}" == "true" ]]; then
+            if [[ "${{contains(github.event.commits[0].message,'[bump-version]')}}" == "true" ]]; then
```
SudilHasithaCognite: [bump-version] can be replaced with suitable other keyword

.github/workflows/release.yaml
```diff
@@ -85,13 +90,27 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
           GITHUB_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
-        run: yarn lerna version --conventional-commits --conventional-prerelease --preid rc --create-release github --no-private --yes
-
+        run: |
+          if [[ "${{ github.ref }}" == "ref/head/master" ]]; then
+            if [[ "${{ contains(github.event.commits[0].message, '[bump-version]') }}" == "true"]]; then
+              yarn lerna version --conventional-commits --create-release github --no-private --yes
```
SudilHasithaCognite, line 113: is it okay to change to patch instead of --conventional-commits

EliasBjorne: Why do we want to change this? is there a big pain point where we get too many patch versions?

.github/workflows/release.yaml
```diff
@@ -85,13 +90,27 @@ jobs:
...
+            fi
+          elif [[ "${{ github.ref }}" == ref/head/release-*  ]]; then
+            if [[ "${{ contains(github.event.commits[0].message, '[bump-version]') }}" == "true"]]; then
```
EliasBjorne: What does this to, is this something we want?

EliasBjorne (commented):
Im not too familiar with this setup, but from what I see, it seems like this changes a bit of the release logic. We should then also update the 
[CONTRIBUTING Readme](https://github.com/cognitedata/cognite-sdk-js/blob/release-v10/CONTRIBUTING.md
)
so it mirrors these changes. Both regarding to the logic we should have when we remove 
--conventional-commits 

The logic around bump-verion etc. 

I'm not too opinionated on what the best release strategy is as long as it is well documented

.github/workflows/release.yaml
SudilHasithaCognite: When experimenting I saw that conventional commit sometimes bumps the major version, instead of the minor version. So to avoid that added the [bump-version] tag for intentionally increase the version number. Even though it may keep the minor or path-level version bump I thought to perform path-level version bumps using the master. So whenever there is a need for a major version bump we can use a separate release branch with conventional commits as the version strategy.

Since [bump-version] is there it is even ok to keep conventional-commits. if the intention is to release both minor and path version update via master branch

.github/workflows/release.yaml
SudilHasithaCognite: This checks whether the PR from the release-n branch and if the commit message has the tag named [bump-version] then it increases the version of the release candidate

SudilHasithaCognite: The purpose is to keep a constant version in a release branch (for rc) by combining it with [release] tag in the publish logic.

.github/workflows/release.yaml
SudilHasithaCognite: https://github.com/cognitedata/cognite-sdk-js/pull/1163#discussion_r1796984943

SudilHasithaCognite:
> Im not too familiar with this setup, but from what I see, it seems like this changes a bit of the release logic. We should then also update the [CONTRIBUTING Readme](https://github.com/cognitedata/cognite-sdk-js/blob/release-v10/CONTRIBUTING.md) so it mirrors these changes. Both regarding to the logic we should have when we remove --conventional-commits
> 
> The logic around bump-verion etc.
> 
> I'm not too opinionated on what the best release strategy is as long as it is well documented

Updated the contributing guidelines.

CONTRIBUTING.md
```diff
@@ -80,18 +80,53 @@ CI/CD will run tests, and if successful, do deploys.
...
+contain `[release]` and `[bump-version]` in its description and the PR title starts with `feat` or `fix`.
+
+#### Examples
+
+1. If you want to release a release candidate version to npm with version bump when you are commiting to a release-* branch.
+```bash
+  feat(dataParser): add xml support [release][bump-version]  
+  fix(graphUtils): circular path detection [release][bump-version] 
+```
+2. If you want to release a release candidate version to npm without version bump when you are commiting to a release-* branch.
```
EliasBjorne: Why would you want to do this? In what scenario do you want to release a change but not have at least a patch bump? and Would you ever do a feat release and not bump the version?

CONTRIBUTING.md
```diff
@@ -80,18 +80,53 @@ CI/CD will run tests, and if successful, do deploys.
...
+  fix(graphUtils): circular path detection [release]
+```
+
+3. If you want to release a npm verison using master branch
+```bash
+  feat(dataParser): add xml support [release]
+  fix(graphUtils): circular path detection [release]
+```
+
+4. If you want to release a version bumped package from master branch
```
EliasBjorne: Not sure I understand this sentence

CONTRIBUTING.md
```diff
@@ -80,18 +80,53 @@ CI/CD will run tests, and if successful, do deploys.
...
+```bash
+  feat(dataParser): add xml support [release][bump-version]  
+  fix(graphUtils): circular path detection [release][bump-version]
+```
+
+5. If you want to release a package to npm from master by merging the release branch to it while keeping the major version number in release condidate as it is.
+- Create a PR from release-* branch to master
```
EliasBjorne: What is the usecase for this? I think the most tidy thing is to always push things to master, and then cherry pick fixes from master to release branches. Do we want to do this? @BugGambit

EliasBjorne (commented):
I think we should add some details on how we want to work with release branches. In other repos I have worked with, all fixes are merged into master, and then fixes are cherry picked from master to release branches. We should have some documentation on how we want to do this?

CONTRIBUTING.md
SudilHasithaCognite: If I understood correctly, this was the main concern that needed to be addressed according to the previous meeting. 
It is such that if we make small changes in the release branch (such as test case adjustment and merge then it will update the version number, which was unintentional (if my understanding is correct).

Your concern here is valid since someone mistakenly did merge a significant change with [release] tag which is intentionally need to adhere to the versioning convention then it won't reflect without [bump-verison] tag.

CONTRIBUTING.md
SudilHasithaCognite: Sorry, my wording makes no sense here. This is to release patches and minor version updates to the stable release

CONTRIBUTING.md
EliasBjorne: but would you not use test( ) if you are updated testing? Maybe we should include this as examples? As updating tests are not patch as far as I know

CONTRIBUTING.md
```diff
@@ -166,72 +172,4 @@ git push && git push --tags
 ```
 
 This will make a commit with the updated `package.json`, create a new git tag, and publish to npm.
-Make sure you are logged in to npm, talk to a maintainer.
-
-## Code overview
```
BugGambit: I wonder why the "Code overview" section is removed in the PR?

.github/workflows/release.yaml
```diff
@@ -81,17 +83,51 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
-        run: yarn lerna version --conventional-commits --conventional-prerelease --preid rc --create-release github --no-private --yes
+        run: |
+          if [[ "${GITHUB_REF}" == "refs/heads/master" ]]; then
+            EVENT_PATH_CONTENT=$(cat ${GITHUB_EVENT_PATH} | jq -r '.')
+            HEAD_COMMIT_MSG=$(echo $EVENT_PATH_CONTENT | jq -r '.head_commit.message')
+            MERGE_COMMIT_MSG=$(echo $EVENT_PATH_CONTENT | jq -r '.commits[-1].message')
+
+            if [[ "$HEAD_COMMIT_MSG" == *"Merge"* && "$HEAD_COMMIT_MSG" == *"release-"* && "$HEAD_COMMIT_MSG" == *"[release]"* ]] ||
+               [[ "$MERGE_COMMIT_MSG" == *"Merge"* && "$MERGE_COMMIT_MSG" == *"release-"* && "$MERGE_COMMIT_MSG" == *"[release]"* ]]; then
```
BugGambit, line 108: These checks are a bit hard to read. What's the intention?

scripts/remove-rc-tags.js
BugGambit, line 1: Nice

BugGambit (approved):

PR #1165, opened 2024-10-21
fix: add instance id support to timeseries API

roligheten:
* Add support to query for timeseries by instance ID on relevant endpoints
* Add instance ID to response type for timeseries and timeseries datapoints
* Add instance ID to core SDK type

### DIFF
--- a/packages/stable/src/__tests__/api/datapoints.int.spec.ts
+++ b/packages/stable/src/__tests__/api/datapoints.int.spec.ts
@@ -2,12 +2,42 @@
 
 import { afterAll, beforeAll, describe, expect, test } from 'vitest';
 import type CogniteClient from '../../cogniteClient';
-import type { DatapointAggregate, Timeseries } from '../../types';
-import { setupLoggedInClient } from '../testUtils';
+import type { DatapointAggregate, NodeWrite, Timeseries } from '../../types';
+import { randomInt, setupLoggedInClient } from '../testUtils';
 
 describe('Datapoints integration test', () => {
   let client: CogniteClient;
   let timeserie: Timeseries;
+
+  const testSpace = {
+    space: 'test_data_space',
+    name: 'test_data_space',
+    description: 'Instance space used for integration tests.',
+  };
+
+  const timeseriesCdmInstance: NodeWrite = {
+    externalId: `external_${randomInt()}`,
+    space: testSpace.space,
+    instanceType: 'node',
+    sources: [
+      {
+        source: {
+          externalId: 'CogniteTimeSeries',
+          space: 'cdf_cdm',
+          type: 'view',
+          version: 'v1',
+        },
+        properties: {
+          type: 'numeric',
+        },
+      },
+    ],
+  };
+
+  const timeseriesCdmInstanceId = {
+    externalId: timeseriesCdmInstance.externalId,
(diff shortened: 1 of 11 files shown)


packages/stable/src/__tests__/api/datapoints.int.spec.ts
```diff
@@ -25,9 +55,20 @@ describe('Datapoints integration test', () => {
...
     client = setupLoggedInClient();
     [timeserie] = await client.timeseries.create([{ name: 'tmp' }]);
+    await client.spaces.upsert([testSpace]);
+    await client.instances.upsert({
+      items: [timeseriesCdmInstance],
+    });
   });
   afterAll(async () => {
     await client.timeseries.delete([{ id: timeserie.id }]);
+    await client.instances.delete([
```
BugGambit, line 65: One downside here is that a failure/exception will cause the instances to not be cleaned up in the CDF project.
We added some clean up here: https://github.com/cognitedata/cognite-sdk-js/pull/1083/files#diff-8bd79ffb443e5a0606897f789e00907501054c1bd8bd81a6295edaa5625d0cedR51

packages/stable/src/types.ts
```diff
@@ -662,7 +664,8 @@ export interface DatapointsDeleteRange {
 
 export type DatapointsDeleteRequest =
   | (InternalId & DatapointsDeleteRange)
-  | (ExternalId & DatapointsDeleteRange);
+  | (ExternalId & DatapointsDeleteRange)
+  | (InstanceId & DatapointsDeleteRange);
```
BugGambit, line 668: This is a duplicate of line 666?

packages/stable/src/types.ts
```diff
@@ -1404,7 +1429,8 @@ export type LIST = 'LIST';
 
 export type LatestDataBeforeRequest =
   | (InternalId & LatestDataPropertyFilter)
-  | (ExternalId & LatestDataPropertyFilter);
+  | (ExternalId & LatestDataPropertyFilter)
+  | (InstanceId & LatestDataPropertyFilter);
```
BugGambit, line 1433: This is a duplicate of line 1431?

packages/stable/src/types.ts
roligheten: Not sure what you mean: line 666 is InternalId and 668 is InstanceId I think?

packages/stable/src/types.ts
roligheten: Same comment as above

BugGambit: ah! I read it too fast

BugGambit (approved):

packages/stable/src/__tests__/api/datapoints.int.spec.ts
roligheten: I think afterAll will only not run after tests if there was some unexpected failure on jest side? (I guess also if the timeseries delete throws) so not sure how important a clean up is right now (I think maybe we can see if this becomes a problem and then introduce a cleanup?)

Also just a comment that in DM tests we try to avoid deleting/recreating spaces and containers because it bloats the elastic index, so we for example do not clean up spaces after tests on purpose. Instances however should be cleaned up I agree. :slightly_smiling_face:

PR #1166, opened 2024-10-23
fix: add instance id support to files API

roligheten:
* Add support to query for files by instance ID on relevant methods
* Add instance ID to response type for files methods

### DIFF
--- a/packages/stable/src/__tests__/api/files.int.spec.ts
+++ b/packages/stable/src/__tests__/api/files.int.spec.ts
@@ -10,6 +10,7 @@ import type {
   FileGeoLocation,
   FileInfo,
   LabelDefinition,
+  NodeWrite,
 } from '../../types';
 import {
   getFileCreateArgs,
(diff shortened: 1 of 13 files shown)


packages/stable/src/__tests__/api/files.int.spec.ts
```diff
@@ -20,11 +21,39 @@ import {
 
 const testfile = join(__dirname, '../test3dFile.fbx');
 
-describe.skip('Files integration test', () => {
```
roligheten, line 23: Note these tests have been disabled since 2021, I am just un-disabling the suite so I can write my own tests and disabling all of the existing tests individually. Only diff is that it now runs at least the integration tests for instance ids but still no test for the rest of the files API, this needs to be addressed in a separate task

packages/stable/src/types.ts
```diff
@@ -1064,14 +1064,30 @@ export interface FileChange {
   };
 }
 
+export interface FileChangeByInstanceId {
```
BugGambit, line 1072: This contains duplicated field from https://github.com/cognitedata/cognite-sdk-js/pull/1166/files#diff-a1cb34e0ba200ad9a54dfae19192d251fe8573729eea371b18dacdbf2850ab44R1051-R1064
I suggest extracting the common fields out

packages/stable/src/__tests__/api/files.int.spec.ts
```diff
@@ -78,30 +118,63 @@ describe.skip('Files integration test', () => {
...
+        instanceId: fileCdmInstanceId,
+        update: { metadata: { set: testMetadata } },
+      },
+    ]);
+    expect(retrievedFile.instanceId).toEqual(fileCdmInstanceId);
+    expect(retrievedFile.metadata).toEqual(testMetadata);
+  });
+
+  test('download by instance id', async () => {
+    try {
```
BugGambit: Rewrite this to use jest's `rejects`: https://jestjs.io/docs/tutorial-async#rejects

roligheten: Done

packages/stable/src/types.ts
roligheten: @BugGambit done

BugGambit (approved):

PR #1177, opened 2024-12-09
feat(stable): add support for sessions API

polomani:
sessions have been there for ages, but the SDK was never prioritized.
adding it now directly to the stable package

### DIFF
--- a/packages/stable/src/api/sessions/types.ts
+++ b/packages/stable/src/api/sessions/types.ts
@@ -33,7 +33,7 @@ export const SessionType = {
 is short-lived and should be immediately passed to the endpoint that will use
 the session.
 */
-export interface SessionCreateResponseItem {
+export interface SessionCreateResultItem {
   /** ID of the session */
   id: number;
   /** Client ID in identity provider. Returned only if the session was created using client credentials */
--- a/packages/stable/src/api/sessions/types.ts
+++ b/packages/stable/src/api/sessions/types.ts
@@ -1,0 +1,107 @@
+import type { EpochTimestamp, FilterQuery } from '../../types';
+
+export type SessionStatus =
+  | 'READY'
+  | 'ACTIVE'
+  | 'CANCELLED'
+  | 'EXPIRED'
+  | 'REVOKED'
+  | 'ACCESS_LOST';
+
+export const SessionStatus = {
+  READY: 'READY' as SessionStatus,
+  ACTIVE: 'ACTIVE' as SessionStatus,
+  CANCELLED: 'CANCELLED' as SessionStatus,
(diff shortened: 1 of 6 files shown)


packages/stable/src/api/sessions/types.ts
```diff
@@ -0,0 +1,107 @@
...
+  | 'ACCESS_LOST';
+
+export const SessionStatus = {
+  READY: 'READY' as SessionStatus,
+  ACTIVE: 'ACTIVE' as SessionStatus,
+  CANCELLED: 'CANCELLED' as SessionStatus,
+  EXPIRED: 'EXPIRED' as SessionStatus,
+  REVOKED: 'REVOKED' as SessionStatus,
+  ACCESS_LOST: 'ACCESS_LOST' as SessionStatus,
+};
```
BugGambit, line 18:
```suggestion
export const SessionStatus: Record<string, SessionStatus> = {
  READY: 'READY',
  ACTIVE: 'ACTIVE',
  CANCELLED: 'CANCELLED',
  EXPIRED: 'EXPIRED',
  REVOKED: 'REVOKED',
  ACCESS_LOST: 'ACCESS_LOST',
};
```

packages/stable/src/api/sessions/types.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export type SessionType =
+  | 'CLIENT_CREDENTIALS'
+  | 'TOKEN_EXCHANGE'
+  | 'ONESHOT_TOKEN_EXCHANGE';
+
+export const SessionType = {
```
BugGambit, line 25: Same here

BugGambit (approved):

packages/stable/src/api/sessions/types.ts
polomani: `Record<string, SessionStatus>` doesn't give a nice dev experience, though.
E.g. autocomplete 1=record, 2=my version

Screenshot 2024-12-11 at 16 20 37

Screenshot 2024-12-11 at 16 21 02

polomani: `union type SessionStatus + const SessionStatus` is quite a nice abstraction to go for instead of the classic enums, as it gives best of both worlds.

polomani: we could ofc use, `Record<SessionStatus, SessionStatus>` but it still not quite the same...

polomani: we could ofc use, `Record<SessionStatus, SessionStatus>` but it still not quite the same...

BugGambit (approved):

PR #1181, opened 2025-01-03
feat: update autogenerated types and add inspect endpoint

roligheten:
* Add missing codegen config for instances API
* Update all autogenerated types
* Add inspect method to instances API


roligheten:
Will remake against master

PR #1187, opened 2025-01-17
fix(vite): adding common js as fallback [release]

notrixbe:
Update : 
- adding cjs
- update exports 

Reminder 
Now we use the default common plugin 
if issues we could use 
const commonjs = require('vite-plugin-commonjs'); // Import CommonJS plugin


notrixbe:
@BugGambit @EliasBjorne 
the discussed change. 
should add a common js in the build and export

EliasBjorne (approved):
LGTM

BugGambit (approved):
LGTM

PR #1189, opened 2025-01-20
feat(stable): add 3d asset mappings with core dm support in API

danpriori:
Adding support for 3d asset mappings with Cognite Asset (Core DM) within the Asset Mappings type definition and including flag support for requesting asset mappings via Cognite Asset (Core DM).

Following the 3d API (the doc updates not merged yet, it seems. But the API is already updated):

https://pr-2743.specs.preview.cogniteapp.com/20230101.json.html#tag/3D-Asset-Mapping/operation/get3DMappings
https://pr-2743.specs.preview.cogniteapp.com/20230101.json.html#tag/3D-Asset-Mapping/operation/filter3DAssetMappings

### DIFF
--- a/packages/stable/src/types.ts
+++ b/packages/stable/src/types.ts
@@ -407,7 +407,7 @@ export interface AssetMappings3DListFilter extends FilterQuery {
    */
   intersectsBoundingBox?: BoundingBox3D;
   /**
-   * If true, the response will include the mappings with assetInstanceId values, for DMS based assets.
+   * If true, the response will include the mappings with assetInstanceId values, for DM based assets.
    */
   getDmsInstances?: boolean;
 }
--- a/packages/stable/src/types.ts
+++ b/packages/stable/src/types.ts
@@ -393,6 +393,10 @@ export interface AssetMapping3DBase {
    * The ID of the associated asset (Cognite's Assets API).
    */
   assetId: CogniteInternalId;
+  /**
+   * The ID of the associated Cognite Asset instance from Core Data Model.
+   */
+  assetInstanceId?: UnitDMSUniqueIdentifier,
 }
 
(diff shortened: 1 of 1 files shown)


packages/stable/src/types.ts
```diff
@@ -402,6 +406,10 @@ export interface AssetMappings3DListFilter extends FilterQuery {
    * If given, only return asset mappings for assets whose bounding box intersects the given bounding box.
    */
   intersectsBoundingBox?: BoundingBox3D;
+  /**
+   * If true, the response will include the mappings with assetInstanceId values, for DMS based assets.
```
haakonflatval-cognite:
```suggestion
   * If true, the response will include the mappings with assetInstanceId values, for DM based assets.
```

packages/stable/src/types.ts
```diff
@@ -3269,6 +3277,11 @@ export interface UnitSystem {
   quantities: UnitSystemQuantity[];
 }
 
+export interface UnitDMSUniqueIdentifier {
+  space: string;
+  externalId: string;
+}
+
```
haakonflatval-cognite, line 3284: Is there really no other type like this in the SDK from before? (Might be true, just checking :sweat_smile: )

haakonflatval-cognite: Also, I wonder if we should document this type too

haakonflatval-cognite (commented):
A few suggestions, but otherwise looks good :+1:

packages/stable/src/types.ts
danpriori: I searched for unit type like this in the SDK but I didnt find it. Do you have any suggestion @BugGambit ?

BugGambit: I couldn't find it either...

BugGambit (approved):

PR #1192, opened 2025-01-20
fix(package): adding core export [release]

notrixbe:
adding core vite config
- copied dist in powerops and seems to fix it.

### DIFF
--- a/packages/beta/src/__tests__/api/filesMultiPart.unit.spec.ts
+++ b/packages/beta/src/__tests__/api/filesMultiPart.unit.spec.ts
@@ -1,7 +1,7 @@
 import { sleepPromise } from '@cognite/sdk-core';
-import { mockBaseUrl } from '../../../../core/src/__tests__/testUtils';
 import nock from 'nock';
 import { beforeEach, describe, expect, it, test } from 'vitest';
+import { mockBaseUrl } from '../../../../core/src/__tests__/testUtils';
 import type { FilesMultipartUploadSessionAPI } from '../../api/files/filesMultipartUploadSessionApi';
 import type CogniteClient from '../../cogniteClient';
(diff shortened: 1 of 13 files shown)


notrixbe:
expect tests fails :)

packages/core/package.json
```diff
@@ -6,6 +6,11 @@
   "main": "dist/index.js",
   "module": "dist/index.js",
   "types": "dist/index.d.ts",
+  "exports": {
```
notrixbe, line 9: export breaks links between tests

`import { randomInt } from '@cognite/sdk-core/src/__tests__/testUtils';

but export fix the core error in powerops

packages/beta/src/__tests__/api/filesMultiPart.unit.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { sleepPromise } from '@cognite/sdk-core';
-import { mockBaseUrl } from '@cognite/sdk-core/src/__tests__/testUtils';
```
notrixbe, line 2: only by changing import have been able to make the test work. 
so need to check for extra export config 

For info 
@BugGambit

packages/core/package.json
notrixbe: fixed with changing imports

BugGambit (commented):
~I don't like that we access the files directly. Think about the `@cognite/sdk-core` as an API. We should access its methods from the API interface. Now, we bypass the interface and access methods directly.
The sdk-core doesn't control breaking changes anymore as other code rely on file locations.~

@notrixbe nvm, I believe this change doesn't change the fundamentals from what existed before. We still accessed the files directly from SDK core

BugGambit (approved):

PR #1243, opened 2025-02-02
feat: release v10 [release]

BugGambit:
This will release new SDK versions for stable, alpha, beta, and core.

New versions will be:
- @cognite/sdk -> 10.0.0
- @cognite/sdk-alpha -> 1.0.0
- @cognite/sdk-beta -> 6.0.0
- @cognite/sdk-core -> 5.0.0


.github/workflows/release.yaml
```diff
@@ -31,16 +31,18 @@ jobs:
...
-        !contains(github.event.commits[0].message, '[skip-ci]') &&
-        (
-          startsWith(github.event.commits[0].message, 'fix:') ||
-          startsWith(github.event.commits[0].message, 'fix(') ||
-          startsWith(github.event.commits[0].message, 'feat:') ||
-          startsWith(github.event.commits[0].message, 'feat(') ||
-          contains(github.event.commits[0].message, '[release]')
+    if: ${{ !contains(github.event.head_commit.message, '[skip-ci]') &&
+      (
+        (github.ref == 'refs/heads/master' && contains(github.event.head_commit.message, '[release]')) ||
```
EliasBjorne, line 36: what does this do?

.github/workflows/release.yaml
```diff
@@ -81,21 +83,55 @@ jobs:
...
+            yarn lerna publish from-package  --pre-dist-tag rc --yes --no-git-reset --no-verify-access
+          fi
 
-      - name: Publish documentation
-        uses: JamesIves/github-pages-deploy-action@releases/v3
-        with:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          BRANCH: gh-pages # The branch the action should deploy to.
-          FOLDER: docs
+      # - name: Publish documentation
```
EliasBjorne, line 132: is this so we wait with pushing documentation until we know it is working?

.github/workflows/release.yaml
BugGambit: It should to the step if:

1. The last commit message doesn't include the substring `[skip-ci]`
2. Last commit message contains the substring `[release]` or we or on the release branch and the commit prefix it one of `fix:`, `fix(`, `feat:`, `feat(` or contains the substring `[release]`

packages/beta/package.json
```diff
@@ -3,26 +3,33 @@
...
+    "esCheck": "yarn g:es-check es6 --module './dist/index.js'",
+    "docs": "yarn g:typedoc --options ./typedoc.cjs --tsconfig tsconfig.build.json src/index.ts",
     "docs:bundle": "yarn docs && mkdir -p ../../docs/beta && cp -r docs/* ../../docs/beta/",
     "extract-snippets": "rm -rf codeSnippets/ && yarn docs --json codeSnippets/docs.json && node ../../scripts/extractCodeSnippets.js --package-name @cognite/sdk-beta",
     "test-snippets": "yarn extract-snippets && yarn g:tsc -p codeSnippets/tsconfig.build.json"
   },
   "dependencies": {
-    "@cognite/sdk": "^9.17.1",
-    "@cognite/sdk-core": "^4.10.7"
+    "@cognite/sdk": "^10.0.0-rc.10",
```
EliasBjorne, line 31: Should we keep this?

packages/stable/package.json
```diff
@@ -3,28 +3,35 @@
...
-  "version": "9.17.1",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    "import": "./dist/index.js",
+    "require": "./dist/index.cjs",
+    "types": "./dist/index.d.ts"
+  },
+  "version": "10.0.0-rc.10",
```
EliasBjorne, line 14: same here are we starting to use "version": "10.0.0-rc. here?

.github/workflows/release.yaml
BugGambit: It's essentially saying that to release from master, you must include `[release]` in the commit message, but this isn't needed when releasing from `release-*` branches through release candidates.

packages/stable/src/__tests__/api/files.int.spec.ts
```diff
@@ -71,30 +118,64 @@ describe.skip('Files integration test', () => {
...
+
+  test.skip('retrieve with non-existent id', async () => {
     const res = await client.files.retrieve([{ id: 1 }], {
       ignoreUnknownIds: true,
     });
     expect(res.length).toBe(0);
   });
 
-  test('count aggregate', async () => {
+  test.skip('count aggregate', async () => {
```
EliasBjorne, line 178: Why are we skipping tests ?

EliasBjorne (commented):
Added some coments. Are we switching over to using release candidates and not versions?

.github/workflows/release.yaml
EliasBjorne: I see, that makse sence!

.github/workflows/release.yaml
BugGambit: This was turned off to ensure our release candidates for v10 didn't override the production documentation. We will re-enable this after we merge this PR.

packages/beta/package.json
BugGambit: Yes, if things work correctly, then this step will clean up the `-rc.*` suffix: https://github.com/cognitedata/cognite-sdk-js/blob/f2ab5e994c85ddb226e20153985ae7aae74fa8fb/.github/workflows/release.yaml#L90-L95

packages/stable/package.json
BugGambit: Yes, if things work correctly, then this step will clean up the `-rc.*` suffix: https://github.com/cognitedata/cognite-sdk-js/blob/f2ab5e994c85ddb226e20153985ae7aae74fa8fb/.github/workflows/release.yaml#L90-L95

packages/stable/src/__tests__/api/files.int.spec.ts
BugGambit: Unstable / flaky test. Aggregation comes from elastic search, and it takes time to sync from postgres to elastic, causing flaky tests. We should have a plan for testing in the SDK moving forward. Maybe we should cut down the number of integration tests?

packages/stable/package.json
EliasBjorne: Nice!

EliasBjorne (approved):
Based on answers we should be good. Lets also be ready to unpublish the latest release in case it has some unwanted consequences

PR #1248, opened 2025-02-03
feat(annotations): shorter annotation types

BugGambit:
Shorten the long annotation types (`AnnotationsCogniteAnnotation...` to `Annotations...`).

Since these long names comes from the code generation tool, I had to manually shorten the names in the API spec copy we have in this repository.


packages/stable/src/exports.gen.ts
```diff
@@ -32,38 +26,16 @@ export type {
...
+  AnnotationsTypesDiagramsInstanceLink,
+  AnnotationsTypesImagesAssetLink,
+  AnnotationsTypesImagesInstanceLink,
+  AnnotationsTypesPrimitivesGeometry2DGeometry,
+  AnnotationsTypesPrimitivesGeometry3DGeometry,
   AnnotationsUnhandledSymbolObject,
   AnnotationsUnhandledTextObject,
   AnnotationsView,
 } from './api/annotations/types.gen';
-export type {
```
BugGambit, line 39: These types are moved to `types.ts` to not interfere with the code-generation.

packages/stable/src/types.ts
```diff
@@ -3292,6 +3292,27 @@ export interface UnitDMSUniqueIdentifier {
   externalId: string;
 }
 
+export type {
```
notrixbe, line 3295: question: 

with the move to types ?
do we also need the changes in index ?
because list of export is not equal to import

BugGambit: I didn’t get the concern. Can you give an example?

BugGambit: It’s only the internals of this SDK that imports this file. The users only import from `@cognite/SDK` so it doesn’t matter in which file these types live

notrixbe: import only : 
`AnnotationsTypesDiagramsAssetLink,
  AnnotationsTypesDiagramsInstanceLink,
  AnnotationsTypesImagesAssetLink,
  AnnotationsTypesImagesInstanceLink,
  AnnotationsTypesPrimitivesGeometry2DGeometry,
  AnnotationsTypesPrimitivesGeometry3DGeometry,`
  
  moved: 
  
  `BtreeIndex,
  ConstraintDefinition,
  ContainerCollectionResponse,
  ContainerCollectionResponseWithCursorResponse,
  ContainerCorePropertyDefinition,
  ContainerCreateCollection,
  ContainerCreateDefinition,
  ContainerDefinition,
  ContainerPropertyDefinition,
  IndexDefinition,
  InvertedIndex,
  ListOfContainerSubObjectIdentifierRequest,
  ListOfContainerSubObjectIdentifierResponse,
  ListOfSpaceExternalIdsRequest,
  ListOfSpaceExternalIdsResponse,
  RequiresConstraintDefinition,
  SpaceQueryParameter,
  UniquenessConstraintDefinition,`

notrixbe: see i mis see the imports

notrixbe: ok :) thanks for more info ! 
if only used internal

notrixbe (approved):
seems ok

PR #1268, opened 2025-02-06
chore(gen): remove deprecated swagger-typescript-api-nextgen

polomani:
swagger-typescript-api-nextgen (fork) was deprecated a long time ago, we need to replace it with the well-maintained swagger-typescript-api (helps with getting rid of CVEs)


polomani:
@andersfylling could you have a look if there is an easy way to fix failing tests?

andersfylling:
> now I am not sure why it was used instead of the original swagger-typescript-api

the maintainer just went afk for a long while, so made the switch. Seems it's active again :pray:

andersfylling:
what had happened was that the updated library had some changes to formatting and supports more openapi features, so we had to update the expected output files to match this. Some of the tests will generate some typescript definitions from openapi yaml, and match that against an expected output file (which we now updated to include the newer features).

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -325,16 +428,12 @@ export type DocumentFilterValueList = DocumentFilterValue[];
  */
 export interface DocumentGeoJsonGeometry {
   /** Coordinates of the geometry. */
-  coordinates?:
-    | number[]
-    | number[][]
-    | number[][][]
-    | (number[] & number[][] & number[][][]);
+  coordinates?: number[] | number[][] | number[][][];
```
andersfylling, line 431: wait, is this a breaking change? could a typescript expect confirm?

packages/stable/src/api/vision/types.gen.ts
```diff
@@ -55,77 +72,134 @@ export interface AnnotationsBoundingBox {
...
 export interface AnnotationsKeypoint {
-  /** Additional attributes data for a compound. */
-  attributes?: Record<string, AnnotationsBoolean | AnnotationsNumerical>;
+  /**
+   * Additional attributes data for a compound.
+   * @default null
+   */
+  attributes?: Record<
+    string,
+    | ({
```
andersfylling, line 86: this looks kind of weird?

andersfylling:
@marvrez or @miladcognite can you verify the changes to vision?

andersfylling:
@peetcremer1 or @olacognite can you verify changes to annotations?

packages/stable/src/api/documents/types.gen.ts
polomani: the new definition is correct `number[] | number[][] | number[][][];`
the old one was faulty, it'd never be satisfied essencially

polomani: Screenshot 2025-02-07 at 09 46 18

packages/stable/src/api/vision/types.gen.ts
```diff
@@ -3,30 +3,47 @@
...
 export interface AnnotationsBoolean {
-  /** The description of a primitive */
+  /**
+   * The description of a primitive
+   * @maxLength 500
+   * @default null
+   */
   description?: string;
-  type: 'boolean';
+  type: boolean;
```
polomani, line 33: looks sketchy

packages/stable/src/api/vision/types.gen.ts
```diff
@@ -3,30 +3,47 @@
...
   id?: number;
+  /**
+   * The external ID of the referenced resource
+   * @maxLength 256
+   * @default null
+   */
   externalId?: string;
 };
 /**
+ * primitives.attributes.Boolean
```
polomani, line 23: do you know if we can remove these `primitives.attributes` annotations? they seem useless
@andersfylling

packages/stable/src/api/vision/types.gen.ts
polomani: string was correct here
Screenshot 2025-02-07 at 09 52 51

packages/stable/src/api/vision/types.gen.ts
polomani: agree this seems broken (initially was correct)

olacognite (approved):
LG annotation-wise

BugGambit (approved):

packages/stable/src/api/vision/types.gen.ts
polomani: @andersfylling can we override this type somehow?

andersfylling: I'll have a look again later today

PR #1272, opened 2025-02-08
chore(deps): update husky to major version 9 and lerna major version to v7

roligheten:
Think this will resolve https://github.com/cognitedata/cognite-sdk-js/security/dependabot/331 and https://github.com/cognitedata/cognite-sdk-js/security/dependabot/463 and https://github.com/cognitedata/cognite-sdk-js/security/dependabot/328

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,4 @@
 {
-  "useWorkspaces": true,
   "version": "independent",
   "npmClient": "yarn",
   "command": {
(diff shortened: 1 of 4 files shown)


lerna.json
```diff
@@ -1,5 +1,4 @@
 {
-  "useWorkspaces": true,
```
roligheten, line 2: This flag is deprecated, lerna assumes all workspaces are used now

BugGambit (approved):

PR #1274, opened 2025-02-26
feat(stable): added datamodelacl and dataModelInstancesacl to group type

EliasBjorne:
Added dataModelAcl and dataModelInstancesAcl to the Group type, as this was not available in the SDK


packages/stable/src/types.ts
```diff
@@ -83,6 +87,11 @@ export type AclAssets = Acl<AclActionAssets, AclScopeAssets>;
 
 export type AclDataSets = Acl<AclActionDataSets, AclScopeDatasets>;
 
+export type AclDataModel = Acl<AclDataModel, AclScopeDataModel>;
+
+
```
EliasBjorne:
```suggestion
```

packages/stable/src/types.ts
```diff
@@ -55,6 +55,10 @@ export type AclActionAssets = READ | WRITE;
 
 export type AclActionDataSets = READ | WRITE | OWNER;
 
+export type AclActionDataModelInstances = READ | WRITE;
```
stefanmaretic: Is this missing WRITE_PROPERTIES -- https://github.com/cognitedata/infrastructure/blob/master/lib/proto/api_acls/cogmono/api_acls/datamodelinstances_acl/datamodelinstances_acl.proto#L14?

packages/stable/src/types.ts
```diff
@@ -55,6 +55,10 @@ export type AclActionAssets = READ | WRITE;
...
+
+export type AclActionDataModel = READ | WRITE | WRITE_PROPERTIES;
```
stefanmaretic: To me it looks like data models acl doesn't have `WRITE_PROPERTIES` -- https://github.com/cognitedata/infrastructure/blob/master/lib/proto/api_acls/cogmono/api_acls/datamodels_acl/datamodels_acl.proto#L12

packages/stable/src/types.ts
```diff
@@ -55,6 +55,10 @@ export type AclActionAssets = READ | WRITE;
...
+
```
EliasBjorne:
```suggestion
export type AclActionDataModelInstances = READ | WRITE | WRITE_PROPERTIES;

export type AclActionDataModel = READ | WRITE;

```

packages/stable/src/types.ts
EliasBjorne: seems like I mixed those two up

stefanmaretic (approved):
lgtm

BugGambit (approved):
LGTM, @EliasBjorne let's pay attention when merging to ensure it will bump the correct version (minor - `0.X.0`)

BugGambit:
:unicorn:

EliasBjorne:
it did not release it seems. It could be because we need to have [RELEASE in our title]

PR #1275, opened 2025-03-01
fix(stable): regenerated-instances-types

EliasBjorne:
Added `packages/stable/src/api/instances/codegen.json` file, and ran: 

```
yarn codegen generate-types --package=stable
yarn lint:fix
```

Also the  `instancesApi.ts` was updated to using types  based on the openAPI spec

After a bit back and forth I had to disable the codegen for instances by renaming it to instances/codegen.disable.json. The reason I had to disable it was that the codegen produces a `any` type, which the linter did not like. Switching from any to unknown was ok for the linter, so we should look into how we can make the codegen use `unknown` instead of `any` so that we can have the codegen enabeled


packages/stable/src/__tests__/api/instances.int.spec.ts
```diff
@@ -47,22 +71,93 @@ const upsertDescribables = async (
...
+        },
+      ],
+    })),
+  });
+};
+
 describe('Instances integration test', () => {
   let client: CogniteClient;
   const timestamp = Date.now();
+  // @cognite/sdk: Unknown Error: The value of property 'sourceUpdatedTime' must be a string in the format 'YYYY-MM-DDTHH:MM:SS[.millis][Z|time zone]' with optional milliseconds having precision of 1-3 decimal digits and optional timezone with format ±HH:MM, ±HHMM, ±HH or Z, where Z represents UTC, Year must be between 0001 and 9999. Got value '1740839823145'.
```
EliasBjorne:
```suggestion
```

packages/stable/src/api/instances/types.gen.ts
```diff
@@ -416,12 +464,6 @@ export interface MinAggregateFunctionV3 {
  * The cursor value used to return (paginate to) the next page of results, when more data is available.
  */
 export type NextCursorV3 = string;
-export interface NodeAndEdgeCollectionResponseV3Response {
```
BugGambit, line 419: Breaking change?

packages/stable/src/api/instances/types.gen.ts
```diff
@@ -1775,17 +1907,27 @@ export type TypePropertyDefinition = ViewCorePropertyDefinition;
...
+  unitSystemName: string;
+}
 export interface UpsertConflict {
   /** Details about the error caused by the upsert/update. */
   error: {
     code: number;
     message: string;
   };
 }
-export type ViewAggregationRequest = CommonAggregationRequest & {
```
BugGambit, line 1785: Breaking change?

packages/stable/src/api/instances/codegen.disable.json
BugGambit, line 1: filename: `codegen.disable.json`?

EliasBjorne: As the code gen makes the lining fail it has to be disabled. alternatively, I added this PR: 
where the autogeneration is enableled here: 
https://github.com/cognitedata/cognite-sdk-js/pull/1276

packages/stable/src/api/instances/types.gen.ts
EliasBjorne: yes, but it is based on the OpenAPI doc, so not sure it is breaking the change here also. how do we handle this?

EliasBjorne: how do we deal with this? Do we add the types that are removed, and deprecate them? can we do that already or do we need to wait to next major release to depricate these types?

EliasBjorne: added them back and marked them as depricated

packages/stable/src/api/instances/codegen.disable.json
EliasBjorne: as the codegen will overwrite what we have breaking change we need to diable it, we could also remove it if you think its confusing

haakonflatval-cognite:
There is an instance query filter I'd like to have included here, which I don't see yet. It's the `instanceReference` filter: https://api-docs.cognite.com/20230101/tag/Instances/operation/queryContent#!path=with/result-expression-name*/0/nodes/filter/1/11/instanceReferences&t=request

EliasBjorne:
Will wait on this PR: 
https://github.com/cognitedata/cognite-sdk-js/pull/1279

And then regenerate after this

PR #1276, opened 2025-03-02
fix(stable): regenerate instances types automaticly

EliasBjorne:
Same as https://github.com/cognitedata/cognite-sdk-js/pull/1275, but allows autogeneration by downgrading lint error to lint waring when any tipe is used, as the codegen produces any types


EliasBjorne:
Autogenerate creates breakning changes so should remove from that until next major release

PR #1279, opened 2025-03-04
feat(stable): generate latest types (04.03.25)

deep-cognite:
Run `yarn codegen fetch-latest --package stable` and `yarn codegen generate-types --package stable`; affected services: documents, spaces, userProfiles, annotations and vision

NOTE: Due to breaking changes in the schema path for "vision" and "annotations" (see thread here: https://github.com/cognitedata/cognite-sdk-js/pull/1120), we have to alter the openapi snapshot to avoid breaking changes to the type names:
image


packages/stable/src/api/annotations/types.gen.ts
```diff
@@ -38,7 +38,7 @@ export type AnnotationsAssetRef = {
 export interface AnnotationsBoolean {
   /** The description of a primitive */
   description?: string;
-  type: 'boolean';
```
BugGambit: This is strictly speaking a breaking change... Is `'boolean'` actually the correct type here?

packages/stable/src/api/annotations/types.gen.ts
```diff
@@ -364,7 +364,7 @@ export interface AnnotationsLine {
 export interface AnnotationsNumerical {
   /** The description of a primitive */
   description?: string;
-  type: 'numerical';
```
BugGambit: Same here.

packages/stable/src/api/annotations/types.gen.ts
```diff
@@ -608,7 +627,7 @@ export interface AnnotationsView {
    * @pattern ^[a-zA-Z][a-zA-Z0-9_-]{0,41}[a-zA-Z0-9]?$
    */
   space: string;
-  type: 'view';
```
BugGambit: Same here

BugGambit (commented):
See comments. Also the PR title is wrong. This is a new feature and should be `feat(stabel): ...`

EliasBjorne (commented):
Can you go bit into detail what was altered from the Open API spec? Should we also depricate the changes that we need to add manually, as they should be removed eventually?

deep-cognite:
> Can you go bit into detail what was altered from the Open API spec?

The steps I took was 1) (as seen in screenshot above) remove all new references of `cogmono__annotations` due to the changes committed in the schema path in service contract (seen here https://github.com/cognitedata/service-contracts/commit/191ddb3b200091afa65a3adf9ba2482ff11aa7ed); and 2) update the annotation types to include 'enum' to resolve the comments by Fredrik (not sure why that part broke, either the it was manually changed every time before or the annotation SDK has changed...) 

> Should we also depricate the changes that we need to add manually, as they should be removed eventually?

Which part will eventually be removed? 🤔

BugGambit (approved):

EliasBjorne:
> > Should we also depreciate the changes that we need to add manually, as they should be removed eventually?
> 
> Which part will eventually be removed? 🤔

My thinking was, if we need to do some manual changes, should these changes be deprecated, as we don't want to do manual changes into the perpetuity? But I see most of the changes were renaming, so maybe not relevant here

EliasBjorne:
I think you need to add [Release] to the commit for the change to be released. https://github.com/cognitedata/cognite-sdk-js/blob/master/CONTRIBUTING.md#releases--versioning

deep-cognite:
> I think you need to add [Release] to the commit for the change to be released. [`master`/CONTRIBUTING.md#releases--versioning](https://github.com/cognitedata/cognite-sdk-js/blob/master/CONTRIBUTING.md#releases--versioning)

Thanks for the heads-up! I believe we have to merge the PR before cutting a release, or?
> "Releases are done from the master branch, so when a pull request is merged, CI/CD will run tests, and if successful, do deploys. Documentation is built and deployed, and code snippets are exported to the service contract repo as a pull request."

simenl (approved):
risk looks ok to me. Only additions. Adding a second reviewer as part of risk review onboarding

larshagencognite (approved):
risk review ok

marvrez (approved):

1991sig (approved):
LGTM

PR #1284, opened 2025-04-28
fix(CD): add seperate pr for lerna versions (potential fix)[skip-ci]

SudilHasithaCognite:
### Addressing Branch Protection (Potential Fix)

- Create a separate PR instead of pushing it directly to the main.
- Caution: Did not test and use code generation tools
- PR Trigger: https://github.com/cognitedata/cognite-sdk-js/actions/runs/14705891969/job/41266174409

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,6 +6,9 @@ on:
       - master
       - 'release-*'
 
+permissions:
+  contents: write
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -24,7 +27,7 @@ jobs:
           path: node_modules
           key: ${{ runner.OS }}-build-${{ hashFiles('package.json') }}
 
-      - name: Install 🔧 # Installs dependencies
+      - name: Install 🔧
         run: yarn install --immutable
 
       - name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -64,74 +66,119 @@ jobs:
...
+            --force-publish \
+            --no-push
+
+      - name: Create bump PR
+        if: github.ref == 'refs/heads/master'
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.GH_LERNA_TOKEN }}
+          commit-message: "chore(release): bump versions [skip ci]"
+          branch: "release/bumps/${{ github.sha }}"
```
polomani: can't be called "release*" as those are protected branches

.github/workflows/release.yaml
```diff
@@ -24,34 +27,33 @@ jobs:
...
-          startsWith(github.event.head_commit.message, 'feat:') ||
-          startsWith(github.event.head_commit.message, 'feat(') ||
-          contains(github.event.head_commit.message, '[release]')
+          (
+            (github.ref == 'refs/heads/master' && contains(github.event.head_commit.message, '[release]')) ||
+            (startsWith(github.ref, 'refs/heads/release-') &&
+              (
+                startsWith(github.event.head_commit.message, 'fix:') ||
+                startsWith(github.event.head_commit.message, 'feat:') ||
+                contains(github.event.head_commit.message, '[release]')
```
polomani: some extra lines touched here. e.g. `startsWith(github.event.head_commit.message, 'feat('` should stay

.github/workflows/release.yaml
```diff
@@ -64,74 +66,119 @@ jobs:
...
-          GH_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
-          GITHUB_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
+      - name: Create remove-rc PR
+        id: create-remove-rc-pr
+        if: >-
+          github.ref == 'refs/heads/master' &&
+          contains(github.event.head_commit.message, 'Merge pull request') &&
+          contains(github.event.head_commit.message, 'release-') &&
+          contains(github.event.head_commit.message, '[release]')
+        uses: peter-evans/create-pull-request@v5
```
polomani: not sure about this part at all, we need someone who wrote the RC tag removal logic in a first place. maybe we don't need it anymore?

polomani (commented):
Maybe @thorkildcognite have some thoughs around this one.
Overall the change seems to be quite huge, and there might be some edge cases, for instance what happens if someone merges something to master in between? The version logic will then be messed up.

So CI/CD must have a safety check that prevents that from happening, if we want to do this automatic PR route

.github/workflows/release.yaml
SudilHasithaCognite: can't we create release branch via workflow PR?

.github/workflows/release.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - master
       - 'release-*'
 
+permissions:
+  contents: write
+  pull-requests: write
```
SudilHasithaCognite, line 11: to allow create PRs via workflow

polomani:
we made a similar change and it is on master now!
Thanks for the idea!

PR #1287, opened 2025-06-17
feat: replace fixed HTTP request retry delay with "exponential jitter delay"

larsmoa:
### What & Why

This PR replaces the fixed retry delay mechanism in HTTP requests with an **exponential backoff with full jitter** algorithm, improving the resilience and performance of API requests under high load conditions.

Benefits:
- **Prevents thundering herd**: Jitter distributes retry attempts across time
- **Maintains familiar timing**: Average delays are comparable to old fixed delays
- **Controlled growth**: Maximum 15s delay vs unbounded growth in old system

### Algorithm Details

The implementation is from [AWS Blog: Exponential Backoff and Jitter](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/):

- **Formula**: `random() * min(baseDelay * 2^(retryCount + 1), maxDelay)`
- **Base delay**: 250ms
- **Max delay**: 15,000ms  
- **Jitter**: Full jitter (0 to calculated delay)

See also [talk by @alexbrasetvik](https://cognitedata.slack.com/files/UMWJU1413/F087PJCPA5Q/three_dots_menu_caused_error.mov).

Closes [AAA-3215]

#### **Retry Delay Comparison**

| Attempt | **Old delay**        | **New delay (exponential jitter)**     |
|---------|------------------------|-----------------------------------|
| 0       | 250ms (fixed)          | 0-500ms (avg: 250ms)             |
| 1       | 750ms (fixed)          | 0-1,000ms (avg: 500ms)           |
| 2       | 1,750ms (fixed)        | 0-2,000ms (avg: 1,000ms)         |
| 3       | 3,750ms (fixed)        | 0-4,000ms (avg: 2,000ms)         |
| 4       | 7,750ms (fixed)        | 0-8,000ms (avg: 4,000ms)         |
| 5       | 15,750ms (fixed)       | 0-15,000ms (avg: 7,500ms)        |

### 🔗 **References**

- Source discussion: https://cognitedata.slack.com/archives/C6KNJCEEA/p1734633903875829

[AAA-3215]: https://cognitedata.atlassian.net/browse/AAA-3215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/httpClient/exponentialJitterBackoff.ts
+++ b/packages/core/src/httpClient/exponentialJitterBackoff.ts
@@ -14,13 +14,13 @@ export class ExponentialJitterBackoff {
 
   /**
    * Initializes a new instance of the ExponentialJitterBackoff class.
-   * @param baseDelayMs - The base delay in milliseconds. This will be the average delay for the first retry attempt.
-   * @param maxDelayMs - The maximum delay in milliseconds. This will be the maximum delay for the last retry attempt.
+   * @param baseDelayMs - The base delay in milliseconds. This will be the average delay for the first retry attempt. Defaults to 250ms.
+   * @param maxDelayMs - The maximum delay in milliseconds. This will be the maximum delay for the last retry attempt. Defaults to 15000ms.
    * @param randomCallback - The callback to use for generating a random number between 0 and 1. Used by test only.
    */
   constructor(
-    baseDelayMs: number = 100,
-    maxDelayMs: number = 5000,
+    baseDelayMs: number = 250,
+    maxDelayMs: number = 15000,
     randomCallback: typeof Math.random = Math.random,
   ) {
     this.baseDelayMs = baseDelayMs;
--- a/packages/core/src/httpClient/exponentialJitterBackoff.ts
+++ b/packages/core/src/httpClient/exponentialJitterBackoff.ts
@@ -1,12 +1,12 @@
-  /**
-   * Calculates the delay in milliseconds for the next retry attempt. Uses "exponential backoff with 
-   * full jitter" algorithm.
-   * 
-   * @see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
-   * 
-   * @param retryCount - The number of retry attempts made so far.
-   * @returns The delay in milliseconds for the next retry attempt.
-   */
+/**
+ * Calculates the delay in milliseconds for the next retry attempt. Uses "exponential backoff with
+ * full jitter" algorithm.
+ *
+ * @see https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
+ *
+ * @param retryCount - The number of retry attempts made so far.
+ * @returns The delay in milliseconds for the next retry attempt.
+ */
 export class ExponentialJitterBackoff {
   private readonly baseDelayMs: number;
   private readonly maxDelayMs: number;
@@ -19,9 +19,9 @@ export class ExponentialJitterBackoff {
    * @param randomCallback - The callback to use for generating a random number between 0 and 1. Used by test only.
    */
   constructor(
-    baseDelayMs: number = 250,
-    maxDelayMs: number = 15000,
-    randomCallback: typeof Math.random = Math.random,
+    baseDelayMs = 250,
+    maxDelayMs = 15000,
+    randomCallback: typeof Math.random = Math.random
   ) {
     this.baseDelayMs = baseDelayMs;
     this.maxDelayMs = maxDelayMs;
@@ -29,7 +29,7 @@ export class ExponentialJitterBackoff {
   }
 
   public calculateDelayInMs(retryCount: number) {
-    const exponentialDelay = this.baseDelayMs * (2 ** (retryCount + 1));
+    const exponentialDelay = this.baseDelayMs * 2 ** (retryCount + 1);
     const cappedDelay = Math.min(exponentialDelay, this.maxDelayMs);
 
     return Math.floor(this.randomCallback() * cappedDelay);
--- a/packages/core/src/httpClient/exponentialJitterBackoff.ts
+++ b/packages/core/src/httpClient/exponentialJitterBackoff.ts
@@ -1,0 +1,37 @@
+  /**
+   * Calculates the delay in milliseconds for the next retry attempt. Uses "exponential backoff with 
+   * full jitter" algorithm.
+   * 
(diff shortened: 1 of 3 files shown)


roligheten:
I happened to be working on jittering too :slightly_smiling_face:  made a simulator to simulate different retry strategies in setting as similar to the JS SDK and added your retry strategy as a test case 
image
Generally seems like any strategy is better than the current implementation, a decorrelated strategy seems slightly better than the equal jitter strat, but not by that much.

Main point was to prove that the new retry strategy actually is better than the old one, which it seems it is. Do I think it is worth implemeneting the decorrelated strategy? Not necessarily.

For reference this is the decorr retry strat:
```
class DecorrJitterRetryStrategy implements RetryStrategy {
    lastSleepTime = 125
    getWaitTime(retryCount: number): number {
        const newSleep = Math.min(4500, 125+Math.random()*this.lastSleepTime*4)
        this.lastSleepTime = newSleep
        return newSleep
    }
}
```

packages/core/src/httpClient/retryableHttpClient.ts
```diff
@@ -38,8 +39,8 @@ import { MAX_RETRY_ATTEMPTS, type RetryValidator } from './retryValidator';
  */
 export class RetryableHttpClient extends BasicHttpClient {
   private static calculateRetryDelayInMs(retryCount: number) {
-    const INITIAL_RETRY_DELAY_IN_MS = 250;
-    return INITIAL_RETRY_DELAY_IN_MS + ((2 ** retryCount - 1) / 2) * 1000;
+    const backoff = new ExponentialJitterBackoff();
```
roligheten: Does this need to be a class? It seems a bit wasteful to instantiate a new class every time we need to calculate the backoff given that the calculation is has no state. :thinking:

larsmoa: I don't have a strong preference about class vs no class. The reason I added it as a class is that it enabled me to allow configuration of parameters without littering `calculateDelayInMs` with optional arguments. I use this in tests.

I'm not really worried about the instantiation of an instance though. The overhead associated with this is probably only academically interesting compared to the resources associated with the actual request. An alternative might be to move the instance outside the function, or maybe even using a singleton.

packages/core/src/httpClient/exponentialJitterBackoff.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.baseDelayMs = baseDelayMs;
+    this.maxDelayMs = maxDelayMs;
+    this.randomCallback = randomCallback;
+  }
+
+  public calculateDelayInMs(retryCount: number) {
+    const exponentialDelay = this.baseDelayMs * 2 ** (retryCount + 1);
+    const cappedDelay = Math.min(exponentialDelay, this.maxDelayMs);
+
+    return Math.floor(this.randomCallback() * cappedDelay);
```
roligheten, line 41: Maybe we should consider a min cap also? Retrying immediately seems intuitively like it would not succeed, especially in the relatively common case a lot of frontends do where they will try to make a lot of request as the page loads. In this case if we assume the project gets maxed out by requests that just initiated (which is fairly easy as the max read concurrency can be as low as 3) it is unlikely that any of the other requests have finished yet if we retry immediately . Maybe a lower cap around 150ms?

packages/core/src/httpClient/retryableHttpClient.ts
roligheten: Agree it is not meaningful difference here, was more wondering whether this should be a class vs a pure function conceptually, and whether it should be instantiated on the client level. I can see the argument for it being a class though to encapsulate the configuration, and instantiating it in the class vs calculate method is just a nit comment :slightly_smiling_face:

packages/core/src/httpClient/exponentialJitterBackoff.ts
larsmoa: I think that might be a bad idea, because that would essentially chunk several requests together at the cap (e.g. 150 ms), potentially causing under-utilization between 0 and "cap ms" and request storm at "cap ms".

The current approach without min cap is [how its done for the Python SDK](https://github.com/cognitedata/cognite-sdk-python/blob/d38294581d13cb475e96b12f23d825dbbb1e6fb2/cognite/client/_http_client.py#L77-L83) and how the underlying article describes this so I think we want to be careful about changing this.

@alexbrasetvik your thoughts on this are welcome.

roligheten: Yes the chunking at the min cap is a concern. We can do `MIN_CAP + this.randomCallback() * (cappedDelay - MIN_CAP)` to avoid the chunking at the bottom while mostly maintaining the distribution.

roligheten: Note I tried running the simulation with vs without a min cap and it doesn't not seem to make a significant difference.

alexbrasetvik: It is very common that a single request fails, and having a very low or even 0 retry for the first request is fine.

A _default_ max retry of ~60 seconds might be better than 15, it's a bit app specific. 

I do not feel like we need to out-science the findings from AWS here. https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

The implementation here of that has served us very well. https://github.com/cognitedata/pgpgpg/blob/main/pgpgpg/util/backoff.py#L28

larsmoa: I agree. I'm open to adjusting the defaults. I do think that 60s max makes more sense in a Python setting where many workloads will run in batch/"offline" jobs, but the JS SDK will typically be used in UIs where 60 seconds is too long

packages/core/src/httpClient/retryableHttpClient.ts
BugGambit: We initialize a new class for each call to `calculateRetryDelayInMs`?
Should we reuse the class instance (which seems safe to do) or convert the method to a static method?

larsmoa: Fixed in d989059119a39fa3f965d26a18bd3eff7631088a

BugGambit (approved):

BugGambit:
risk review ok

PR #1289, opened 2025-07-02
chore(codegen): fix various codegen regression due to spec changes

roligheten:
Fix various codegen regression due to spec changes
* Correct "CogmonoAnnotation" prefix automatically (This has been
  manually resolved in recent type updates)
* Automatically transform "consts" into "enums" in the OpenAPI schema.
  Before service-contracts repo used to do this transformation in the
  schema which made consts behave like enums to ensure OpenAPI v2 spec
  complience. After a recent update it seems like it no longer does
  this, and since our codegen rely on the v2 spec it now generates the
  wrong type for "consts" in the OpenAPI spec.

Also generate newest types, these types were generated without any manual intervention as was needed before (Except linting)


packages/codegen/src/codegen.ts
```diff
@@ -50,6 +54,52 @@ export interface CodeGenOptions extends AutoNameInlinedRequestOption {
...
 }
 
+/**
+ * Due to service-contracts having removed backwards compatibility
+ * with OpenAPI v2 spec, and the swagger-typescript-api-nextgen
+ * lacking support for at least the "const" concept introduced in
+ * the OpenAPI v3 spec, we need to manually transform consts into
+ * enums such that the type gen will correctly handle const properties.
+ */
+const hotfixConstsAsEnums = (component: SchemaComponent) => {
```
roligheten, line 64: Tbh I think this should be resolved by upgrading to a newer version of swagger-typescript-api (If one exists that supports consts) We are currently using "swagger-typescript-api-nextgen" which seems to have been archived, not sure why we are using this fork instead of "swagger-typescript-api" or how hard it would be to migrate.

Tried switching to swagger-typescript-api-nextgen and immedietly got build errors so think it won't be trivial. 

@andersfylling looks like you worked on this originally, any expert insight? :slightly_smiling_face:

packages/codegen/src/codegen.ts
```diff
@@ -50,6 +54,52 @@ export interface CodeGenOptions extends AutoNameInlinedRequestOption {
...
+    );
+
+    component.rawTypeData.properties = patchedProperties;
+  }
+
+  return component;
+};
+
+const formatProblematicServiceContractTypeNames = (typeName: string) => {
+  return typeName.replace('CogmonoAnnotation', '');
```
roligheten, line 95: Context: https://cognitedata.slack.com/archives/C06SHGQ8LAH/p1728387245037219

packages/codegen/src/codegen.ts
polomani: Fyi we should move to the maintained version of the generator.
I started the work but didn't have time to fix the issues:
https://github.com/cognitedata/cognite-sdk-js/pull/1268

andersfylling: right, we fixed up a bunch there and then I kind of forgot about it. I'm going on vacation tomorrow, so I may not have much time to spare today - but I'll make it a priority when I'm back next week.

BugGambit (approved):
LGTM

BugGambit:
@roligheten ref the PR title "fix(codegen): fix various codegen regression due to spec changes"

Two things:
1. "fix" is repeated twice. No need for the 2nd "fix".
2. This isn't a "fix". Yes, it is a "fix" from the SDK maintainers' point of view, but not for the SDK user. There is no bug they have this PR fixes. So the correct thing to use here is "chore"

roligheten:
:unicorn:

PR #1290, opened 2025-07-02
chore(codegen): persist updated types in package root and add editorconfig

roligheten:
Changes to the root types.gen.ts were not being saved when running type generation. Includes updated types based on this issue being fixed.

Also add .editorconfig for better IDE formatting support.

Also worked on fixing some flaky tests, see [1b5279d](https://github.com/cognitedata/cognite-sdk-js/pull/1290/commits/1b5279d53999154e4f59daa02a61e97aafaa9dee) and [fe60163](https://github.com/cognitedata/cognite-sdk-js/pull/1290/commits/fe60163db4aeec2e98e7852382783a0f72ee0304)


packages/codegen/src/codegen.ts
```diff
@@ -414,6 +414,7 @@ export class CodeGen {
     const docJson = JSON.stringify(doc);
     const result = await this.generator.generateTypes(docJson, CODE_GEN_HOOKS);
     result.astProcessedCode = this.astPostProcessing(result.code);
+    await this.saveGeneratedTypes(result);
```
roligheten, line 417: Basically I believe this function was in the wrong place, originally this was called from generateTypes function, which calls generateTypesFromSchemas, however some other code paths (for root type generation) did not go through generateTypes but calls generateTypesFromSchemas from another function, meaning generateTypesFromSchemas is the correct place to call this to ensure generated types are always being saved imo

danlevings (dismissed):

danlevings:
🦄

danlevings (approved):

danlevings:
🦄

PR #1305, opened 2025-09-04
chore: manual release process

danlevings:
# 🔒 Implement Simple Manual Release Process with Security Fixes

Example release PR created from release command: https://github.com/cognitedata/cognite-sdk-js/pull/1307

**Problem**: The existing automated release workflow used `lerna version` commands that pushed commits directly to the `master` branch, bypassing code review processes and branch protection rules. The new Policy bot makes this fail, as every commit to master requires approval.

## 📋 Summary

This PR implements a streamlined manual release process that eliminates direct master branch commits while maintaining simplicity and reliability. The new process requires only 3 steps:

1. **Release**: `yarn release` (creates PR automatically)
2. **Review**: Review and approve the generated PR
3. **Publish**: Merge PR → packages automatically publish to NPM

### New Files Added

- **`scripts/create-release-pr.js`** - Automated script that creates release PRs
- **`.github/workflows/publish-on-merge.yaml`** - New workflow for safe publishing
- **`docs/MANUAL_RELEASE_DESIGN.md`** - Complete documentation of new process

### Files Modified

- **`package.json`** - Added `yarn release` command
- **`lerna.json`** - Security fixes: removed master from allowBranch, added push/gitTagVersion protections
- **`.github/workflows/release.yaml`** - Disabled old workflow (renamed to `.disabled`)

### For Release Managers

First time only:
```bash
brew install gh
gh auth login
```

Then:
```bash
# Simple 3-step process
git checkout master && git pull origin master
yarn release
# → Review PR → Merge → Auto-publish ✨
```

### What Happens Automatically

1. **`yarn release`** calculates versions and creates PR with changes
2. **PR merge** triggers publishing workflow
3. **`lerna publish from-package`** publishes to NPM and creates git tags

### Lerna Configuration Security Fixes

```json
{
  "allowBranch": ["release/*"], // Removed "master"
  "push": false, // Prevents automatic pushing
  "gitTagVersion": false // Prevents automatic tagging
}
```

## 📚 Documentation

Complete implementation guide available in `docs/MANUAL_RELEASE_DESIGN.md` including:

- Step-by-step usage instructions
- Security configuration requirements
- Emergency procedures
- Troubleshooting guide

## ⚠️ Required Setup (Post-Merge)

**Enable Branch Protection**:
   - Require PR reviews for master branch
   - Prevent direct pushes (including from administrators)


scripts/create-release-pr.js
```diff
@@ -0,0 +1,40 @@
...
+    // Commit changes
+    execSync("git add .");
+    execSync('git commit -m "chore(release): version bump [skip ci]"');
+
+    // Push branch
+    execSync(`git push origin ${branchName}`);
+
+    // Create PR
+    execSync(
+      `gh pr create --title "Release: Version Bump" --body "Automated version bump for release. Merging this PR will publish packages to NPM." --base master --head ${branchName}`
```
polomani: what happens if someone pushes a commit to master while this is being under review.
we need a CI check that prevent this from making an outdated version bump / changelog.

perhaps CI check should run `lerna version` in dry mode and fail PR if it's outdated?

danlevings: Have adjusted the design doc + PR to accomodate for this. We'll need to also make the relevant github security adjustments also 👍

polomani:
PR description is a halucination `Critical Security Issue Resolved ... Risk Level: HIGH `, there is no security issue as of now. Nothing can be pushed to master.

PR title `fix: ` is wrong, there is user-facing fixes. It should be `chore: ` instead.

lerna.json
```diff
@@ -9,8 +9,10 @@
         "**/*.spec.ts",
         "**/*.test.ts"
       ],
-      "message": "chore(release): publish new package versions [skip ci]",
-      "allowBranch": ["master", "release-*"]
+      "message": "chore(release): version bump [skip ci]",
```
polomani: nit: better to keep the original message I think because there are multiple packages, not a single version bump

.github/workflows/validate-release-pr.yaml
```diff
@@ -0,0 +1,25 @@
...
+    if: startsWith(github.head_ref, 'release/automated-')
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check if release PR is up-to-date
+        run: |
+          COMMITS_BEHIND=$(git rev-list --count HEAD..origin/master)
```
polomani: no, this is a very weak check, because one can do a merge commit from master and that will make it succeed. 
you need to run lerna version on master I think, same as in create-release-pr, and compare with this branch

documentation/MANUAL_RELEASE_DESIGN.md
```diff
@@ -0,0 +1,227 @@
...
+A streamlined release process that eliminates security risks while maintaining simplicity:
+
+1. **Release Manager**: `yarn release` (creates PR automatically)
+2. **Review**: Review and approve the generated PR
+3. **Publish**: Merge PR → packages automatically publish to NPM
+
+## Security Issue Resolved
+
+**Problem**: The previous automated workflow used `lerna version` commands that pushed commits directly to the `master` branch, bypassing code review and branch protection rules.
+
```
polomani: if they weren't blocked, it could've been a security issue (but they are since feb anyway)

.github/workflows/validate-release-pr.yaml
danlevings: understood the issue, the proposed solution seems cumbersome to me. how about ensuring that the release branch was created from the current master HEAD? That seems more simple to manage and solves the problem

documentation/MANUAL_RELEASE_DESIGN.md
danlevings: yes the AI went hard on considering this a big security issue. I didn't correct it, thought it would make it take it seriously
but will adjust the wording now!

.github/workflows/validate-release-pr.yaml
danlevings: I also imagine that the real security comes from @cognitedata/javascript-sdk-maintainers as reviewers - double checking that the release PR is up to date. I'm not sure we should add too high complexity systems to maintain, atleast immediately.

scripts/create-release-pr.js
```diff
@@ -0,0 +1,43 @@
...
+const { execSync } = require('node:child_process');
+
+async function createReleasePR() {
+  try {
+    // Create release branch first (with timestamp for uniqueness)
+    const timestamp = new Date()
+      .toISOString()
+      .replace(/[:.]/g, '-')
+      .slice(0, 19);
+    const branchName = `release/automated-${timestamp}`;
```
polomani, line 56: `release/` branches are usually protected. maybe `automated-version-bump-${date}`.

.github/workflows/validate-release-pr.yaml
polomani: It's not much more complex than this check.
The current logic is wrong and will let outdated/incorrect versioning and changelogs

polomani: another way of solving it is perhaps run lerna version on the current branch again and see what the output is.
it might print out something like "no version bump needed" or smth

.github/workflows/publish-on-merge.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+      - name: Build packages
+        run: yarn build
+
+      - name: Configure NPM
+        run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> .npmrc
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
+
+      - name: Publish to NPM and create tags
```
polomani: I don't think the tags will be pushed because they are normally created on `lerna version` and now it's disabled `"gitTagVersion": false`.

So we have to tag manually here

.github/workflows/validate-release-pr.yaml
```diff
@@ -0,0 +1,42 @@
...
+          fetch-depth: 0
+
+      - name: Validate release branch base
+        run: |
+          # Get the merge-base (common ancestor) of this branch and master
+          MERGE_BASE=$(git merge-base HEAD origin/master)
+          MASTER_HEAD=$(git rev-parse origin/master)
+
+          # Check if this branch was created from current master HEAD
+          if [ "$MERGE_BASE" != "$MASTER_HEAD" ]; then
```
danlevings, line 23: @polomani (redirecting comment to latest)
I'm not sure how this could let in incorrect/invalid setups, hoping to understand more 🤔 

We ensure PRs must be up to date with master before merge via github. 

This action will check that the merge base of this branch and the HEAD of master are equal, meaning if this PR stems from something else it'll fail. (merge-base finds the common ancestor) 
We even, somewhat accidentally, add protection on release branches. 

That means that if the branch is out of date with master, you cant even push a merge commit to fix that. and because of the 'branches must be up to date before merging' setup, that PR becomes impossible to merge in, forcing a new one. Infact, I suppose with just those two things, we dont even need this check!

Or maybe Im missing something obvious that you can help me understand?

scripts/create-release-pr.js
danlevings: noted async: creating release branches tend to be allowed. once created, then they become write protected

danlevings: so we're probably fine having these be 'release' branches, feels like it adds a bit more security

.github/workflows/publish-on-merge.yaml
danlevings: hmm true, maybe we turn on gitTagVersion again? i think the risk review process might not prevent us from tagging, and the tagging seems nice

scripts/create-release-pr.js
polomani: yep

.github/workflows/validate-release-pr.yaml
polomani: yeah, this one seems to be not that bad. agree

.github/workflows/publish-on-merge.yaml
polomani: should only tag when merged to master, otherwise this would probably be misleading (tagged even before being approved)
`gitTagVersion` will tag as soon as you run the command, lerna says you have to tag yourself othewise

polomani (commented):
Did some sanity check and I believe this is a viable solution for now.
One thing to fix (maybe in the next PRs) is to reimplement the git tagging feature

.github/workflows/validate-release-pr.yaml
```diff
@@ -0,0 +1,42 @@
+name: Validate Release PR
+
+on:
+  pull_request:
+    branches: [master]
+
+jobs:
+  validate-release:
```
roligheten, line 8: Afaik unless a job is manually made required in repo settings they will not block the PR from being merged if they have not finished or failed. It would be preferable if this was a required workflow to merge as otherwise i am worried people will merge without it having finished or failed due to human error.

I think we may not be able to mark it as required though as we still want to be able to merge non-release branches.

Possible suggestion: Mark it required and always run the job on all PRs to master, but instantly pass it if it not a release PR.

danlevings: Gemini is telling me this
> No, you will not be blocked from merging pull requests if the if statement fails.

> When a job in a GitHub Actions workflow has an if condition that evaluates to false, the job is skipped. For the purpose of required status checks on a protected branch, GitHub treats a skipped check as neutral—it is neither a success nor a failure. Therefore, it will not block a PR from being merged.

I'm actually not sure if its correct or not. but we can definitely add the workflow to the required checks and find out! worst case we quickly throw in another PR to do as you suggest, which for sure should work

.github/workflows/release.yaml
```diff
@@ -1,137 +0,0 @@
...
-          echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" >> .npmrc
-        env:
-          NPM_TOKEN: ${{secrets.NPM_PUBLISH_TOKEN}}
-          GITHUB_TOKEN: ${{secrets.GH_LERNA_TOKEN}} # Temporary fix to make lerna able to push the new versions commit to master
-
-      - name: Build
-        run: yarn build
-
-      - name: Build doc
-        run: yarn docs:bundle
```
roligheten, line 84: I am not really sure what this does tbh, but I guess now it will not run this. Seems like some kind of experiment to bundle docs in alpha version? @polomani you seem to have worked on this do you think we can remove it?

.github/workflows/release.yaml
```diff
@@ -1,137 +0,0 @@
...
-          GITHUB_TOKEN: ${{ secrets.GH_LERNA_TOKEN }} # Temporary fix to make lerna able to push the new versions commit to master
-        run: |
-          if [[ "${GITHUB_REF}" == "refs/heads/master" ]]; then
-            EVENT_PATH_CONTENT=$(cat ${GITHUB_EVENT_PATH} | jq -r '.')
-            HEAD_COMMIT_MSG=$(echo $EVENT_PATH_CONTENT | jq -r '.head_commit.message')
-            MERGE_COMMIT_MSG=$(echo $EVENT_PATH_CONTENT | jq -r '.commits[-1].message')
-
-            if [[ "$HEAD_COMMIT_MSG" == *"Merge"* && "$HEAD_COMMIT_MSG" == *"release-"* && "$HEAD_COMMIT_MSG" == *"[release]"* ]] ||
-               [[ "$MERGE_COMMIT_MSG" == *"Merge"* && "$MERGE_COMMIT_MSG" == *"release-"* && "$MERGE_COMMIT_MSG" == *"[release]"* ]]; then
-              echo "Running major version bump for release merge"
```
roligheten, line 109: We lose the ability to do major versions with this setup, maybe we can merge this now to enable minor version bumps but I think we need a way to do major versions. 

I guess we could do this via a setting to the createReleasePR script?

scripts/create-release-pr.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+
+const { execSync } = require('node:child_process');
+
+async function createReleasePR() {
+  try {
```
roligheten, line 6: This is kind of dangerous I discovered: I had the .env file open in vim which creates a .env.swp which is not in .gitignore, this got auto added by the "git add" command which luckily triggered a secret push rule violation.

Additionally I can totally see a user failing to release and ending up in a weird state where they end up with multiple commits or similar and rerunning the release command.

Suggest a "pre-flight check" that ensures the user is in the correct state before making a release

```suggestion
  try {
      // Pre-flight checks
    console.log('Running pre-flight checks...');
    
    // 1. Fetch latest changes from remote
    execSync('git fetch origin', { stdio: 'inherit' });
    
    // 2. Check that we're on master branch
    const currentBranch = execSync('git rev-parse --abbrev-ref HEAD', { encoding: 'utf8' }).trim();
    if (currentBranch !== 'master') {
      console.error(`❌ Expected to be on 'master' branch, but currently on '${currentBranch}'. Please switch to master branch first.`);
      process.exit(1);
    }
    
    // 3. Check that we're on the latest commit of remote master
    const localCommit = execSync('git rev-parse HEAD', { encoding: 'utf8' }).trim();
    const remoteCommit = execSync('git rev-parse origin/master', { encoding: 'utf8' }).trim();
    if (localCommit !== remoteCommit) {
      console.error('❌ Local master is not up to date with origin/master. Please pull the latest changes first.');
      process.exit(1);
    }
    
    // 4. Check that working directory is clean
    const gitStatus = execSync('git status --porcelain', { encoding: 'utf8' }).trim();
    if (gitStatus.length > 0) {
      console.error(`❌ Working directory is not clean. Please commit or stash your changes first.\nUnstaged/staged files:\n${gitStatus}`);
      process.exit(1);
    }
    
    console.log('🎉 All pre-flight checks passed!\n');

```

roligheten:
/gemini review

:shrug: no gemini i guess

.github/workflows/release.yaml
polomani: Someone has disabled docs publish step in this PR: https://github.com/cognitedata/cognite-sdk-js/pull/1243/files#r1939318273
It needs to reenabled, and `yarn docs:bundle` should be added to publish-on-merge.yaml


@BugGambit it was forgotten, right?

.github/workflows/release.yaml
polomani: idk what this code is for, but lerna will generate major bump if `BREAKING CHANGE:` is mentioned in the changelog

.github/workflows/publish-on-merge.yaml
danlevings: I'll update the text, can just start with no tagging for now

.github/workflows/release.yaml
danlevings: hm, if its currently disabled, suggest we leave it disabled in this PR for now. but the team taking over should look into this

scripts/create-release-pr.js
danlevings: interesting! added your suggestions. maybe it would just be better to let the user make the PR themselves for safety even, but we can try the auto creation for now. striving for the best devx!

.github/workflows/release.yaml
danlevings: yes, hopefully the script will be able to figure this out itself. can leave it as it is for now and try!

.github/workflows/validate-release-pr.yaml
roligheten: Lets test after this PR is merged and add it if it is easy to skip the check

roligheten (dismissed):

roligheten:
:unicorn:

polomani (dismissed):

CONTRIBUTING.md
```diff
@@ -75,50 +75,60 @@ Once the change is pushed to the master branch, it is time for a release.
...
+- ✅ **Clear audit trail** - Every release documented in PR history
+- ✅ **Branch protection compatible** - Works with strict branch protection rules
 
-In order to perform a major release by merging a release candidate branch to master by keeping the same major version in release candidate version.
+### Prerequisites
 
-- Create a PR from release-* to master
-- Make a PR title and the message when merging to be ` chore(): details [release]`
-- The release tag is to trigger the release of the package.
+- Install GitHub CLI: `brew install gh`
```
polomani: remove this

roligheten: done

.github/workflows/publish-on-merge.yaml
```diff
@@ -0,0 +1,34 @@
+name: Publish on Release PR Merge
+
+on:
+  push:
+    branches: [master]
+    paths: ['packages/*/package.json']
+
+jobs:
+  publish:
+    if: contains(github.event.head_commit.message, 'chore(release): version bump')
```
polomani:
```suggestion
    if: contains(github.event.head_commit.message, 'chore(release): publish new package versions')
```
it was renamed

scripts/create-release-pr.js
```diff
@@ -0,0 +1,90 @@
...
+    // Calculate versions without pushing (now on the release branch)
+    console.log('Calculating version changes...');
+    execSync(
+      'lerna version --conventional-commits --no-push --no-git-tag-version --yes',
+      { stdio: 'inherit' }
+    );
+
+    // Commit changes
+    execSync('git add .');
+    execSync('git commit -m "chore(release): version bump [skip ci]"');
```
polomani:
```suggestion
    execSync('git commit -m "chore(release): publish new package versions"');
```
we should not skip ci anymore (that was a previous workflow)

CONTRIBUTING.md
```diff
@@ -75,50 +75,59 @@ Once the change is pushed to the master branch, it is time for a release.
...
+### Emergency Releases
+
+Use the same process - no exceptions. The security model ensures all releases are reviewable:
+
+```bash
+yarn release  # Creates emergency release PR
+# Review → Merge → Auto-publish
+```
+
+See `docs/MANUAL_RELEASE_DESIGN.md` for complete implementation details.
```
polomani: this doesn't exist anymore

CONTRIBUTING.md
roligheten: done

polomani (approved):

roligheten:
:unicorn:

PR #1306, opened 2025-09-04
fix: update types from codegen [release]

haakonflatval-cognite:
I have updated the types in the stable package by running these commands:
```
$ yarn codegen fetch-latest --package stable
$ yarn codegen generate-types --package stable
```
from the root of the repository.

I am personally only interested in the changes in the point cloud bounding volume annotation type, but many other things have seemingly been changed since that


haakonflatval-cognite:
I understand that this SDK is currently facing challenges with respect to new releases. Will close this PR and rely on future updates/type codegen PRs to do what I intended to do here

PR #1312, opened 2025-09-11
chore(release): publish new package versions

polomani:
using the new workflow to release the packages, changelog included


polomani:
need to remove templates first

PR #1320, opened 2025-09-17
chore(release): fix broken release workflow

roligheten:
* Fix : in if condition not being escaped, which caused the file to not parse
* Move if condition to job level because there is no point performing the other actions of "build-and-publish" job currently without the publish step being performed


polomani (approved):
🦄

PR #1324, opened 2025-09-22
chore(release): create github release as part of release CI process

roligheten:
This modifies the tagging script to also make a github release by parsing the changelog, looking for the corresponding changelog entry for the version being released, and pushing the release to github


scripts/push-git-tag-and-release.js
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Create GitHub release using GitHub CLI
+ * @param {string} tagName - Git tag name
+ * @param {string} releaseNotes - Release notes content
+ */
+async function createGitHubRelease(tagName, releaseNotes) {
+    console.log(`Creating GitHub release: ${tagName}`);
+    
+    // Create the release using GitHub CLI
+    const releaseCommand = `gh release create "${tagName}" --title "${tagName}" --notes "${releaseNotes.replace(/"/g, '\\"')}"`;
```
polomani: can it not use the `gh` util? it's not installed

roligheten: The gh tool is pre-installed on all runners in Github actions :slightly_smiling_face:

scripts/push-git-tag-and-release.js
```diff
@@ -0,0 +1,104 @@
...
+  const startIndex = lines.findIndex((line) => versionPattern.test(line));
+
+  if (startIndex === -1) {
+    throw new Error(`Version ${version} not found in changelog`);
+  }
+
+  // Find the end of this version entry (next ## header)
+  let endIndex = lines.length;
+  for (let i = startIndex + 1; i < lines.length; i++) {
+    if (lines[i].startsWith('## [')) {
```
polomani: how about just splitting the file by `## [` and take the first entry?

roligheten: That would work, but in the context of the problem I am not sure if it is better? Ie. I need to find the index of the line that starts the next change log entry, so I can split out the relevant change log lines. Using split gives me a string including the rest of the "header line" after "## [" meaning I would have to do a follow up step to find where the actual next line starts. Sounds like it would be a more complicated way of doing what is done with the "iterate over lines" solution above :slightly_smiling_face:

polomani: perhaps, but I am afraid this parser won't work anyway, because changelog has different title size based on version change.
maybe just use the git diff between HEAD and last version for this file? or better - current commit should be "chore(release): publish new package versions", right? it already contains correct changelog diff for the given file. E.g. https://github.com/cognitedata/cognite-sdk-js/commit/47ed532234f19c7ef273b8168283123b9a22195c
Just need to extract lines that where added in the current commit for the `./CHANGELOG.md"


Screenshot 2025-09-24 at 09 31 44

danlevings: The pattern itself seems regex-able? `(#)# xx.x.x (xxxx-xx-xx)`

roligheten: Good catch! I think this is fixable by making a small change to the regex, will look at that

scripts/push-git-tag-and-release.js
```diff
@@ -0,0 +1,104 @@
...
+  // Create the release using GitHub CLI
+  const releaseCommand = `gh release create "${tagName}" --title "${tagName}" --notes "${releaseNotes.replace(/"/g, '\\"')}"`;
+  console.log(releaseCommand);
+  execSync(releaseCommand, { stdio: 'inherit' });
+
+  console.log(`✅ Successfully created GitHub release: ${tagName}`);
+}
+
+async function pushGitTagAndRelease() {
+  const packageName = process.env.npm_package_name;
```
danlevings, line 65: where does it get these vars from?

roligheten: they are autopopulated by the "npm publish" command. https://docs.npmjs.com/cli/v8/using-npm/scripts#packagejson-vars

scripts/push-git-tag-and-release.js
roligheten: Fixed

danlevings (approved):
🦄

danlevings (approved):
🦄

PR #1333, opened 2025-11-03
feat(simint-alpha): add simulatorExternalIds filter support for routines

mohammad-arif662:
This PR is to add support for filtering routines by `simulatorExternalIds` in the SDK package. This allows to filter simulator routines using simulator external IDs alongside existing filters like `modelExternalIds` and `simulatorIntegrationExternalIds`.

### DIFF
--- a/packages/alpha/src/__tests__/api/simulatorRoutinesApi.spec.ts
+++ b/packages/alpha/src/__tests__/api/simulatorRoutinesApi.spec.ts
@@ -112,6 +112,39 @@ describeIf('simulator routines api', () => {
     expect(routineFound?.externalId).toBe(routineExternalId);
   });
 
+  test('list routines with simulatorExternalIds filter', async () => {
+    const listFilterResponse = await client.simulators.listRoutines({
+      filter: { simulatorExternalIds: [simulatorExternalId] },
+    });
+    expect(listFilterResponse.items.length).toBeGreaterThan(0);
+    const routineFound = listFilterResponse.items.find(
+      (item) => item.externalId === routineExternalId
+    );
+    expect(routineFound?.externalId).toBe(routineExternalId);
+  });
+
+  test('list routines with non-existent simulatorExternalId', async () => {
+    const listFilterResponse = await client.simulators.listRoutines({
+      filter: { simulatorExternalIds: ['non-existent-simulator'] },
+    });
+    expect(listFilterResponse.items.length).toBe(0);
+  });
+
+  test('list routines with all filters combined', async () => {
+    const listFilterResponse = await client.simulators.listRoutines({
+      filter: {
+        simulatorExternalIds: [simulatorExternalId],
+        simulatorIntegrationExternalIds: [simulatorIntegrationExternalId],
+        modelExternalIds: [modelExternalId],
+      },
+    });
+    expect(listFilterResponse.items.length).toBeGreaterThan(0);
+    const routineFound = listFilterResponse.items.find(
+      (item) => item.externalId === routineExternalId
+    );
+    expect(routineFound?.externalId).toBe(routineExternalId);
+  });
+
   test('create routine revision', async () => {
     const response = await client.simulators.createRoutineRevisions([
       {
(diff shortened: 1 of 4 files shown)


packages/alpha/src/__tests__/api/simulatorRoutinesApi.spec.ts
```diff
@@ -112,6 +112,39 @@ describeIf('simulator routines api', () => {
...
+        simulatorIntegrationExternalIds: [simulatorIntegrationExternalId],
+        modelExternalIds: [modelExternalId],
+      },
+    });
+    expect(listFilterResponse.items.length).toBeGreaterThan(0);
+    const routineFound = listFilterResponse.items.find(
+      (item) => item.externalId === routineExternalId
+    );
+    expect(routineFound?.externalId).toBe(routineExternalId);
+  });
```
polomani, line 128: I think this single happy-path test is enough for the integration part, if we want to test all the combinations it's better to use unit tests for that (which we don't have yet).

You can use for example `/packages/stable/src/__tests__/api/sessions.unit.spec.ts` if you end up adding unit tests for this as well (although not a requirement, can do some other time)

polomani:
better to use `feat(simint-alpha): ...` for  PR title to show which part of the API is getting changed

packages/alpha/src/__tests__/api/simulatorRoutinesApi.spec.ts
mohammad-arif662: added unit test and kept one happy path integration test.

packages/alpha/src/__tests__/api/simulatorRoutinesApi.unit.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+    };
+    const response = {
+      items: [],
+    };
+
+    nock(mockBaseUrl)
+      .post(/\/api\/v1\/projects\/.*\/simulators\/routines\/list/, {
+        filter: {
+          simulatorExternalIds: ['non-existent-simulator'],
+        },
```
polomani: here and in other tests above, the filter is defined twice, can the definition at line 98 be reused?

packages/alpha/src/__tests__/api/simulatorRoutinesApi.unit.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+const mockBaseUrl = 'https://example.com';
+
+function setupMockableClient() {
+  return new CogniteClientAlpha({
+    appId: 'JS SDK unit tests (alpha)',
+    project: process.env.COGNITE_PROJECT as string,
+    baseUrl: mockBaseUrl,
+    oidcTokenProvider: () => Promise.resolve('test accessToken'),
+  });
+}
```
polomani: maybe can be imported from `import { setupMockableClient } from '../../../../core/src/__tests__/testUtils';?

mohammad-arif662: testUtils of core returns BaseCogniteClient which does not have simulator, I have added a separate setupMockableClient method in testUtils of alpha package which will return CogniteClientAlpha which has simulator

packages/alpha/src/__tests__/api/simulatorRoutinesApi.unit.spec.ts
mohammad-arif662: reused the filter variables

polomani (approved):
🦄

packages/alpha/src/__tests__/api/simulatorRoutinesApi.unit.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      })
+      .reply(200, response);
+
+    const result = await client.simulators.listRoutines({
+      filter,
+    });
+    expect(result.items).toEqual([]);
+    expect(result.items.length).toBe(0);
+  });
+});
```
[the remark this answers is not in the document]
polomani: this will throw 400

PR #1334, opened 2025-11-04
chore(cd): do not version private packages

polomani:
Currently there is an issue with codegen being version bumped as a part of release script.
Example: https://github.com/cognitedata/cognite-sdk-js/blob/81362e7ac3e394e5061d5c063a73027f452cf86d/packages/codegen/CHANGELOG.md

This fixes the issue by excluding any private package from `lerna version` command (e.g. templates, codegen, etc)


cemreyavuz (approved):
risk review ok

PR #1337, opened 2025-11-05
chore(gh-workflow): run tests on master

polomani:


polomani:
@mohammad-arif662 is working on this

PR #1338, opened 2025-11-24
feat(streams): add support for Streams API

evertoncolling:
This PR adds support for the Streams endpoints.

### DIFF
--- a/packages/stable/src/__tests__/api/streams.int.spec.ts
+++ b/packages/stable/src/__tests__/api/streams.int.spec.ts
@@ -14,10 +14,10 @@ describe('streams integration test', () => {
   beforeAll(async () => {
     client = setupLoggedInClient();
 
-    const existingStreams = await client.streams.list();
-    const existingIds = existingStreams.map((s) => s.externalId);
-
-    if (!existingIds.includes(immutableStreamId)) {
+    // Check if immutable stream exists, create if not
+    try {
+      await client.streams.retrieve({ externalId: immutableStreamId });
+    } catch {
       await client.streams.create({
         externalId: immutableStreamId,
         settings: {
@@ -28,7 +28,10 @@ describe('streams integration test', () => {
       });
     }
 
-    if (!existingIds.includes(mutableStreamId)) {
+    // Check if mutable stream exists, create if not
+    try {
+      await client.streams.retrieve({ externalId: mutableStreamId });
+    } catch {
       await client.streams.create({
         externalId: mutableStreamId,
         settings: {
--- a/packages/stable/src/__tests__/api/streams.int.spec.ts
+++ b/packages/stable/src/__tests__/api/streams.int.spec.ts
@@ -53,13 +53,11 @@ describe('streams integration test', () => {
 
     expect(immutableStream).toBeDefined();
     expect(immutableStream?.createdTime).toBeInstanceOf(Date);
-    expect(immutableStream?.lastUpdatedTime).toBeInstanceOf(Date);
     expect(immutableStream?.settings).toBeDefined();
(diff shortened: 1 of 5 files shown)


packages/stable/src/__tests__/api/streams.int.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  let client: CogniteClient;
+
+  // Persistent streams for testing
+  const immutableStreamId = 'sdk_test_immutable_stream';
+  const mutableStreamId = 'sdk_test_mutable_stream';
+
+  beforeAll(async () => {
+    client = setupLoggedInClient();
+
+    const existingStreams = await client.streams.list();
```
roligheten: Can we replace this with a retrieve on the specific stream ids instead of listing all streams?

packages/stable/src/types.ts
```diff
@@ -2359,6 +2363,131 @@ export interface SyntheticQueryResponse {
   datapoints: SyntheticDatapoint[];
 }
 
+/**
+ * Stream write - the input type for creating a stream
+ */
+export interface StreamWrite {
```
roligheten: Can we move stream specific types into the api/streams folder and do a `export * from ...` here instead? :slightly_smiling_face: Would prefer to modularize the types

packages/stable/src/types.ts
```diff
@@ -2359,6 +2363,131 @@ export interface SyntheticQueryResponse {
...
+ */
+export interface Stream extends CreatedTime {
+  /**
+   * Stream identifier
+   */
+  externalId: string;
+  /**
+   * Name of the template used for creating this stream.
+   */
+  createdFromTemplate?: string;
```
roligheten: This and "type" appear to be required on the return type from the API, can we make them non-nullable here too?

packages/stable/src/types.ts
```diff
@@ -2359,6 +2363,131 @@ export interface SyntheticQueryResponse {
...
+  /**
+   * Defines type of the stream.
+   */
+  type?: 'Immutable' | 'Mutable';
+  /**
+   * Stream settings
+   */
+  settings: StreamSettings;
```
roligheten: This is required in this type but seems to be non-required in the API spec, is that correct?

packages/stable/src/__tests__/api/streams.int.spec.ts
evertoncolling: Sure! Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1338/commits/605fc9b5696fee3c8ee39952b46274be20b07d12

packages/stable/src/types.ts
evertoncolling: Yes, it looks much better this way! I was following the current pattern, but modularizing the types should make it easier to maintain in the long run

Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1338/commits/605fc9b5696fee3c8ee39952b46274be20b07d12

packages/stable/src/types.ts
evertoncolling: Yes, good catch! Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1338/commits/605fc9b5696fee3c8ee39952b46274be20b07d12

packages/stable/src/types.ts
evertoncolling: This is a required field, but I see it's not properly labeled as so in the API spec. We'll fix it there ASAP

evertoncolling: PR to fix the API spec: https://github.com/cognitedata/service-contracts/pull/3069

roligheten (approved):
:unicorn:

PR #1341, opened 2025-12-04
feat(records): add support for ingesting/filtering immutable records

evertoncolling:
This PR add new RecordsAPI with ingest and filter methods for working with immutable records. We also update the valid options for usedFor in containers to support record containers, which is needed to test the new feature.


packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * Record to write/ingest into a stream
+ */
+export interface RecordWrite {
+  /**
+   * The space that the record belongs to
+   */
+  space: string;
```
roligheten: Replace with SpaceSpecification from instances types

roligheten: Actually never mind, we should do this once we consolidate the autogenerated types some time in the future

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
+// Copyright 2025 Cognite AS
+
+import type {
+  ContainerReference,
+  PropertyValueGroupV3,
+  RawPropertyValueV3,
+} from '../instances/types.gen';
```
roligheten: I am not sure if we should depend on these types, they are currently autogenerated by the instance api service contracts, so if these changes in DM for some reason to be incompatible with records we likely would not pick that up while updating the DM types.

At some point ™️  we need to consolidate autogenerated types (it is currently quite a mess) I'd rather that you just duplicate the types you need from these types into the records types module until we can clean up a bit in the SDK types.

packages/stable/src/api/records/types.ts
roligheten: Actually actually can you do this but copy the definition from DM into a duplicate type? Ie. so we get the regex spec the same which will make consolidating the types easier in the future

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
...
+ */
+export interface RecordWrite {
+  /**
+   * The space that the record belongs to
+   */
+  space: string;
+  /**
+   * External ID of the record
+   */
+  externalId: string;
```
roligheten: Can you do the same as mentioned above based on NodeOrEdgeExternalId (but rename it to RecordExternalId)

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * Last updated time filter for records
+ */
+export interface LastUpdatedTimeFilter {
+  /**
+   * Greater than or equal to
+   */
+  gte?: string | number;
```
roligheten: Nit: would be great to use a type alias for string called "LastUpdatedTimeRangeBound" or something that explains the spec for the string input here 🙂

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
...
+ */
+export interface RecordSort {
+  /**
+   * Property to sort by
+   */
+  property: string[];
+  /**
+   * Sort direction
+   */
+  direction?: 'ascending' | 'descending';
```
roligheten: Can we replace 'ascending' | 'descending' with a type? From experience it is nice to be able to import this as a "RecordSortDirection" in a client calling the SDK so you can assign this argument to a variable in the client code easily without redefining this union type locally.

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/**
+ * A record retrieved from a stream
+ */
+export interface RecordItem {
+  /**
+   * The space that the record belongs to
+   */
+  space: string;
```
roligheten: Ditto here ref my other comment on typing of space /externalid

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,163 @@
...
+   */
+  lastUpdatedTime?: LastUpdatedTimeFilter;
+  /**
+   * List of containers and their properties to return
+   */
+  sources?: SourceSelector[];
+  /**
+   * Filter specification
+   */
+  filter?: Record<string, unknown>;
```
roligheten, line 150: Would ofc prefer if this was typed out, even if I know it will be a chore 😅 (If you need it out asap its ok to skip this but we should type this spec at a later point anyway)

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+        },
+      };
+
+      await client.containers.upsert([containerDefinition]);
+    }
+
+    // Wait for eventual consistency only when we created new resources
+    if (needsWait) {
+      await new Promise((resolve) => setTimeout(resolve, 5000));
+    }
```
roligheten: What are we waiting for here? Container create should be non-eventually consistent

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+              value: testValue,
+              timestamp: '2025-01-01T00:00:00.000Z',
+            },
+          },
+        ],
+      },
+    ]);
+
+    // Wait for eventual consistency (record should be available at most within 2 seconds)
+    await new Promise((resolve) => setTimeout(resolve, 2000));
```
roligheten: Can we do vi.waitFor instead? Ie. wrap the filter call in waitfor with a shorter wait time. That has the benefit of potentially finishing faster if consistency takes shoter time, while also handing longer consistency times better

packages/stable/src/api/records/types.ts
evertoncolling: Since we are very close to 500 lines of code already, it would be best to not take this on this PR. I can take that on a follow up PR (there's a couple more needed to get full records support in).

packages/stable/src/api/records/types.ts
evertoncolling: Sounds good! Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/062c66c5b35afcb771c67e0f864a10666912c7b7

packages/stable/src/api/records/types.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/062c66c5b35afcb771c67e0f864a10666912c7b7

packages/stable/src/api/records/types.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/062c66c5b35afcb771c67e0f864a10666912c7b7

packages/stable/src/api/records/types.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/062c66c5b35afcb771c67e0f864a10666912c7b7

packages/stable/src/api/records/types.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/062c66c5b35afcb771c67e0f864a10666912c7b7

packages/stable/src/api/records/types.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/062c66c5b35afcb771c67e0f864a10666912c7b7

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/b39389cf4e560b697b86fff0f0fb03984feea068

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1341/commits/b39389cf4e560b697b86fff0f0fb03984feea068

packages/stable/src/api/records/types.ts
evertoncolling: I have the next PR ready with this addressed here: https://github.com/cognitedata/cognite-sdk-js/pull/1343

packages/stable/src/api/records/types.ts
```diff
@@ -0,0 +1,205 @@
...
+  type: 'container';
+}
+
+/**
+ * Sort direction for record queries
+ */
+export type RecordSortDirection = 'ascending' | 'descending';
+
+/**
+ * Timestamp bound for time range filters (ISO 8601 string or Unix ms since epoch)
```
roligheten, line 49: Nice, NB in the API docs it is not clear what the expected input format is for string, could use another look 🙂

packages/stable/src/api/records/types.ts
roligheten: :+1:

roligheten (approved):
🦄

packages/stable/src/api/records/types.ts
evertoncolling: Good catch! I can see the spec explains a lot about this field, but does not specify the supported formats. I'll open a PR to update the API spec.

PR #1344, opened 2025-12-05
feat(records): add sync endpoint and dedicated filter types

evertoncolling:
This PR builds on top of https://github.com/cognitedata/cognite-sdk-js/pull/1341 and adds:
- Sync endpoint (`client.records.sync()`) with auto-pagination support for cursor-based synchronization of records from a stream
- Dedicated filter types (`RecordFilter`, `BoolFilter`, `LeafFilter`) replacing the generic `Record<string, unknown>` type

### DIFF
--- a/packages/stable/src/api/records/recordsApi.ts
+++ b/packages/stable/src/api/records/recordsApi.ts
@@ -1,11 +1,15 @@
 // Copyright 2025 Cognite AS
 
+import type { CursorAndAsyncIterator } from '@cognite/sdk-core';
 import { BaseResourceAPI } from '@cognite/sdk-core';
 import type {
   RecordFilterRequest,
   RecordFilterResponse,
   RecordItem,
+  RecordSyncRequest,
+  RecordSyncResponse,
   RecordWrite,
+  SyncRecordItem,
 } from './types';
 
 export class RecordsAPI extends BaseResourceAPI<RecordItem> {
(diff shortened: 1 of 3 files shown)


packages/stable/src/api/records/recordsApi.ts
```diff
@@ -79,4 +83,54 @@ export class RecordsAPI extends BaseResourceAPI<RecordItem> {
...
+    const path = this.url(`${streamExternalId}/records/sync`);
+
+    const callSyncEndpoint = async (params?: RecordSyncRequest) => {
+      const response = await this.post<RecordSyncResponse>(path, {
+        data: params,
+      });
+      const items = this.addToMapAndReturn(
+        response.data.items,
+        response
+      ) as SyncRecordItem[];
```
BugGambit: `as` is scary in Typescript as we disable type-checking.

Just do:

```suggestion
      );
```

BugGambit (commented):
Nice PR. Just a minor comment

packages/stable/src/api/records/recordsApi.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1344/commits/5ca23fbb2d50b008e473401bb60eaffdb4d01c9d

BugGambit (approved):
risk review ok

PR #1345, opened 2025-12-05
feat(containers): add usedFor filter parameter to containers list

evertoncolling:
This PR adds the `usedFor` query parameter to `containers.list()` to filter containers by their intended usage (`node`, `edge`, `all`, `record`) as per the [API spec](https://api-docs.cognite.com/20230101/tag/Containers/operation/listContainers#!in=query&path=usedFor&t=request). 

The existing `callListEndpointWithGet` method serializes array values as JSON (e.g., `?usedFor=["node","record"]`), but this endpoint expects repeated query parameters (`?usedFor=node&usedFor=record`). To support this format, we added a new `callListGetEndpointWithRepeatedQueryParams` method to `BaseResourceAPI`.


packages/stable/src/api/containers/containersApi.ts
```diff
@@ -14,8 +14,15 @@ import type {
   ListOfSpaceExternalIdsResponse,
   ReducedLimitQueryParameter,
   SpaceQueryParameter,
+  UsedForQueryParameter,
 } from './types.gen';
 
+type ContainerListParams = IncludeGlobalQueryParameter &
```
roligheten: Nice! But also this type should be exported as it is part of the public API. Suggest putting this type in a new types.ts file adjacent to the types.gen.ts and exporting it from the root types.ts

evertoncolling: Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1345/commits/5af3dd6dc3cc6b6de2903c219ade3c74d4d4905b

packages/stable/src/__tests__/api/containers.int.spec.ts
```diff
@@ -104,6 +104,24 @@ describe('Containers integration test', () => {
     expect(globalContainer).toBeDefined();
   });
 
+  it('should successfully list Containers with single usedFor filter', async () => {
```
BugGambit, line 107: This test doesn't verify that the provided `usedFor` filter is used towards the API? I am assume this test will succeed even if we forget to set the query param in the implementation.

evertoncolling: That's fair! Tests now verify that every returned container matches the requested filter. Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1345/commits/1c70f2ba30ec5b9d24c6789e8891dcc33117e746

BugGambit (approved):
risk review ok

PR #1348, opened 2025-12-12
ci: enable codecov and test suite on master branch

mohammad-arif662:
- This PR is to enables code coverage reporting on PRs by adding push trigger to master branch and integrating coverage into the tests job (PRs run on changed files, master runs full coverage).
- Added `@vitest/coverage-v8@^2.0.5` dependency and resolved `minimatch` version conflict.
- Fixed flaky `instances.int.spec.ts > retrieve` test by adding `vi.waitFor()` retry logic
- Updated template package to use latest `@cognite/sdk` and `@cognite/sdk-core` versions.

Note that master codecov is not up to date, that's why we get a red mark on this PR

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -146,3 +146,4 @@ jobs:
         uses: codecov/codecov-action@v3.1.6
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -137,13 +137,10 @@ jobs:
       - name: Test
         run: |
           if [ "${{ github.event_name }}" == "pull_request" ]; then
-            yarn test --since origin/master
+            yarn test --since origin/master --coverage
           else
-            yarn test
+            yarn test --coverage
           fi
-
-      - name: Test/Codecov
-        run: yarn test:codecov
 
       - name: Codecov
         uses: codecov/codecov-action@v3.1.6
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -10,48 +10,6 @@ on:
       - master
 
 jobs:
-  codecov:
-    runs-on: ubuntu-latest
-
-    env:
-      COGNITE_PROJECT: cognitesdk-js
-      COGNITE_CREDENTIALS: ${{secrets.COGNITE_CREDENTIALS}}
-      COGNITE_BASE_URL: https://api.cognitedata.com
-      COGNITE_CLIENT_SECRET: ${{ secrets.COGNITE_CLIENT_SECRET }}
-      COGNITE_CLIENT_ID: cb491d7a-c346-4910-9a10-1650df34fbbb
-      COGNITE_AZURE_DOMAIN: cogniteappdev.onmicrosoft.com
-
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
-      - name: Use Node.js
-        uses: actions/setup-node@v4
-        with:
-          node-version: '20'
-
-      - name: Cache node_modules
-        id: cache-modules
-        uses: actions/cache@v3
-        with:
-          path: node_modules
-          key: ${{ runner.OS }}-build-${{ hashFiles('package.json') }}
-
-      - name: Install 🔧
-        run: yarn install --immutable
-
-      - name: Build
-        run: yarn build
-
-      - name: Test/Codecov
-        run: yarn test:codecov
-
-      - name: Codecov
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -13,6 +16,10 @@ jobs:
     env:
       COGNITE_PROJECT: cognitesdk-js
       COGNITE_CREDENTIALS: ${{secrets.COGNITE_CREDENTIALS}}
+      COGNITE_BASE_URL: https://api.cognitedata.com
+      COGNITE_CLIENT_SECRET: ${{ secrets.COGNITE_CLIENT_SECRET }}
+      COGNITE_CLIENT_ID: cb491d7a-c346-4910-9a10-1650df34fbbb
+      COGNITE_AZURE_DOMAIN: cogniteappdev.onmicrosoft.com
```
polomani: I am puzzled to understand why this is codecov a separate job?
is it because tests below only run with the flag `yarn test --since origin/master`?

packages/template/package.json
```diff
@@ -28,8 +28,8 @@
     "test-snippets": "yarn extract-snippets && yarn g:tsc -p codeSnippets/tsconfig.build.json"
   },
   "dependencies": {
-    "@cognite/sdk": "^9.17.1",
-    "@cognite/sdk-core": "^4.10.7"
+    "@cognite/sdk": "workspace:*",
+    "@cognite/sdk-core": "workspace:*"
```
polomani:
```suggestion
    "@cognite/sdk": "^10.4.0",
    "@cognite/sdk-core": "^5.1.3"
```
it's a template package, it should be aligned with the other packages in the folder.
If you look to a specific package (e.g. beta) it locks the version more strictly

.github/workflows/tests.yaml
polomani: if yes, let's try to reuse the tests job to push to codecov instead (remove this one). maybe it's enough for us to run only the coverage on changed files in the PR only. (needs to be tested, but still better than no coverage report)

mohammad-arif662: removed separate codecov job and added steps in tests job

packages/template/package.json
mohammad-arif662: changed version

.github/workflows/tests.yaml
```diff
@@ -183,3 +141,11 @@ jobs:
           else
             yarn test
           fi
+
+      - name: Test/Codecov
+        run: yarn test:codecov
```
polomani: why is it running the tests twice? just use `--coverage` flag on the commands above

mohammad-arif662: added this flag

.github/workflows/tests.yaml
```diff
@@ -170,4 +135,14 @@ jobs:
...
+          if [ "${{ github.event_name }}" == "pull_request" ]; then
+            yarn test --since origin/master -- --coverage
+          else
+            yarn test -- --coverage
+          fi
+
+      - name: Codecov
+        uses: codecov/codecov-action@v3.1.6
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
polomani: you are missing the new line in the end of the file

mohammad-arif662: added it

.github/workflows/tests.yaml
polomani, line 106:
```suggestion
        node-version: [20, 22]
```
node 18 is deprecated, let's do one step forward towards LTS perhaps?

polomani: never mind, we need to change terraform for that

packages/stable/src/__tests__/api/instances.int.spec.ts
```diff
@@ -276,24 +276,29 @@ describe('Instances integration test', () => {
...
+        expect(response.items[0].properties);
+        const title =
+          response.items[0].properties?.[view.space][
+            `${view.externalId}/${view.version}`
+          ].title.toString() || '';
+        expect(title.startsWith('titl'));
+      },
+      { interval: 1000, timeout: 25_000 }
+    );
+  }, 25_000);
```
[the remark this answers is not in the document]
polomani: 25 seconds is already enough, I believe

polomani (approved):

packages/stable/src/__tests__/api/instances.int.spec.ts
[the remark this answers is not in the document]
msusag: I see 25_000 is used also for other test cases. 
So if those are not experiencing flakyness, and the tests are usually completing in much less than 25 seconds, I agree this comment from Gemini seems to be not very relevant.

msusag:
🦄

PR #1351, opened 2025-12-16
feat(records): add aggregate endpoint support

evertoncolling:
This PR adds support for the Aggregate endpoint for records with full type definitions for requests and responses.


packages/stable/src/api/records/recordsApi.ts
```diff
@@ -130,4 +133,34 @@ export class RecordsAPI extends BaseResourceAPI<RecordItem> {
...
+   *     }
+   *   }
+   * });
+   * ```
+   */
+  public aggregate = async (
+    streamExternalId: string,
+    request: RecordAggregateRequest
+  ): Promise<RecordAggregateResults> => {
+    const path = this.url(`${streamExternalId}/records/aggregate`);
```
BugGambit: To be safe, let's do:

```suggestion
    const path = this.url(`${encodeURIComponent(streamExternalId)}/records/aggregate`);
```

packages/stable/src/api/records/types.ts
```diff
@@ -294,3 +294,242 @@ export interface RecordSyncResponse {
...
+}
+
+// Metric aggregates
+/** Calculates the average from the data stored by the specified property */
+export interface AvgAggregate {
+  avg: { property: AggregateProperty };
+}
+/** Counts the number of items (or non-null values when property is specified) */
+export interface CountAggregate {
+  count: { property?: AggregateProperty };
```
BugGambit, line 336: Correct, that `property` should be optional?

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -447,4 +447,179 @@ describe('records integration test', () => {
...
+              timestamp: '2025-01-01T00:00:00.000Z',
+            },
+          },
+        ],
+      },
+    ]);
+
+    await vi.waitFor(
+      async () => {
+        const prop = [testSpaceId, testContainerId, 'value'] as [
```
BugGambit:
```suggestion
        const prop = [testSpaceId, testContainerId, 'value'] satisfies [
```

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -447,4 +447,179 @@ describe('records integration test', () => {
...
+              property: [testSpaceId, testContainerId, 'name'],
+              value: testName,
+            },
+          },
+          aggregates: {
+            total_count: { count: {} },
+            total_sum: { sum: { property: prop } },
+            avg_value: { avg: { property: prop } },
+            min_value: { min: { property: prop } },
+            max_value: { max: { property: prop } },
```
BugGambit: Where did this snake-cases come from 🤔 ? Our API should use camel-case ([API design guide](https://github.com/cognitedata/service-contracts/blob/master/design_guide/API_design_guide.md#camel-case)), and this SDK uses camel-case

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -447,4 +447,179 @@ describe('records integration test', () => {
...
+          },
+        });
+        expect((aggregates.total_count as { count: number }).count).toBe(2);
+        expect((aggregates.total_sum as { sum: number }).sum).toBe(300);
+        expect((aggregates.avg_value as { avg: number }).avg).toBe(150);
+        expect((aggregates.min_value as { min: number }).min).toBe(100);
+        expect((aggregates.max_value as { max: number }).max).toBe(200);
```
BugGambit: Should we have a discriminator for these types 🤔 ?
https://dev.to/darkmavis1980/what-are-typescript-discriminated-unions-5hbb

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -447,4 +447,179 @@ describe('records integration test', () => {
...
+                aggregates: {
+                  cat_sum: {
+                    sum: { property: [testSpaceId, testContainerId, 'value'] },
+                  },
+                },
+              },
+            },
+          },
+        });
+        const result = aggregates.by_cat as BucketResult;
```
BugGambit: Same here (by_cat, cat_sum): https://github.com/cognitedata/cognite-sdk-js/pull/1351/changes#r2622633237

packages/stable/src/api/records/recordsApi.ts
evertoncolling: Good catch! I applied this to all relevant endpoints as who knows what weird stream externalIds users might end up creating.

Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1351/commits/156312727babd992c8187be87ac50098deb43fc0

packages/stable/src/api/records/types.ts
evertoncolling: Yes, If users want to just count the over number of records, they don't need to specify any property. If they want to count the number of non-null entries for a given property, then they have to specify a property.

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1351/commits/9c51251609a42ab98465bba5f2a14addeb2b7bd8

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: That's my Python biased brain mixing it up. Fixed in https://github.com/cognitedata/cognite-sdk-js/pull/1351/commits/a493cd11ccbe9d65f1e07b03a687e025a2c99ed7

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: The API does not return a discriminator field, so the result type depends on what aggregate was requested, which the SDK currently does not track at the type level. We could potentially add runtime type guards as a future enhancement (as that would probably get us beyond 500 LOC in this PR).

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: Fixed in: https://github.com/cognitedata/cognite-sdk-js/pull/1351/commits/0ce395c0c533d3ca4510821ad4200443bb9f7090

BugGambit (approved):
risk review ok

PR #1352, opened 2025-12-17
feat(api): add cognite functions

sceniclife:
Add cognite functions APIs to sdk

Tests added.
Tested functionality with the quickstart.ts


packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,142 @@
...
+   * ```
+   */
+  public list = async (
+    scope?: FunctionListScope
+  ): Promise<CogniteFunction[]> => {
+    const response = await this.post<ItemsWrapper<CogniteFunction[]>>(
+      this.listPostUrl,
+      { data: scope || {} }
+    );
+    return this.addToMapAndReturn(response.data.items, response);
```
polomani, line 52: why not the built-in one? is it because the endpoint doesn't support the cursor?
```suggestion
    return this.listEndpoint<FunctionListScope>(
      this.callListEndpointWithPost,
      filter
    );
```

packages/stable/src/api/functions/functionsApi.ts
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * [Retrieve functions by IDs](https://developer.cognite.com/api#tag/Functions/operation/byIdsFunctions)
+   *
+   * ```js
+   * const functions = await client.functions.retrieveMultiple([{id: 123}, {externalId: 'abc'}]);
+   * ```
+   */
+  public retrieveMultiple = (ids: IdEither[]): Promise<CogniteFunction[]> => {
+    return super.retrieveEndpoint(ids);
+  };
```
polomani: maybe enough having the regular array retrieve only? array of one support one element anyway

polomani: I meant keep this retrieveMultiple only (and rename to retrieve)

packages/stable/src/api/functions/functionsApi.ts
sceniclife: Correct! It does not support cursor.

I dug into infra repo to see if adding cursor would be feasible. It is.
But do we need cursor when projects are limited to 100-250 cognite functions?

packages/stable/src/api/functions/functionsApi.ts
sceniclife: You're right. I can remove retrieveMultiple and just use retrieve calling `super.retrieveEndpoint`

sceniclife: Done

christjt (approved):
risk review ok

PR #1353, opened 2025-12-18
feat(records): add upsert and delete endpoints for mutable records

evertoncolling:
This PR adds support for the upsert and delete endpoints for mutable records. This is the final PR in a series, enabling full feature parity of the Records and Streams API in the Javascript SDK.


roligheten (approved):
🦄

PR #1355, opened 2025-12-19
ci: run with node 22, remove 18

polomani:
Node 18 is deprecated and node 20 is going to in 2026, we have to switch to LTS

Changes:

- set node 22 by default for most jobs
- run tests with node 20 + 22


roligheten (approved):
🦄

PR #1358, opened 2026-01-06
feat(instances): implement support for explicitly setting search operator

roligheten:
Currently it is only possible to use OR as search operator (API default) with this the user can set it manually.

Also added a warning that the default operator will change in the next major version.

### DIFF
--- a/packages/stable/src/__tests__/api/instances.unit.spec.ts
+++ b/packages/stable/src/__tests__/api/instances.unit.spec.ts
@@ -1,0 +1,60 @@
+// Copyright 2020 Cognite AS
+
+import type { ViewReference } from '@cognite/sdk-stable';
+import nock from 'nock';
+import { beforeAll, describe, expect, test, vi } from 'vitest';
+import { mockBaseUrl, setupMockableClient } from '../testUtils';
+
+describe('instances unit tests', () => {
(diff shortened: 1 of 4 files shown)


packages/stable/src/api/instances/instancesApi.ts
```diff
@@ -21,6 +21,8 @@ import type {
 } from './types.gen';
 
 export class InstancesAPI extends BaseResourceAPI<NodeOrEdge> {
+  private searchOperatorChangeWarningDisplayed = false;
```
BugGambit, line 24: Should we have a TODO to clean this up?

packages/stable/src/__tests__/api/instances.unit.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    space: 'cdf_core',
+    type: 'view',
+    version: 'v1',
+  };
+
+  beforeAll(async () => {
+    nock.cleanAll();
+  });
+
+  test('search operator warning should be logged the first time making a search request with default operator', async () => {
```
BugGambit, line 20: Strictly speaking, we don't assert that it was logged the first time. Just that it was logged once after two requests

PR #1360, opened 2026-01-06
refactor(timeseries): move time series types to dedicated module

evertoncolling:
- Moved all time series related types from the global `types.ts` to a dedicated `api/timeSeries/types.ts` module
- Extracted shared primitive types (`Timestamp`, `Metadata`, `CreatedAndLastUpdatedTime`, `DatapointInfo`, patch types, etc.) into `types/common.ts` to avoid circular dependencies
- Updated imports in `timeSeriesApi.ts` and `syntheticTimeSeriesApi.ts` to use the new modules
- Re-exported types from both new modules in `types.ts` to maintain backward compatibility


packages/stable/src/types.ts
```diff
@@ -2967,7 +2711,7 @@ export type SequencesSource = {
 
 export type TimeSeriesSource = {
   type: 'timeSeries';
-  filter?: ObjectOrString<TimeseriesFilter>;
+  filter?: ObjectOrString<import('./api/timeSeries/types').TimeseriesFilter>;
```
roligheten: I actually didn't know you could do an inline import like this, is in done for a specific reason? I would prefer to have all the imports toplevel

packages/stable/src/api/timeSeries/types.ts
```diff
@@ -0,0 +1,395 @@
+// Copyright 2025 Cognite AS
+
+import type {
```
roligheten, line 3: There are a lot of types moved here, hard to review everything but in  general we need to ensure:
1. All the types are unchanged after the move, including type names (Ie. all that should have been done is to copy-paste definitions from types.ts into this file, if the types changed we need to do a more through review or do this change in the next major version)
2. All the types that used to be exported from the root types.ts file, need to still be exported from this file indirectly
I think 2 looks good but can you confirm 1 @evertoncolling ?

packages/stable/src/api/timeSeries/types.ts
```diff
@@ -0,0 +1,395 @@
...
+  ExternalId,
+  FilterQuery,
+  IdEither,
+  InstanceId,
+  InternalId,
+  Limit,
+} from '@cognite/sdk-core';
+
+/** Timestamp - either a number (ms since epoch) or a Date object */
+type Timestamp = number | Date;
```
roligheten: The types not exported, are they copied from some other place? (I cant see them removed from types.ts)

packages/stable/src/types.ts
evertoncolling: It's been a while since I worked on this. If I recall I had to resort to this pattern due to a circular reference. But I'll give it another shot this week and see if I can find another way to solve it.

packages/stable/src/api/timeSeries/types.ts
evertoncolling: Yes, that was exactly the change done here -> "copy-paste definitions from types.ts into this file". I also confirmed that the linter and compiler didn't complain and existing tests continue to work as they should.

packages/stable/src/api/timeSeries/types.ts
evertoncolling: Good catch! These types are duplicated here as local types. They still exist and are exported from types.ts.

The reason for the duplication is again the circular import issue -> since types.ts re-exports from this new file `(export * from './api/timeSeries/types')`, this file cannot import from types.ts at the top level. The local types are intentionally not exported to avoid conflicting with the original exports in types.ts.

This is not ideal of course, but I wanted to avoid a larger PR and introducing any breaking changes. Some possible alternatives could be:
1. Extract these shared primitive types to a separate module (like, common/types.ts or move them to @cognite/sdk-core) that both files can import from
2. Keep the duplication as-is since these are simple type aliases that are unlikely to diverge

What do you think? If needed we can grow this beyond 500 lines by going for option 1 or some other option...

evertoncolling: Or alternatively, I can create yet another PR starting with something like option 1, and then I revisit this one later...

packages/stable/src/types.ts
evertoncolling: If we decide to proceed with the option we discuss below, to move the shared primitive types to a separate module, I should be able to get rid of this ugly thing here as well 😅

roligheten:
I would maybe suggest we push this off to the next major release, while I think it will work I don't really like that we need to to these dynamic imports to avoid conflicts, I'd rather move the common types around so we dont need them, but I think that should be done in a major release.

roligheten (dismissed):

packages/stable/src/types.ts
BugGambit: Agree, we shouldn't have inline imports like this

evertoncolling: Addressed here: https://github.com/cognitedata/cognite-sdk-js/pull/1360/commits/190b9ce5954af2f9d1f0ff12472638c4c8c3ee7f

packages/stable/src/api/timeSeries/types.ts
evertoncolling: Addressed here: https://github.com/cognitedata/cognite-sdk-js/pull/1360/commits/190b9ce5954af2f9d1f0ff12472638c4c8c3ee7f

packages/stable/src/api/timeSeries/types.ts
```diff
@@ -0,0 +1,395 @@
...
+type NullableSinglePatchString = { set: string } | { setNull: true };
+
+/** Nullable patch for number fields */
+type NullableSinglePatchLong = { set: number } | { setNull: true };
+
+/** Metadata patch operations */
+type MetadataPatch = { set: Metadata } | { add?: Metadata; remove?: string[] };
+
+/** Array patch for number arrays */
+type ArrayPatchLong = { set: number[] } | { add?: number[]; remove?: number[] };
```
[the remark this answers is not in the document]
evertoncolling: Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1360/commits/190b9ce5954af2f9d1f0ff12472638c4c8c3ee7f

BugGambit (approved):
risk review ok

PR #1361, opened 2026-01-07
feat(records): add hasNext support for manual pagination in sync endpoint

evertoncolling:
This PR adds support for manual pagination in the Records `sync` endpoint by exposing `hasNext` and `nextCursor` in the response, enabling users to implement their own pagination logic while preserving the existing auto-pagination functionality.

The current implementation was handling autopagination correctly, but was setting the `nextCursor` to `undefined` on the last page. The sync endpoint always returns a `nextCursor`, allowing user to save it and use it later for polling results since the last request.

To fix this, we extend the autopagination logic to properly support endpoints that also include a `hasNext` parameter, like records `sync` and instances `sync`, without introducing breaking changes.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -6,6 +6,13 @@ export interface ListResponse<T> extends CursorResponse<T> {
 
 export interface CursorResponse<T> extends ItemsWrapper<T> {
   nextCursor?: string;
+  /**
+   * Indicates whether more data is immediately available.
+   * When present, this takes precedence over `nextCursor` for determining
+   * if auto-pagination should continue. When `hasNext` is `false`, pagination
+   * stops even if `nextCursor` is defined.
+   */
+  hasNext?: boolean;
 }
 
 export type ItemsResponse<T> = ItemsWrapper<T[]>;
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -6,13 +6,6 @@ export interface ListResponse<T> extends CursorResponse<T> {
 
 export interface CursorResponse<T> extends ItemsWrapper<T> {
   nextCursor?: string;
-  /**
-   * Indicates whether more data is immediately available.
-   * When present, this takes precedence over `nextCursor` for determining
-   * if auto-pagination should continue. When `hasNext` is `false`, pagination
-   * stops even if `nextCursor` is defined.
-   */
-  hasNext?: boolean;
 }
 
 export type ItemsResponse<T> = ItemsWrapper<T[]>;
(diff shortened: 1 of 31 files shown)


packages/core/src/types.ts
```diff
@@ -6,6 +6,13 @@ export interface ListResponse<T> extends CursorResponse<T> {
 
 export interface CursorResponse<T> extends ItemsWrapper<T> {
   nextCursor?: string;
+  /**
+   * Indicates whether more data is immediately available.
+   * When present, this takes precedence over `nextCursor` for determining
+   * if auto-pagination should continue. When `hasNext` is `false`, pagination
+   * stops even if `nextCursor` is defined.
+   */
+  hasNext?: boolean;
```
roligheten: Maybe I am being pedantic but it does bother me a bit that we are adding this parameter to the common CursorResponse that will afaik only ever be returned by the records API. This means that all api users will see that in their cursor response there is this additional parameter that they will think they need to handle, but which is actually only relevant in the records api.

To address this could we instead:
1. Implement a separate cursor type returned only by the records endpoint
2. Implement a separate "cursorBasedEndpointWithHasNext" method that implements the logic specific to this endpoint.
3. Revert the changes to cursorBasedEndpoint and the CursorResponse type.
4. Change the return type of sync to be the alternative cursor response type.

For the record (no pun intended) I think having a shared cursoring implementation is probably something that does not make sense anymore, it made sense in the classic api where every api had the same cursoring logic, but now in the data modelling / records world it seems to me that this does not hold any more.

evertoncolling: Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1361/commits/f939f10978d0bd9ad62691e1e82a029ebbb4c153

packages/stable/src/api/records/recordsApi.ts
```diff
@@ -155,50 +156,57 @@ export class RecordsAPI extends BaseResourceAPI<RecordItem> {
...
+      callSyncEndpoint(request).then((syncData) =>
+        addSyncNextPageFunction(callSyncEndpoint, syncData, request)
+      );
+
+    const autoPaginationMethods = makeAutoPaginationMethods(firstPagePromise);
+
+    return Object.assign(
+      firstPagePromise,
+      autoPaginationMethods
+    ) as RecordsSyncCursorAndAsyncIterator<SyncRecordItem>;
```
BugGambit: Would be great to remove this `as` type-cast. What makes typescript complain 🤔 ?

BugGambit (commented):
Looks good. But have a comment

packages/stable/src/api/records/recordsApi.ts
evertoncolling: Good catch! It was not needed at all. Removed in: https://github.com/cognitedata/cognite-sdk-js/pull/1361/commits/bbc262979de9c52bfe454c3ddfeecfde035d4bc1

BugGambit (approved):
risk review ok

PR #1362, opened 2026-01-07
fix(docs): restore documentation deployment and add missing API exports

evertoncolling:
Fixes the SDK documentation not being deployed to GitHub Pages since February 2025.

### DIFF
--- a/packages/stable/src/index.ts
+++ b/packages/stable/src/index.ts
@@ -45,4 +45,19 @@ export { SyntheticTimeSeriesAPI } from './api/timeSeries/syntheticTimeSeriesApi'
 export { TimeSeriesAPI } from './api/timeSeries/timeSeriesApi';
 export * from './api/templates';
 
+export { AnnotationsAPI } from './api/annotations/annotationsApi';
+export { ContainersAPI } from './api/containers/containersApi';
+export { DocumentsAPI } from './api/documents/documentsApi';
+export { GeospatialAPI } from './api/geospatial/geospatialAPI';
+export { InstancesAPI } from './api/instances/instancesApi';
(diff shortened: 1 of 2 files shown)


packages/stable/src/index.ts
```diff
@@ -45,4 +45,19 @@ export { SyntheticTimeSeriesAPI } from './api/timeSeries/syntheticTimeSeriesApi'
...
+export { InstancesAPI } from './api/instances/instancesApi';
+export { DataModelsAPI } from './api/models/datamodelsApi';
+export { RecordsAPI } from './api/records/recordsApi';
+export { SessionsApi } from './api/sessions/sessionsApi';
+export { SpacesAPI } from './api/spaces/spacesApi';
+export { StreamsAPI } from './api/streams/streamsApi';
+export { UnitsAPI } from './api/units/unitsApi';
+export { ProfilesAPI } from './api/userProfiles/profilesApi';
+export { ViewsAPI } from './api/views/viewsApi';
+export { VisionAPI } from './api/vision/visionApi';
```
roligheten, line 62: Is this necessary for the docs publishing?

evertoncolling: Good question! I previewed the docs locally and I was not getting streams and records to be included in the generated docs. I remember I looked into the script and found out I had to update this file, and that did the trick. Maybe someone who was involved with the generation of the SDK docs might know if that is the right approach? Would you know by chance @polomani?

roligheten: I guess these should be here anyway, seems like everything missing here are newer apis that we forgot to export in this file, I dont see any harm in adding them

roligheten (approved):
🦄

PR #1378, opened 2026-02-25
feat(FUN-462): add lastCalled field to CogniteFunction type

trygveu:
## Summary

- Adds `lastCalled?: Date` to the `CogniteFunction` interface in `types.ts`
- Adds `'lastCalled'` to `pickDateProps` in `functionsApi.ts` so the SDK converts the raw ms integer from the API into a `Date` object, consistent with `createdTime`
- The field is absent when the function has never been called

### DIFF
--- a/packages/stable/src/__tests__/api/functions.unit.spec.ts
+++ b/packages/stable/src/__tests__/api/functions.unit.spec.ts
@@ -82,6 +82,7 @@ describe('Functions unit test', () => {
     const func = await client.functions.retrieve([{ id: 123 }]);
     expect(func[0].id).toBe(mockFunction.id);
     expect(func[0].name).toBe(mockFunction.name);
+    expect(func[0].lastCalled).toBeInstanceOf(Date);
   });
 
   test('delete', async () => {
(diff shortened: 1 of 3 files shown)


jonasblanc (approved):
I'm not familiar enough with the codebase to say if the new field was added in all required places, but the added code is reasonable.

packages/stable/src/__tests__/api/functions.unit.spec.ts
```diff
@@ -64,6 +66,7 @@ describe('Functions unit test', () => {
     });
     expect(functions.length).toBe(1);
     expect(functions[0].id).toBe(mockFunction.id);
+    expect(functions[0].lastCalled).toBeInstanceOf(Date);
```
notrixbe, line 69: @trygveu 
What date do you use? string? Or date object?

trygveu: The api returns Unix ms, we want the SDK to convert to Date object

trygveu:
risk review ok

PR #1382, opened 2026-02-26
refactor(datapoints): move data points types to dedicated module

evertoncolling:
- Moved all data points related types from the global `types.ts` to a dedicated `api/dataPoints/types.ts` module
- Moved `IgnoreUnknownIds` to `types/common.ts` as a shared primitive to avoid circular dependencies
- Updated imports in `dataPointsApi.ts` to use the new modules
- Re-exported all types from `types.ts` to maintain backward compatibility


BugGambit (approved):
risk review ok

PR #1386, opened 2026-03-02
feat(timeseries,datapoints): align types with latest API spec

evertoncolling:
This PR aligns the stable SDK types with the latest API specification for time series and data points, adding support for status codes, new aggregate types, and additional query capabilities. Redundant type overrides in beta are removed since they are now covered by stable.

**Time Series types** (`api/timeSeries/types.ts`):
- Added `TimeseriesType` (`'numeric' | 'string' | 'state'`) and optional `type` field to `Timeseries` response type
- Added `unitQuantity` filter to `TimeseriesFilter`
- Added `timeZone` to `SyntheticQuery`

**Data Points types** (`api/dataPoints/types.ts`):
- Added `DatapointStatus` interface and `status` field to `DoubleDatapoint`, `StringDatapoint`, and `ExternalDatapoint`
- Added new aggregate types: `maxDatapoint`, `minDatapoint`, `countGood`, `countUncertain`, `countBad`, `durationGood`, `durationUncertain`, `durationBad`
- Added `DatapointExtremum` interface for min/max aggregate responses
- Added `includeStatus`, `ignoreBadDataPoints`, `treatUncertainAsBad` to `DatapointsQueryProperties` and `LatestDataPropertyFilter`
- Added `targetUnit` and `targetUnitSystem` to `LatestDataPropertyFilter`
- Fixed date parsing for nested `timestamp` fields inside `maxDatapoint`/`minDatapoint` aggregates

**Beta cleanup**:
- Removed redundant datapoints type overrides from `@cognite/sdk-beta` (`DatapointStatus`, `Aggregate`, `DatapointsQueryProperties`, `DatapointAggregate`, `Datapoints`, `DoubleDatapoint`, `StringDatapoint`, etc.) now that they are available in stable
- Removed explicit re-exports from `beta/src/index.ts` as `export * from '@cognite/sdk'` now covers them

All changes are additive to stable. No breaking changes for consumers of either package.

PR #1387, opened 2026-03-02
feat(datapoints): simplify monthly aggregates and promote beta to stable

evertoncolling:
Builds on top of #1386. This PR simplifies the `retrieveDatapointMonthlyAggregates` implementation by leveraging the API's native monthly granularity support, and completes the promotion of beta datapoints functionality to stable.

**Simplified `retrieveDatapointMonthlyAggregates`** (`api/dataPoints/dataPointsApi.ts`):
- Replaced the client-side workaround (N separate API calls per month with day-based granularity + client-side merge) with a single API call using the now-supported `1mo` granularity
- Removed `getMonthsBetweenDates` helper and `MonthInfo` interface
- Fixed the previously-skipped monthly granularity integration tests by using UTC timestamps throughout, making them timezone-agnostic

**Beta cleanup**:
- Moved quality-indicator integration tests from beta to stable
- Removed the `DataPointsAPI` override class from beta (now redundant)
- Removed the custom `datapoints` getter from beta's `CogniteClient`
- Beta consumers now inherit all datapoints functionality directly from stable

No breaking changes for consumers of either package.


packages/stable/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -64,70 +59,10 @@ export class DataPointsAPI extends BaseResourceAPI<
...
-      }
-
-      return resultSet;
-    }
-
-    return this.retrieveDatapointsEndpoint<DatapointAggregates[]>(query);
+    return this.retrieveDatapointsEndpoint<DatapointAggregates[]>({
+      ...query,
+      granularity: '1mo',
+    });
```
BugGambit, line 65: And this will yield the same result as before? Before we used the exact start/end date of the month

packages/stable/src/__tests__/api/datapoints.int.spec.ts
```diff
@@ -156,6 +162,120 @@ describe('Datapoints integration test', () => {
...
+      {
+        id: timeserie.id,
+        includeStatus: true,
+        ignoreBadDataPoints: false,
+        treatUncertainAsBad: false,
+      },
+    ]);
+    expect(response[0].datapoints.length).toBeGreaterThan(0);
+    const dp = response[0].datapoints[0] as DoubleDatapoint;
+    if (dp.status) {
```
BugGambit: What happens if this statement is falsy? Should the test fail?

packages/stable/src/__tests__/api/datapoints.int.spec.ts
```diff
@@ -354,267 +415,366 @@ describe.skip('Datapoints integration test for monthly granularity', () => {
...
+        {
+          id: timeserie.id,
+          start: new Date('2022-10-01T00:00:00Z'),
+          end: new Date('2023-04-01T00:00:00Z'),
+          includeStatus: true,
+          ignoreBadDataPoints: false,
+          treatUncertainAsBad: false,
+        },
+      ],
+    })) as Datapoints[];
```
BugGambit, line 766: Do you need `as Datapoints[]` here?

packages/stable/src/__tests__/api/datapoints.int.spec.ts
evertoncolling: Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1387/commits/fc6100ad7c03299418fb2ec8d03a92e65708a44a

packages/stable/src/api/dataPoints/dataPointsApi.ts
evertoncolling: Yes, the API now natively supports 1mo as a granularity unit, which aligns aggregates to calendar month boundaries. The old approach was a workaround that approximated months using day-based granularity (28d/30d/31d). The server-side 1mo is actually more correct since it handles calendar months, time zones, and leap years properly. The integration tests for monthly aggregates pass with this change.

packages/stable/src/__tests__/api/datapoints.int.spec.ts
evertoncolling: Yeah, the cast is needed because `retrieve()` returns `DatapointAggregates[] | Datapoints[]`. Without it, TS doesn't know we're getting raw datapoints here, so accessing `.status` on the individual data points would be a type error (since `DatapointAggregate` doesn't have a `status` field).

BugGambit (dismissed):
risk review ok

BugGambit (approved):
risk review ok

PR #1396, opened 2026-03-18
fix(core): restrict retryable HTTP status codes to match Python SDK

silverchen-cognite:
## Summary

- Restricts retryable HTTP status codes in `isValidRetryStatusCode()` from `{1xx, 429, 5xx}` to `{429, 502, 503, 504}`, matching the [Python SDK's default `status_forcelist`](https://github.com/cognitedata/cognite-sdk-python/blob/65b6af3e34be86337966af8273f6e44e9786a687/cognite/client/config.py#L52).
- Removes the `lodash/inRange` dependency from the retry validator.
- Updates unit tests to cover the new status code boundaries (502 for retryable scenarios, 500 as explicitly non-retryable).

## Motivation

The TypeScript SDK was retrying on status codes like 500 (Internal Server Error) and 501 (Not Implemented), which the Python SDK intentionally excludes. Retrying 500s can cause duplicate side effects for non-idempotent operations. This change brings the two SDKs into alignment.

### DIFF
--- a/packages/core/src/httpClient/retryValidator.ts
+++ b/packages/core/src/httpClient/retryValidator.ts
@@ -1,6 +1,5 @@
 // Copyright 2020 Cognite AS
 
-import inRange from 'lodash/inRange';
 import some from 'lodash/some';
 
 import type { HttpMethod, HttpResponse } from './basicHttpClient';
@@ -65,12 +64,10 @@ export const createUniversalRetryValidator =
     return isValidRetryStatusCode(response.status);
   };
 
+const RETRYABLE_STATUS_CODES = new Set([429, 502, 503, 504]);
+
 function isValidRetryStatusCode(status: number) {
-  return (
-    inRange(status, 100, 200) ||
-    inRange(status, 429, 430) ||
-    inRange(status, 500, 600)
-  );
+  return RETRYABLE_STATUS_CODES.has(status);
 }
 
 function matchPathWithEndpoints(path: string, endpoints: string[]) {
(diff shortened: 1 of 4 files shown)


packages/core/src/httpClient/retryValidator.ts
```diff
@@ -65,12 +64,10 @@ export const createUniversalRetryValidator =
     return isValidRetryStatusCode(response.status);
   };
 
+const RETRYABLE_STATUS_CODES = new Set([429, 502, 503, 504]);
```
andeplane, line 67: Are we not retrying 408?

andeplane, line 67: And not 500 quite clearly?

packages/core/src/httpClient/retryValidator.unit.spec.ts
```diff
@@ -42,6 +42,10 @@ describe('cdfRetryValidator', () => {
     ...baseResponse,
     status: 500,
   };
+  const response502: HttpResponse<unknown> = {
+    ...baseResponse,
+    status: 502,
+  };
```
andeplane, line 48: Should we add a new test that we don't retry e.g. 500?

andeplane (commented):
Ok I suppose, but I had some questions and one proposal. I do trust the PR (or really the choices in Python SDK)

packages/core/src/httpClient/retryValidator.ts
silverchen-cognite: I don't know if we should, but we don't do it in Python SDK

packages/core/src/httpClient/retryValidator.unit.spec.ts
silverchen-cognite: It's already there, at line 71

PR #1397, opened 2026-03-18
fix(core): stop retrying DELETE requests on 5xx errors to align with …

silverchen-cognite:
## Summary

DELETE is not safely idempotent in all cases, so it should not be automatically retried on server errors. This aligns with the Python SDK's behavior where DELETE is routed through the non-retryable HTTP
client. Retry on 429 (Too Many Requests) is preserved since that applies unconditionally regardless of HTTP method.

- Remove `DELETE` from the retryable HTTP methods list in `createUniversalRetryValidator`, aligning retry behavior with the Python SDK where DELETE requests are not automatically retried on 5xx errors.
- 429 (Too Many Requests) responses are still retried for DELETE, since that logic applies unconditionally to all methods.

## Background

The Python SDK routes DELETE requests through its non-retryable HTTP client, meaning they are only retried on 429 (via rate-limit handling) and never on 5xx. The TypeScript SDK was treating DELETE as fully idempotent and retrying it on all server errors, which could lead to unintended duplicate deletions in failure scenarios.

### DIFF
--- a/packages/core/src/httpClient/cdfHttpClient.unit.spec.ts
+++ b/packages/core/src/httpClient/cdfHttpClient.unit.spec.ts
@@ -196,17 +196,17 @@ describe('CDFHttpClient', () => {
       const headerValue = '123';
       client.addOneTimeHeader(headerKey, headerValue);
       nock(baseUrl, { reqheaders: { [headerKey]: headerValue } })
-        .delete('/')
+        .put('/')
         .reply(101, {});
       nock(baseUrl, { reqheaders: { [headerKey]: headerValue } })
-        .delete('/')
+        .put('/')
         .reply(200, [1]);
       nock(baseUrl, { badheaders: [headerKey] })
-        .delete('/')
+        .put('/')
         .reply(200, {});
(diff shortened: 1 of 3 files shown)


packages/core/src/httpClient/cdfHttpClient.unit.spec.ts
```diff
@@ -196,17 +196,17 @@ describe('CDFHttpClient', () => {
       const headerValue = '123';
       client.addOneTimeHeader(headerKey, headerValue);
       nock(baseUrl, { reqheaders: { [headerKey]: headerValue } })
-        .delete('/')
+        .put('/')
```
andeplane, line 199: Why do we update this test here?

andeplane (commented):
See one question

packages/core/src/httpClient/cdfHttpClient.unit.spec.ts
silverchen-cognite: This tests "should send it after retry", and DELETE is no longer retry-able after this change

PR #1401, opened 2026-03-19
fix(core): align HTTP method retry behavior with Python SDK

silverchen-cognite:
## Summary

- **Remove DELETE** from the default retryable methods list in `createUniversalRetryValidator` — the Python SDK does not retry it on 5xx.
- **Add PATCH** to the default retryable methods list — the Python SDK treats PATCH as retryable alongside GET and PUT, but the JS SDK was missing it.
- Add unit tests for both PATCH retry behavior and DELETE non-retry behavior.

## Motivation

The Python SDK's `_is_retryable` considers `GET`, `PUT`, and `PATCH` retryable, while the JS SDK had `GET`, `HEAD`, `OPTIONS`, `DELETE`, `PUT`. This PR brings them into alignment.

### DIFF
--- a/packages/core/src/httpClient/retryValidator.ts
+++ b/packages/core/src/httpClient/retryValidator.ts
@@ -56,7 +56,7 @@ export const createUniversalRetryValidator =
       return true;
     }
     // By default, retry requests with HTTP verbs that are meant to be idempotent
-    const httpMethodsToRetry = ['GET', 'HEAD', 'OPTIONS', 'PUT'];
+    const httpMethodsToRetry = ['GET', 'HEAD', 'OPTIONS', 'PUT', 'PATCH'];
     const isRetryableHttpMethod =
       httpMethodsToRetry.indexOf(request.method.toUpperCase()) !== -1;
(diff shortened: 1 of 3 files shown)


packages/core/src/httpClient/retryValidator.ts
```diff
@@ -56,7 +56,7 @@ export const createUniversalRetryValidator =
       return true;
     }
     // By default, retry requests with HTTP verbs that are meant to be idempotent
-    const httpMethodsToRetry = ['GET', 'HEAD', 'OPTIONS', 'DELETE', 'PUT'];
+    const httpMethodsToRetry = ['GET', 'HEAD', 'OPTIONS', 'PUT', 'PATCH'];
```
EliasBjorne, line 59: Will this be breaking? Could we keep the Delete retry just to be safe?

EliasBjorne (commented):
Looks good, my only worry would be that removing the `DELETE` can have some unwanted consequences, but other than that I think it makes sence to align the retry policies for the cognite sdks

PR #1415, opened 2026-04-22
feat(alpha): add support for metering-service

Aflases:
MUT-1430: This PR adds SDK support for the Metering service API in the alpha package with four endpoints:

`retrieveMeters` — Filter meters with advanced filtering (POST /metering/meters/list). Only the prefix operator is supported (e.g. on meterId). This could be use for fetching all the meterIds when filter attribute is not provided.
`retrieveConsumptionData` — Fetch multiple meters by id (POST /metering/meters/byids). Takes meterIds[] and optional range param and builds the items body which uses retrieveEndpoint for chunked requests and returns { items: MeterReading[] }, also used for fetching ConsumptionData for single meterId
filterMeters use listEndpoint for pagination (CursorAndAsyncIterator) and metadata handling consistent with other alpha resources.

This change also refactors the alpha Limits client: limitsApi.ts follows the same BaseResourceAPI pattern as other resources (custom listGetUrl / listPostUrl for /limits/values and /limits/values/list), and the public types (LimitsValue, LimitAdvanceFilter, MeteringLimitsPrefixFilter) are aligned with the alpha Limits API docs.

Changes:

- CogniteClientAlpha sets cdf-version: 20230101-alpha as a default header on the HTTP client in initAPIs (applies to alpha APIs using that client, not only Metering).
- getDateProps on MeteringAPI — parse timestamp on MeteringDatapoint under items / datapoints so responses match typed Date fields.
- TypeScript types: MeteringDatapoint, MeterReading, MeterConsumptionRangeParams, MeterConsumptionListParams, MeterConsumptionFilterQuery and shared MeteringLimitsPrefixFilter with Limits where the filter shape matches.
- Integrate MeteringAPI and LimitsAPI into CogniteClientAlpha as client.metering and client.limits.
- Unit tests: meteringApi.unit.spec.ts (Metering), limitsApi.unit.spec.ts (Limits).
- Limits API: model as BaseResourceAPI<LimitsValue> with listGetUrl → values, listPostUrl → values/list and retrieveByLimitId (GET …/values/{limitId}), getAllLimits (GET …/values + listEndpoint), retrieveByAdvancedFilter (POST …/values/list + listEndpoint).
- Introduce / align LimitsValue, LimitAdvanceFilter, and shared MeteringLimitsPrefixFilter for the advanced list filter shape (prefix on property paths).

### DIFF
--- a/packages/alpha/src/__tests__/api/meteringApi.unit.spec.ts
+++ b/packages/alpha/src/__tests__/api/meteringApi.unit.spec.ts
@@ -17,102 +17,7 @@ describe('Metering API unit tests', () => {
     client = setupMockableClient();
   });
 
-  describe('retrieveConsumptionData', () => {
-    test('should retrieve consumption data for a single meter', async () => {
-      const meterId = 'atlas.monthly_ai_tokens';
-      const mockedResponse = {
-        meterId: 'atlas.monthly_ai_tokens',
-        datapoints: [],
-      };
-
-      nock(mockBaseUrl)
-        .get(
-          /\/api\/v1\/projects\/.*\/metering\/meters\/atlas\.monthly_ai_tokens$/
-        )
-        .matchHeader('cdf-version', '20230101-alpha')
-        .reply(200, mockedResponse);
-
-      const result = await client.metering.retrieveConsumptionData(meterId);
-      expect(result.meterId).toEqual(mockedResponse.meterId);
-      expect(result.datapoints).toEqual([]);
-    });
-
-    test('should convert datapoint timestamps from epoch ms to Date', async () => {
-      const meterId = 'files.storage_bytes';
-      const mockedResponse = {
-        meterId: 'files.storage_bytes',
-        datapoints: [{ timestamp: epochMs, value: 42.5 }],
-      };
-
-      nock(mockBaseUrl)
-        .get(/\/api\/v1\/projects\/.*\/metering\/meters\/files\.storage_bytes$/)
-        .matchHeader('cdf-version', '20230101-alpha')
-        .reply(200, mockedResponse);
-
-      const result = await client.metering.retrieveConsumptionData(meterId);
-      expect(result.datapoints).toHaveLength(1);
-      expect(result.datapoints[0].value).toBe(42.5);
-      expect(result.datapoints[0].timestamp).toEqual(new Date(epochMs));
-    });
-
-    test('should pass optional range query params', async () => {
-      const meterId = 'atlas.monthly_ai_tokens';
-      const query = {
-        start: 1000,
-        end: 2000,
-        numberOfDatapoints: 3,
-      };
-      const mockedResponse = {
-        meterId,
-        datapoints: [{ timestamp: epochMs, value: 1 }],
-      };
-
-      nock(mockBaseUrl)
-        .get(
-          /\/api\/v1\/projects\/.*\/metering\/meters\/atlas\.monthly_ai_tokens$/
-        )
-        .query(query)
-        .matchHeader('cdf-version', '20230101-alpha')
-        .reply(200, mockedResponse);
-
-      const result = await client.metering.retrieveConsumptionData(
-        meterId,
-        query
-      );
-      expect(result.datapoints[0].timestamp).toEqual(new Date(epochMs));
-    });
-
-    test('should URL-encode special characters in meterId', async () => {
-      const meterId = 'service/meter_name';
-      const mockedResponse = {
-        meterId: 'service/meter_name',
-        datapoints: [],
-      };
-
-      nock(mockBaseUrl)
-        .get(/\/api\/v1\/projects\/.*\/metering\/meters\/service%2Fmeter_name$/)
-        .matchHeader('cdf-version', '20230101-alpha')
-        .reply(200, mockedResponse);
-
-      const result = await client.metering.retrieveConsumptionData(meterId);
-      expect(result.meterId).toEqual(mockedResponse.meterId);
-    });
-
-    test('should reject when meter is not found', async () => {
-      const meterId = 'non.existent_meter';
-
-      nock(mockBaseUrl)
-        .get(/\/api\/v1\/projects\/.*\/metering\/meters\/non\.existent_meter$/)
-        .matchHeader('cdf-version', '20230101-alpha')
-        .reply(404, { error: { message: 'Meter not found', code: 404 } });
-
-      await expect(
-        client.metering.retrieveConsumptionData(meterId)
-      ).rejects.toThrow();
-    });
-  });
-
-  describe('filterMeters', () => {
+  describe('retrieveMeters', () => {
     test('should filter meters with prefix on meterId', async () => {
       const filter: MeterConsumptionFilterQuery = {
         filter: {
@@ -148,7 +53,7 @@ describe('Metering API unit tests', () => {
         .matchHeader('cdf-version', '20230101-alpha')
         .reply(200, mockedResponse);
 
-      const result = await client.metering.filterMeters(filter);
+      const result = await client.metering.retrieveMeters(filter);
       expect(result.items).toEqual(mockedResponse.items);
       expect(result.nextCursor).toBe('cursor-123');
       expect(result.items.length).toBe(2);
@@ -170,12 +75,12 @@ describe('Metering API unit tests', () => {
         .matchHeader('cdf-version', '20230101-alpha')
         .reply(200, mockedResponse);
 
-      const result = await client.metering.filterMeters(filter);
+      const result = await client.metering.retrieveMeters(filter);
       expect(result.items).toEqual(mockedResponse.items);
       expect(result.items.length).toBe(1);
     });
 
-    test('should handle empty filter results', async () => {
+    test('should handle empty filter results in retrieve all Meters', async () => {
       const filter: MeterConsumptionFilterQuery = {
         filter: {
           prefix: {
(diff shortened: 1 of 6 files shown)


rbtcollins:
This is not ready for risk review; it does not have team approvals.

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+    meterId: string,
+    params?: MeterConsumptionRangeParams
+  ): Promise<MeterReading> => {
+    const path = this.url(`meters/${encodeURIComponent(meterId)}`);
+    const response = await this.get<MeterReading>(path, {
+      params,
+      headers: this.header,
+    });
+    return this.addToMapAndReturn(response.data, response);
+  };
```
polomani: have a look like other APIs are done, e.g. 
```
// runsApi.ts
 public retrieve = (ids: CogniteInternalId[]) => {
    const items = ids.map((id) => ({ id }));
    return this.retrieveEndpoint(items);
  };
```

So the absolute minimum to change here, is to rename the method into `retrieveConsumptionData`, so the method tells what you are actually retrieving (since the API is called metering, not consumptionData)

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  /**
+   * [List all meters](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/listConsumptionData)
+   *
+   * @param queryparams - Cursor, limit, and optional historical range (epoch ms).
+   */
+  public listMeters = (
+    queryparams?: MeterConsumptionListParams
```
polomani:
```suggestion
    query?: MeterConsumptionListParams
```
to align with other APIs

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+      return {
+        ...response,
+        data: {
+          items: response.data.items || [],
+          nextCursor: response.data?.nextCursor,
+        },
+      };
+    };
+
+    return this.listEndpoint(endpointCaller, queryparams);
```
polomani: why do you need this endpointCaller? 

why not just using the standard list helper?
```
return this.listEndpoint(this.callListEndpointWithGet, {
      ...query,
    });
```

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+  MeterConsumptionByIdsRequest,
+  MeterConsumptionByIdsResponse,
+  MeterConsumptionListParams,
+  MeterConsumptionRangeParams,
+  MeterReading,
+} from '../../types';
+
+export class MeteringAPI extends BaseResourceAPI<MeterReading> {
+  private header = {
+    'cdf-version': '20230101-alpha',
```
polomani: you can do this in `this.httpClient.setDefaultHeader` cogniteClient.ts, I don't think other APIs would care much (currently it says alpha)

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+      return {
+        ...response,
+        data: {
+          items: response.data.items || [],
+          nextCursor: response.data?.nextCursor,
+        },
+      };
+    };
+
+    return this.listEndpoint(endpointCaller, filter);
```
polomani: if you remove the header implementation from here, you can use the standard `list` implementation

packages/alpha/src/api/metering/meteringApi.ts
polomani: is this consumptionData or reading?
you need to land on a single you call it

packages/alpha/src/types.ts
```diff
@@ -631,3 +631,41 @@ export interface FilterObject {
     value: string;
   };
 }
+
+export interface Datapoint {
+  timestamp: Date;
+  value: number;
+}
```
polomani: you need to prefix Datapoint with something, MeteringDataPoint perhaps? the other name is primarily used by time series, it'll create ambiguity

packages/alpha/src/types.ts
```diff
@@ -631,3 +631,41 @@ export interface FilterObject {
...
+  filter?: FilterObject;
+}
+
+/** POST /metering/meters/byids request body. */
+export interface MeterConsumptionByIdsRequest
+  extends MeterConsumptionRangeParams {
+  items: { meterId: string }[];
+}
+
+export interface MeterConsumptionByIdsResponse {
```
polomani: you don't need this type. use ItemsResponse

polomani:
generally, I suggest looking to how older/classic APIs are done and try to replicate it, there are a lot of inconsistencies and misalignment here, unfortunately

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  /**
+   * [Retrieve multiple consumption data by their ids](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/fetchConsumptionDataByIds)
+   *
+   * @param request - Meter ids and optional historical range (epoch ms).
+   */
+  public retrieveByMeterIds = async (
+    request: MeterConsumptionByIdsRequest
```
polomani: user should probably just pass list of ids `string[]`, don't you think?

packages/alpha/src/types.ts
```diff
@@ -631,3 +631,41 @@ export interface FilterObject {
...
+
+/** GET /metering/meters — cursor, limit, and optional historical range. */
+export type MeterConsumptionListParams = FilterQuery &
+  MeterConsumptionRangeParams;
+
+/** POST /metering/meters/list — prefix filter on `meterId` only. */
+export interface MeterConsumptionAdvancedFilter
+  extends FilterQuery,
+    MeterConsumptionRangeParams {
+  filter?: FilterObject;
```
polomani: FilterObject (not part of the PR but has to be fixed) shouldn't just be called without a prefix, it'll clash with other APIs. Use `MeterConsumptionFilter` or whatever, you need to come up with a good prefix you can use across

packages/alpha/src/types.ts
```diff
@@ -631,3 +631,41 @@ export interface FilterObject {
...
+}
+
+/** POST /metering/meters/byids request body. */
+export interface MeterConsumptionByIdsRequest
+  extends MeterConsumptionRangeParams {
+  items: { meterId: string }[];
```
polomani: you can use `ItemsWrapper` here

packages/alpha/src/types.ts
```diff
@@ -631,3 +631,41 @@ export interface FilterObject {
...
+  end?: number;
+  numberOfDatapoints?: number;
+}
+
+/** GET /metering/meters — cursor, limit, and optional historical range. */
+export type MeterConsumptionListParams = FilterQuery &
+  MeterConsumptionRangeParams;
+
+/** POST /metering/meters/list — prefix filter on `meterId` only. */
+export interface MeterConsumptionAdvancedFilter
```
polomani: why is it called advanced? would there be less advanced?
usually this would have been called `MeterConsumptionFilterQuery`

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,123 @@
...
+  ): Promise<MeterConsumptionByIdsResponse> => {
+    const path = this.url('meters/byids');
+    const response = await this.post<MeterConsumptionByIdsResponse>(path, {
+      data: request,
+      headers: this.header,
+    });
+    return this.addToMapAndReturn(response.data, response);
+  };
```
polomani, line 63: perhaps `this.retrieveEndpoint` should be used instead?

packages/alpha/src/api/metering/meteringApi.ts
Aflases: yes make sense.

packages/alpha/src/types.ts
Aflases: There won't be ,but we named the method according to the public Doc , but true , `MeterConsumptionFilterQuery` makes more sense.

packages/alpha/src/types.ts
Aflases: removed , and decoupled the structure for ease of caller implementation.

packages/alpha/src/api/metering/meteringApi.ts
Aflases: Understood, it is indeed consumptionData

packages/alpha/src/types.ts
```diff
@@ -621,25 +621,25 @@ export interface LimitsValue {
...
-export interface FilterObject {
+export interface MeteringLimitsPrefixFilter {
   prefix?: {
     property: string[];
     value: string;
   };
 }
 
-export interface Datapoint {
+export interface LimitAdvanceFilter extends FilterQuery {
```
polomani: is it supposed to be `AdvanceD`? in that case perhaps better to either fix the typo or drop the "Advanced" part completely, as you did with the other one

packages/alpha/src/api/limits/limitsApi.ts
```diff
@@ -37,52 +35,18 @@ export class LimitsAPI extends BaseResourceAPI<LimitsValue> {
...
+   * @param query - The query parameters to apply to the limits.
    * @returns The list of limits.
    */
   public getAllLimits = (
-    queryparams?: FilterQuery
+    query?: FilterQuery
   ): CursorAndAsyncIterator<LimitsValue> => {
-    const path = this.url('values');
-
-    const endpointCaller = async (query?: FilterQuery) => {
```
polomani, line 71: nice, thank you!

polomani:
Talked with Aman, this will be simplified even more:
- single list endpoint instead of separate list and filter
- single byIds endpoint

packages/alpha/src/__tests__/api/meteringApi.unit.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+
+describe('Metering API unit tests', () => {
+  let client: CogniteClientAlpha;
+
+  beforeEach(() => {
+    nock.cleanAll();
+    client = setupMockableClient();
+  });
+
+  describe('retrieveMeters', () => {
```
polomani:
```suggestion
  describe('listMeters', () => {
```

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * [Filter meters](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/listConsumptionDataAdvanced)
+   *
+   * Only the `prefix` filter operator is supported (e.g. on `meterId`).
+   * This method would be used to list all meters or filter meters by prefix.
+   * @param query - Filter, cursor, limit, and optional historical range.
+   */
+  public retrieveMeters = (
```
polomani:
```suggestion
  public listMeters = (
```

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,66 @@
...
+      ['timestamp']
+    );
+  }
+
+  /**
+   * [Filter meters](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/listConsumptionDataAdvanced)
+   *
+   * Only the `prefix` filter operator is supported (e.g. on `meterId`).
+   * This method would be used to list all meters or filter meters by prefix.
+   * @param query - Filter, cursor, limit, and optional historical range.
```
polomani:
```suggestion
   * [Filter meters](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/listConsumptionDataAdvanced)
   *
   * List meters matching he filter
   * @param query - Filter, cursor, limit, and optional historical range.
```

phrase `would be used` looks weird

packages/alpha/src/types.ts
```diff
@@ -627,13 +621,37 @@ export interface LimitsValue {
   value: number;
 }
 
-export interface LimitAdvanceFilter extends FilterQuery {
-  filter?: FilterObject;
-}
-
-export interface FilterObject {
+export interface MeteringLimitsPrefixFilter {
```
```suggestion
export interface MeteringLimitsFilter {
```
this way you can add more filters later

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,64 @@
...
+    return this.pickDateProps<MeteringDatapoint>(
+      ['items', 'datapoints'],
+      ['timestamp']
+    );
+  }
+
+  /**
+   * [Filter meters](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/listConsumptionDataAdvanced)
+   *
+   * List meters matching he filter
```
polomani:
```suggestion
   * List meters matching the filter
```

packages/alpha/src/api/metering/meteringApi.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  protected get byIdsUrl() {
+    return this.url('meters/byids');
+  }
+
+  /**
+   * @hidden
+   */
+  protected getDateProps() {
```
polomani, line 24: are you planning to add more endpoints that do not call metering/meters? I mean if all the methods call `metering/meters` you can just define API as `this.apiFactory(MeteringAPI, 'metering/meters');` and then no need to override the URLs here

polomani (approved):
🦄

PR #1419, opened 2026-05-03
chore(stable): upgrade code generated types

andersfylling:
Bump snapshot for the stable package and upgrade types.

Documents API deprecated search aggregate over a year ago. The recommended way is to use the dedicated aggregate endpoint instead.

In the last 90 days, no customer project has used this feature

> [!NOTE]
> Only user is cognitesdk-js
image


packages/codegen/src/generate.ts
```diff
@@ -113,7 +114,10 @@ async function generateForAllServices(options: GenerateOptions) {
     '\n'
   )}`;
 
-  await fs.writeFile(exportFilePath, fileContent);
+  await fs.writeFile(
+    exportFilePath,
+    formatWithBiome(fileContent, exportFilePath)
```
EliasBjorne, line 119: what is this?

EliasBjorne (commented):
some tests are failing but all in all look good

roligheten (approved):

roligheten:
NB: This is technically a breaking change as someone may still theoretically call, but since its been deprecated for a very long time and afaik Anders has metrics that it has not been called for half a year I think its fine that we remove it

EliasBjorne:
> NB: This is technically a breaking change as someone may still theoretically call, but since its been deprecated for a very long time and afaik Anders has metrics that it has not been called for half a year I think its fine that we remove it

Could there be that some one rely on the types without using the api?

roligheten:
> > NB: This is technically a breaking change as someone may still theoretically call, but since its been deprecated for a very long time and afaik Anders has metrics that it has not been called for half a year I think its fine that we remove it
> 
> Could there be that some one rely on the types without using the api?

Yes but they seem quite specific so not sure what someone would use them for outside this context

andersfylling:
:unicorn:

silverchen-cognite:
What's the reason of not doing the major version bump?

roligheten:
> What's the reason of not doing the major version bump?

There is no specific reason, traditionally we have aimed to do fewer major version bumps where we bundle several breaking changes together to avoid the need for customers to manually check for breaking changes too often.

We could change that to do more frequent smaller major version bumps but we then risk more customers being stuck on older major versions and not getting automated minor/patch updates for the newest major.

PR #1420, opened 2026-05-04
feat(records): add support for unit conversion

evertoncolling:
This PR brings the Records SDK up to parity with the API spec by adding unit conversion (`targetUnits`) and typing information (`includeTyping`) to the `/records/filter`, `/records/sync`, and `/records/aggregate` endpoints.

## Non-breaking API design

`filter()` and `aggregate()` keep their original return types unless `includeTyping: true` is passed:

- `client.records.filter(...)` returns `RecordItem[]` (unchanged)
- `client.records.filter(..., { includeTyping: true, ... })` returns `{ items: RecordItem[]; typing?: RecordTypeInformation }`
- `client.records.aggregate(...)` returns `RecordAggregateResults` (unchanged)
- `client.records.aggregate(..., { includeTyping: true, ... })` returns `{ aggregates: RecordAggregateResults; typing?: RecordTypeInformation }`

TypeScript overloads discriminate the return shape when `includeTyping` is the literal `true`. A dynamic `boolean` variable will not narrow the return type.

`sync()` already returned a page object; this PR only adds optional `typing` on that shape.

## Test plan

- [x] Integration tests for `targetUnits` and `includeTyping` on filter, aggregate, and sync
- [x] Existing filter/aggregate tests unchanged for default return shapes


BugGambit:
Ref your suggestions, I like both. However, in the 2nd option where we add `includeTyping: true`, we must ensure the existing type does not break. I.e. typescript knows it will return the type `{ aggregates: RecordAggregateResults; typing?: RecordTypeInformation }` if and only if `includeTyping: true` was passed.
I think going for option 2 is the best.

evertoncolling:
> Ref your suggestions, I like both. However, in the 2nd option where we add `includeTyping: true`, we must ensure the existing type does not break. I.e. typescript knows it will return the type `{ aggregates: RecordAggregateResults; typing?: RecordTypeInformation }` if and only if `includeTyping: true` was passed. I think going for option 2 is the best.

Thanks for the feedback! Option 2 is now implemented and the PR is ready for review.

packages/stable/src/api/records/recordsApi.ts
```diff
@@ -140,17 +145,33 @@ export class RecordsAPI extends BaseResourceAPI<RecordItem> {
...
+    return items;
+  }) as {
+    (
+      streamExternalId: string,
+      request: RecordFilterRequestWithTyping
+    ): Promise<RecordFilterResponse>;
+    (
+      streamExternalId: string,
+      request?: RecordFilterRequestWithoutTyping
+    ): Promise<RecordItem[]>;
```
BugGambit: You rather want to use overloads here: https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

`as`/type-casting disabled type-checking

packages/stable/src/api/records/recordsApi.ts
```diff
@@ -225,29 +247,50 @@ export class RecordsAPI extends BaseResourceAPI<RecordItem> {
...
       data: request,
     });
+    if (request.includeTyping === true) {
+      return {
+        aggregates: response.data.aggregates,
+        typing: response.data.typing,
+      };
+    }
     return response.data.aggregates;
+  }) as {
```
BugGambit: Same here

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -78,6 +79,50 @@ describe('records integration test', () => {
 
       await client.containers.upsert([containerDefinition]);
     }
+
+    const unitsContainerLookup = await client.containers.retrieve([
+      { space: testSpaceId, externalId: unitsContainerId },
+    ]);
+    if (
+      !unitsContainerLookup.items.some((c) => c.externalId === unitsContainerId)
```
BugGambit: Just curious, since we do call upsert methods (`client.spaces.upsert` and `client.containers.upsert`) then  the operations are already idempotent. What's the value of this check?

packages/stable/src/__tests__/api/records.int.spec.ts
```diff
@@ -41,12 +42,12 @@ describe('records integration test', () => {
...
-      await client.containers.retrieve([
-        { space: testSpaceId, externalId: testContainerId },
-      ]);
-    } catch {
-      // Container doesn't exist, need to create space and container
+    const baseContainerLookup = await client.containers.retrieve([
+      { space: testSpaceId, externalId: testContainerId },
+    ]);
+    if (
+      !baseContainerLookup.items.some((c) => c.externalId === testContainerId)
```
BugGambit: Same here: https://github.com/cognitedata/cognite-sdk-js/pull/1420/changes#r3339722946

packages/stable/src/api/records/recordsApi.ts
evertoncolling: Yep, good call! Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1420/commits/c16b6a19401e9fdff15d290b49b0c205bcd00d8c

packages/stable/src/api/records/recordsApi.ts
evertoncolling: Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1420/commits/c16b6a19401e9fdff15d290b49b0c205bcd00d8c

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: Fair point! We can trust the API! Addressed in: https://github.com/cognitedata/cognite-sdk-js/pull/1420/commits/29bd597c3eff1d162001700604f7965bcf14aeb8

packages/stable/src/__tests__/api/records.int.spec.ts
evertoncolling: Addressed in https://github.com/cognitedata/cognite-sdk-js/pull/1420/commits/29bd597c3eff1d162001700604f7965bcf14aeb8

BugGambit (approved):
risk review ok

PR #1427, opened 2026-05-08
feat(beta): add datapoints.insertWithUnitConversion for client-side unit conversion on write

evertoncolling:
Adds a beta-only method for inserting time series datapoints when values are expressed in a different unit than the time series stored `unitExternalId`. Values are converted in the SDK before calling the existing datapoints insert endpoint, so stable `insert()` behavior and types stay unchanged.

## Motivation
https://hub.cognite.com/ideas/cognite-infield-measurement-units-1148?tid=1148&fid=None

CDF supports unit conversion on datapoint read (`targetUnit` / `targetUnitSystem`), but not on write. This is not a problem for the vast majority of use cases when data comes straight from an extractor which taps into a data source with a consistent source unit. This is not the case though for manual data entry, like what is offered in Infield. With the method introduced in this PR, Infield can allow users to select which unit they want to use when entering data.

### DIFF
--- a/packages/beta/src/api/dataPoints/dataPointsApi.ts
+++ b/packages/beta/src/api/dataPoints/dataPointsApi.ts
@@ -19,6 +19,17 @@ import type {
 import { CogniteError } from '@cognite/sdk-core';
 import type { DatapointsInsertWithUnitItem } from './types';
 
+const CONVERSION_SIG_DIGITS = 12;
+
+function roundToSigDigits(value: number, digits: number): number {
+  if (value === 0) return 0;
+  const factor = Math.pow(
+    10,
+    digits - Math.floor(Math.log10(Math.abs(value))) - 1
+  );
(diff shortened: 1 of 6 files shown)


packages/beta/src/api/dataPoints/dataPointsApi.ts
```diff
@@ -0,0 +1,238 @@
...
+function dedupeRetrieveIds(
+  ids: IdEitherWithInstance[]
+): IdEitherWithInstance[] {
+  const seen = new Map<string, IdEitherWithInstance>();
+  for (const id of ids) {
+    const key = identifierKey(toIdentifier(id));
+    if (!seen.has(key)) seen.set(key, id);
+  }
+  return [...seen.values()];
+}
```
evertoncolling, line 92: Maybe a bit overkill, but we need some dedup logic here 😅

PR #1438, opened 2026-06-10
feat(alpha): add support for workflows execution

Roshiniswetha:
This PR adds Workflow executions support to the JavaScript SDK.

**Included**

Added WorkflowExecutionsAPI with support for:

list
retrieve
run
retry
cancel

**Exposed workflow executions through:**

client.workflowExecutions

Added and exported public workflow execution types, including:

WorkflowExecution
WorkflowExecutionDetails
WorkflowTaskExecution
WorkflowExecutionFilterQuery
WorkflowExecutionRunRequest
WorkflowExecutionRetryRequest
WorkflowExecutionCancelRequest
Testing
Added unit tests for all supported operations (list, retrieve, run, retry, cancel with and without reason).
Related: CDF-27832

### DIFF
--- a/packages/alpha/src/__tests__/api/workflowExecutions.int.spec.ts
+++ b/packages/alpha/src/__tests__/api/workflowExecutions.int.spec.ts
@@ -1,0 +1,163 @@
+// Copyright 2026 Cognite AS
+
+import { afterAll, beforeAll, describe, expect, test } from 'vitest';
+import type CogniteClientAlpha from '../../cogniteClient';
+import type {
+  TaskDefinition,
+  Version,
+  Workflow,
+  WorkflowExecution,
+} from '../../types';
+import { randomInt, setupLoggedInClient } from '../testUtils';
+
+const workflowTasks: TaskDefinition = {
+  externalId: 'int-execution-task-1',
+  type: 'function',
+  parameters: {
+    function: { externalId: 'int-fn-external-id' },
+  },
+};
+
+describe('Workflow executions integration test', () => {
+  let client: CogniteClientAlpha;
+  const workflowExternalId = `int-workflow-execution-${randomInt()}`;
+  const versionExternalId = '1';
+  let createdWorkflow: Workflow | undefined;
+  let createdVersion: Version | undefined;
+  let createdExecution: WorkflowExecution | undefined;
+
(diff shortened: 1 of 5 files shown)


packages/alpha/src/api/workflows/workflowExecutionsApi.ts
```diff
@@ -0,0 +1,113 @@
...
+   *   authentication: { nonce: 'session-nonce' },
+   * });
+   * ```
+   */
+  public run = (
+    workflowExternalId: CogniteExternalId,
+    version: string,
+    request: WorkflowExecutionRunRequest
+  ): Promise<WorkflowExecution> => {
+    const path = `${this.workflowsBasePath()}/${encodeURIComponent(workflowExternalId)}/versions/${encodeURIComponent(version)}/run`;
```
bkuzma: Disclaimer: this is AI generated but it's a valid concern about this base path logic:

The `run` method uses `workflowsBasePath()` to derive the workflows root URL by regex-stripping `/executions` from `this.url('')`. The risk is **silent failure**: `String.replace()` is a no-op when the regex doesn't match, so if the internal URL structure ever changes (e.g. a trailing query string, a refactor of `BaseResourceAPI.url()`), the method silently POSTs to the wrong path — no error is thrown, just a 404 or unexpected behavior at runtime.

Suggested fix: add a guard that throws if the regex doesn't match, or better, accept the workflows base URL as a constructor argument so the path is injected rather than reverse-engineered.

bkuzma (approved):

packages/alpha/src/__tests__/api/workflowExecutions.int.spec.ts
```diff
@@ -0,0 +1,163 @@
+// Copyright 2026 Cognite AS
```
notrixbe, line 1: @Roshiniswetha 
Don't we need to add it somewhere in index.js to make it accessible in code ?

can you check pr of @Ratna-Phaneendra-Matukumalli

Roshiniswetha: It is added in workflows and versions api unit and integration tests because test files are not part of index.js

notrixbe (approved):
🦄

PR #1445, opened 2026-06-19
chore: add queryTyped API endpoint

christjt:
## Description
Adds a queryTyped method to InstancesAPI that infers return types from a const query declaration, providing full TypeScript type safety on query results.

- New method instances.queryTyped<TQueryRequest>(params) — wraps the existing /instances/query endpoint with generic type inference. When the query is declared as const query = { ... } as const satisfies QueryRequest, the return type is fully inferred: result expression keys, source spaces, view external IDs, and individual property names are all typed.
- New query.types.ts — contains the QueryResult<TQueryRequest, TSelectSourceWithParams> and SelectSourceWithParams utility types that power the inference. Supports a second generic TSelectSourceWithParams for passing explicit property value types when richer typing is needed.
- Type-level feature highlights:
  - Result keys match select keys from the query
  - Source spaces and externalId/version view identifiers are inferred per result expression
  - Property names are inferred per source; * (wildcard) falls back to Record<string, unknown>
  - Falls back gracefully to Record<string, unknown> for non-const / runtime queries
- Exports QueryResult and SelectSourceWithParams from packages/stable/src/types.ts
- Type tests added in query.spec.ts covering all inference scenarios using vitest's expectTypeOf

## Example usage:
 ```ts
const query = {
  with: {
    result_set_1: {
      nodes: {},
    },
  },
  select: {
    result_set_1: {
      sources: [{ source: { type: 'view', space: 'mySpace', externalId: 'MyView', version: 'v1' }, properties: ['name', 'age'] }],
    },
  },
} as const satisfies QueryRequest;

const typedResponse = await client.instances.queryTyped(query);
// For fully typed results, declare the query variable as:
// const query = { ... } as const satisfies QueryRequest;
// then: typedResponse.items.result_set_1[0].properties.mySpace['MyView/v1'].name
```

## Visual Aids

https://github.com/user-attachments/assets/671b519e-3fe8-4c76-baf8-2bcbbecc1953

### DIFF
--- a/packages/stable/src/__tests__/api/instances.int.spec.ts
+++ b/packages/stable/src/__tests__/api/instances.int.spec.ts
@@ -1,7 +1,15 @@
 // Copyright 2024 Cognite AS
 
-import type { ViewReference } from 'stable/src/types';
-import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
+import type { QueryRequest, ViewReference } from 'stable/src/types';
+import {
+  afterAll,
+  beforeAll,
+  describe,
+  expect,
+  expectTypeOf,
+  test,
+  vi,
+} from 'vitest';
 import type CogniteClient from '../../cogniteClient';
 import { randomInt, setupLoggedInClient } from '../testUtils';
 
@@ -58,12 +66,12 @@ describe('Instances integration test', () => {
     name: TEST_SPACE_EXT_ID,
     description: 'Instance space used for integration tests.',
   };
-  const view: ViewReference = {
+  const view = {
     externalId: 'Describable',
     space: 'cdf_core',
     type: 'view',
     version: 'v1',
-  };
+  } as const satisfies ViewReference;
 
   const describable1: Describable = {
     externalId: `describable_1_${timestamp}`,
@@ -318,7 +326,50 @@ describe('Instances integration test', () => {
         result_set_1: {},
       },
     });
+    expect(response.items.resuslt_set_1).toHaveLength(1);
+  }, 10_000);
+
+  test('queryTyped', async () => {
+    const query = {
+      with: {
+        result_set_1: {
+          nodes: {
+            filter: {
+              equals: {
+                property: ['node', 'externalId'],
+                value: describable1.externalId,
+              },
+            },
+          },
+        },
+      },
+      select: {
+        result_set_1: {
+          sources: [
+            { source: view, properties: ['title', 'description', 'labels'] },
+          ],
+        },
+      },
+    } as const satisfies QueryRequest;
+
+    const response = await client.instances.queryTyped<
+      typeof query,
+      [
+        {
+          source: typeof view;
+          properties: { title: string; description: string; labels: string[] };
+        },
+      ]
+    >(query);
+
     expect(response.items.result_set_1).toHaveLength(1);
+
+    const resultProperties =
+      response.items.result_set_1[0].properties.cdf_core['Describable/v1'];
+
+    expectTypeOf(resultProperties.title).toEqualTypeOf<string>();
+    expectTypeOf(resultProperties.description).toEqualTypeOf<string>();
+    expectTypeOf(resultProperties.labels).toEqualTypeOf<string[]>();
   }, 10_000);
 
   test('sync', async () => {
--- a/packages/stable/src/__tests__/api/instances.int.spec.ts
+++ b/packages/stable/src/__tests__/api/instances.int.spec.ts
@@ -326,7 +326,7 @@ describe('Instances integration test', () => {
         result_set_1: {},
       },
     });
-    expect(response.items.resuslt_set_1).toHaveLength(1);
+    expect(response.items.result_set_1).toHaveLength(1);
   }, 10_000);
 
   test('queryTyped', async () => {
--- a/packages/stable/src/__tests__/types/instances/query.spec.ts
+++ b/packages/stable/src/__tests__/types/instances/query.spec.ts
@@ -1,0 +1,182 @@
+/*!
+ * Copyright 2024 Cognite AS
+ */
+
+import type { QueryRequest } from 'stable/src/types';
+import { describe, expectTypeOf, test } from 'vitest';
+import type { InstancesAPI } from '../../../api/instances/instancesApi';
+
+describe('queryNodesEdges type tests', () => {
+  test('query result keys should match const query select keys', () => {
+    type QueryResultSelectKeys = keyof Awaited<
+      ReturnType<typeof InstancesAPI.prototype.queryTyped<typeof testQuery>>
+    >['items'];
+
+    expectTypeOf<QueryResultSelectKeys>().toEqualTypeOf<
+      keyof typeof testQuery.select
+    >();
+  });
+
+  test('Each source with unique space should map to a key in properties of result', () => {
+    type ResultSourceSpaces = keyof Awaited<
+      ReturnType<typeof InstancesAPI.prototype.queryTyped<typeof testQuery>>
+    >['items']['resultExpressionA'][number]['properties'];
+
+    type QueryResultSpaces =
+      (typeof testQuery.select.resultExpressionA.sources)[number]['source']['space'];
+    expectTypeOf<ResultSourceSpaces>().toEqualTypeOf<QueryResultSpaces>();
+  });
+
+  test('property keys of result should match property keys of sources', () => {
+    type ResultPropertiesA = keyof Awaited<
+      ReturnType<typeof InstancesAPI.prototype.queryTyped<typeof testQuery>>
+    >['items']['resultExpressionA'][number]['properties']['spaceA']['externalIdA/v1'];
+    type QueryPropertiesA =
+      (typeof testQuery.select.resultExpressionA.sources)[0]['properties'][number];
+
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: packages/stable/src/api/instances/query.types.ts [hunk 1]
```diff
@@ -0,0 +1,80 @@
...
+  Pick<SelectSource, SOURCE>
+>[PROPERTIES];
+
+export type QueryResult<
+  TQueryRequest extends QueryRequest,
+  TSelectSourceWithParams extends
+    SelectSourceWithParams = SelectSourceWithParams,
+> = {
+  items: {
+    [SELECT_KEY in keyof TQueryRequest[SELECT]]: Array<
```
EliasBjorne: nit: 
does this need to be screaming case? Could we do this pascal case instead? Screaming case is mostly used for CONST

packages/stable/src/api/instances/query.types.ts
```diff
@@ -0,0 +1,80 @@
...
+type EDGES = 'edges';
+type PROPERTIES = 'properties';
+type SOURCES = 'sources';
+type SOURCE = 'source';
+type SPACE = 'space';
+type EXTERNALID = 'externalId';
+type VERSION = 'version';
+type ALLPROPERTIES = '*';
+
+type InstanceType<
```
EliasBjorne: InstanceType can have a conflict with the typescript utillity InstanceType?

packages/stable/src/api/instances/query.types.ts
```diff
@@ -0,0 +1,80 @@
...
+    SelectSourceWithParams = SelectSourceWithParams,
+> = {
+  items: {
+    [SELECT_KEY in keyof TQueryRequest[SELECT]]: Array<
+      InstanceType<TQueryRequest, SELECT_KEY> & {
+        properties: {
+          [SELECT_SOURCE in NonNullable<
+            NonNullable<TQueryRequest[SELECT][SELECT_KEY]>[SOURCES]
+          >[number] as SELECT_SOURCE[SOURCE][SPACE]]: {
+            [SELECT_SOURCE_VAR in SELECT_SOURCE as `${SELECT_SOURCE_VAR[SOURCE][EXTERNALID]}/${SELECT_SOURCE_VAR[SOURCE][VERSION]}`]: SELECT_SOURCE_VAR[PROPERTIES][0] extends ALLPROPERTIES
```
EliasBjorne: should we add some docstring to this, so we know how this works, that we are reverse engineering the typre relationships between the DMS query requests and response?

packages/stable/src/__tests__/types/instances/query.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+  });
+
+  test('Passing a non-constant query should be valid', () => {
+    type QueryResult = Awaited<
+      ReturnType<typeof InstancesAPI.prototype.queryTyped<QueryRequest>>
+    >;
+
+    type TestType = QueryResult extends {
+      items: Record<string, unknown>;
+      nextCursors?: Record<string, unknown>;
```
EliasBjorne:
```suggestion
      nextCursor?: Record<string, unknown>;
```

EliasBjorne (commented):
Brilliant, added some minor coments, also the code Cov indicates that we could add some more tests, but other than that this is fantastic 🚀

packages/stable/src/__tests__/types/instances/query.spec.ts
christjt: Looking at this test again, I don't actually like it. Instead of checking against this hard coded type I should be comparing `QueryResult` with the actual `QueryResponse` type from the SDK. Refactored this.

packages/stable/src/api/instances/query.types.ts
christjt: Woops, nice catch. Renamed to `DmsInstanceType`

packages/stable/src/api/instances/instancesApi.ts
```diff
@@ -244,6 +245,39 @@ export class InstancesAPI extends BaseResourceAPI<NodeOrEdge> {
...
+   *    select: {
+   *      result_set_1: {
+   *        sources: [{ source: { type: 'view', space: 'mySpace', externalId: 'MyView', version: 'v1' }, properties: ['name', 'age'] }],
+   *      },
+   *    },
+   *  });
+   *  // For fully typed results, declare the query variable as:
+   *  // const query = { ... } as const satisfies QueryRequest;
+   *  // then: typedResponse.items.result_set_1[0].properties.mySpace['MyView/v1'].name
+   * ```
```
EliasBjorne, line 269: Nit: show the preffered way of building the query in the doc-string instead of just a coment
```suggestion
   * ```js
   * 
   *  const query = {
   *    with: {
   *      result_set_1: {
   *        nodes: {},
   *      },
   *    },
   *    select: {
   *      result_set_1: {
   *        sources: [{ source: { type: 'view', space: 'mySpace', externalId: 'MyView', version: 'v1' }, properties: ['name', 'age'] }],
   *      },
   *    },
   *  } as const satisfies QueryRequest
   *  const typedResponse = await client.instances.queryTyped(query);
   *  // For fully typed results, declare the query variable as:
   *  // const query = { ... } as const satisfies QueryRequest;
   *  // then: typedResponse.items.result_set_1[0].properties.mySpace['MyView/v1'].name
   * ```
```

EliasBjorne (commented):
Looks good, could we add a minimal runtime test to
```
packages/stable/src/__tests__/api/instances.int.spec.ts
```
I think this also would make the CodeCov improve. I know, this is implicitly tested through the query endpoint, but if for some reason, things change and the query endpoint is removed in a stupid way, all the testing og this queryTyped is removed

packages/stable/src/__tests__/api/instances.int.spec.ts
```diff
@@ -318,7 +326,50 @@ describe('Instances integration test', () => {
         result_set_1: {},
       },
     });
+    expect(response.items.resuslt_set_1).toHaveLength(1);
```
EliasBjorne:
```suggestion
    expect(response.items.result_set_1).toHaveLength(1);
```

EliasBjorne (commented):
I think we need to update the: 
scripts/extractCodeSnippets.js

As done here: 
https://github.com/cognitedata/cognite-sdk-js/pull/1454/changes#diff-860d9cd7b089cd56ed2b2b3a2e97f40c0d013d2ab9e707d3e38d373256efa519R88

you can alternativly merge in PR: 
https://github.com/cognitedata/cognite-sdk-js/pull/1454

To fix the errors

packages/stable/src/api/instances/instancesApi.ts
```diff
@@ -244,6 +245,41 @@ export class InstancesAPI extends BaseResourceAPI<NodeOrEdge> {
...
+   *      },
+   *    },
+   *  } as const satisfies QueryRequest
+   *  const typedResponse = await client.instances.queryTyped(query);
+   *  // For fully typed results, declare the query variable as:
+   *  // const query = { ... } as const satisfies QueryRequest;
+   *  // then: typedResponse.items.result_set_1[0].properties.mySpace['MyView/v1'].name
+   * ```
+   */
+  public queryTyped = async <
```
silverchen-cognite, line 271: I still don't like having 2 public methods that share the same behavior, this is an anti-pattern and may confuse others. Typing is a compile-time concern and it shouldn't bleed into the runtime code.

silverchen-cognite: Thinking out loud here, have we considered introducing this as breaking change in next major version upgrade?

EliasBjorne: Thats a fair point, one way to avoid having to methods was described here: 
https://github.com/cognitedata/cognite-sdk-js/pull/1444

This was rejected, but we could look at this again and see if this could work, wdyt?

EliasBjorne: Regarding any major update, Im not sure we have any soon incoming, we could sit down and discuss what a major bump could include: 
- DataModel only SDK
- TypedQuery
- Remove some older scervices...

niekbeckers (commented):
@EliasBjorne @silverchen-cognite instead of a `typedQuery`, could we also introduce the new and shiny typed `query` under:

```
client.data_modeling.instances.query()  # <- better typed
```

instead of 

```
client.instances.typedQuery()
```

this is also how the Python SDK namespaces these methods (e.g. `.data_modeling`). 

Pro: easier to migrate people from the untyped to typed SDK. 

Drawback is that we probably would want to start filling in all to the `data_modeling` namespace.

silverchen-cognite:
> @EliasBjorne @silverchen-cognite instead of a `typedQuery`, could we also introduce the new and shiny typed `query` under:
> 
> ```
> client.data_modeling.instances.query()  # <- better typed
> ```
> 
> instead of
> 
> ```
> client.instances.typedQuery()
> ```
> 
> this is also how the Python SDK namespaces these methods (e.g. `.data_modeling`).
> 
> Pro: easier to migrate people from the untyped to typed SDK.
> 
> Drawback is that we probably would want to start filling in all to the `data_modeling` namespace.

IMO, I think this is better than changing the method name. A namespace change can be contained easily on consumer side. And it's confusing to have a new method name just because of a type difference, when the behavior is otherwise the same.

PR #1451, opened 2026-06-23
feat(alpha): add data product versions API support

sgalchynski-cognite:
### Summary
Add Data Product Versions API support to @cognite/sdk-alpha

### What’s included

- New dataProductVersions API in alpha client: create, list, retrieve, update, delete
- Data Product Versions types added in alpha
- Unit tests for all CRUD operations
- Integration tests covering lifecycle flows

### DIFF
--- a/packages/alpha/src/__tests__/api/dataProductVersions.unit.spec.ts
+++ b/packages/alpha/src/__tests__/api/dataProductVersions.unit.spec.ts
@@ -1,0 +1,129 @@
+// Copyright 2026 Cognite AS
+
+import matches from 'lodash/matches';
+import nock from 'nock';
+import { beforeEach, describe, expect, test } from 'vitest';
+import { mockBaseUrl } from '../../../../core/src/__tests__/testUtils';
+import type CogniteClientAlpha from '../../cogniteClient';
+import { setupMockableClient } from '../testUtils';
+
+describe('Data product versions unit test', () => {
+  let client: CogniteClientAlpha;
+
+  const dataProductExternalId = 'dp-1';
+  const versionCreateBody = {
+    version: '1.0.0',
+    views: [{ space: 'dp_1_space', externalId: 'my-view', version: '1' }],
+    status: 'draft' as const,
+    description: 'Test version',
+    terms: { usage: 'Internal use only' },
+  };
+
+  const mockVersion = {
+    ...versionCreateBody,
+    createdTime: Date.now(),
+    lastUpdatedTime: Date.now() + 1000,
+  };
+
+  beforeEach(() => {
+    client = setupMockableClient();
+    nock.cleanAll();
+  });
+
+  test('create', async () => {
+    nock(mockBaseUrl)
+      .post(
+        /\/dataproducts\/dp-1\/versions$/,
+        matches({ items: [versionCreateBody] })
+      )
+      .once()
+      .reply(201, {
+        items: [mockVersion],
+      });
+
+    const items = await client.dataProductVersions.create(dataProductExternalId, [
+      versionCreateBody,
+    ]);
+    expect(items).toHaveLength(1);
+    expect(items[0].version).toEqual('1.0.0');
+  });
+
+  test('list', async () => {
+    nock(mockBaseUrl)
+      .get(/\/dataproducts\/dp-1\/versions\/?$/)
+      .query({ limit: '10', cursor: 'abc' })
+      .once()
+      .reply(200, {
+        items: [mockVersion],
+        nextCursor: 'next',
+      });
+
+    const response = await client.dataProductVersions.list(dataProductExternalId, {
+      limit: 10,
+      cursor: 'abc',
+    });
+    expect(response.items).toHaveLength(1);
+    expect(response.items[0].version).toBe('1.0.0');
+    expect(response.nextCursor).toBe('next');
+  });
+
+  test('retrieve by version', async () => {
+    nock(mockBaseUrl)
+      .get(/\/dataproducts\/dp-1\/versions\/1\.0\.0$/)
+      .once()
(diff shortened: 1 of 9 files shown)


packages/alpha/src/api/dataProducts/types.ts
```diff
@@ -47,3 +47,59 @@ export interface DataProductChange extends DataProductDelete {
...
+}
+
+export interface DataProductVersionTerms {
+  usage?: string;
+  limitations?: string;
+}
+
+export interface DataProductVersion {
+  version: string;
+  views: DataProductViewReference[];
```
bdimitrijoski, line 66: I think this should be the same as ViewRef that we already have in the SDK. Any reason why not reuse that instead?

bdimitrijoski (commented):
Left one minor comment. Code looks good, but I can see that the tests are failing

packages/alpha/src/api/dataProducts/types.ts
sgalchynski-cognite: `ViewReference` has a required `type: 'view'` field. Data Products OpenAPI only has `{ space, externalId, version }` - no type

bdimitrijoski (dismissed):

thorkildcognite:
@sgalchynski-cognite I tried to rerun the build, but tests don't seem to pass. Could you please take a look so it can be fixed before risk review?

bdimitrijoski:
> @sgalchynski-cognite I tried to rerun the build, but tests don't seem to pass. Could you please take a look so it can be fixed before risk review?

I re-run the tests and it is all green now. The part of the pipeline with the integration tests is very flaky.

packages/alpha/src/api/dataProducts/dataProductVersionsApi.ts
```diff
@@ -0,0 +1,129 @@
...
+export class DataProductVersionsAPI extends BaseResourceAPI<DataProductVersion> {
+  /**
+   * @hidden
+   */
+  protected getDateProps() {
+    return this.pickDateProps(['items'], ['createdTime', 'lastUpdatedTime']);
+  }
+
+  private versionsPath(dataProductExternalId: CogniteExternalId) {
+    return this.url(`${encodeURIComponent(dataProductExternalId)}/versions`);
```
silverchen-cognite: nit: while this is functionally correct and secure, but I think we could improve it by introducing a generic helper that accepts `...paths` param, iterates over them, applies `encodeURIComponent` to each segment, and returns them joined with `/`. That would make the code a bit cleaner, avoid repeating `encodeURIComponent` everywhere, and reduce the chance of accidentally forgetting to encode. This is obviously not specific to this PR, so I leave it as nitpick

packages/alpha/src/__tests__/api/dataProductVersions.unit.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      '1.0.0'
+    );
+    expect(response.version).toEqual('1.0.0');
+  });
+
+  test('update', async () => {
+    const updateBody = {
+      version: '1.0.0',
+      update: {
+        description: { set: 'updated description' },
```
silverchen-cognite, line 91: Can we also include test case for `setNull`?

packages/alpha/src/__tests__/api/dataProductVersions.unit.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+      dataProductExternalId,
+      '1.0.0'
+    );
+    expect(response.version).toEqual('1.0.0');
+  });
+
+  test('update', async () => {
+    const updateBody = {
+      version: '1.0.0',
+      update: {
```
silverchen-cognite, line 90: Can we also add test cases for `views`?

packages/alpha/src/api/dataProducts/dataProductVersionsApi.ts
```diff
@@ -0,0 +1,129 @@
...
+
+export class DataProductVersionsAPI extends BaseResourceAPI<DataProductVersion> {
+  /**
+   * @hidden
+   */
+  protected getDateProps() {
+    return this.pickDateProps(['items'], ['createdTime', 'lastUpdatedTime']);
+  }
+
+  private versionsPath(dataProductExternalId: CogniteExternalId) {
```
BugGambit: Private methods should ideally appear below public methods, as developers looking at this file is most likely interested in the public methods and not helper functions.

packages/alpha/src/__tests__/testUtils.ts
```diff
@@ -29,6 +29,84 @@ export function randomInt() {
...
+    }
+
+    await client.spaces.delete([space]).catch(() => {});
+  }
+}
+
+export async function cleanupOrphanedDataProductTestResources(
+  client: CogniteClientAlpha
+) {
+  const spaces = (await client.spaces.list({ limit: 100 })).items
```
BugGambit: Here we only fetch the first 100, and to a client side post filtering. Can't many of the spaces with the prefix `DATA_PRODUCT_TEST_SPACE_PREFIX` never be fetched? Shouldn't we do `autoPaginateToArray`?

packages/alpha/src/api/dataProducts/dataProductVersionsApi.ts
bdimitrijoski: Ok, I can move it

packages/alpha/src/__tests__/testUtils.ts
bdimitrijoski: Added `autoPaginateToArray` here and in `cleanupDataProductSchemaSpaces`

BugGambit (approved):
risk review ok

PR #1452, opened 2026-06-25
fix(alpha): align MeteringDatapoint with API

Aflases:
## Summary
- Rename `MeteringDatapoint.value` to `average` in `types.ts` to match the [Metering API `dataPoint` schema](https://api-docs.cognite.com/20230101-alpha/tag/Metering/operation/fetchConsumptionDataByIds).
- Update `meteringApi.unit.spec.ts` mocks and assertions to use `average`.

This is a type bug fix: the wire response always exposed `average`,`value` was never returned by the API. No changes were needed in `meteringApi.ts` (it only parses `timestamp` to `Date`).


andeplane (approved):
risk review ok

PR #1453, opened 2026-06-25
fix(alpha): fix flaky test in workflows execution

Roshiniswetha:
Fix flaky workflowExecutions.int.spec.ts integration test cleanup by keeping createdExecution set after the cancel test.

After a successful cancel test, createdExecution was set to undefined, so afterAll skipped the cancel step and went straight to deleting the workflow and version. The API rejects workflow deletion while executions are still running, which caused CI failures with:

Some of the specified workflows have running executions

This matches the pattern used in workflowVersions.int.spec.ts and workflows.int.spec.ts, where tracked resources are not cleared before afterAll runs.


andeplane (approved):
🦄

PR #1454, opened 2026-07-13
fix(instances): fix typo in query integration test assertion

EliasBjorne:
## What's wrong

In `instances.int.spec.ts`, the `query` integration test has a typo introduced in this PR:

\`\`\`ts
// BAD - 'resuslt' (letters swapped)
expect(response.items.resuslt_set_1).toHaveLength(1);

// GOOD
expect(response.items.result_set_1).toHaveLength(1);
\`\`\`

\`resuslt_set_1\` is not a key that exists on the response — accessing it returns \`undefined\`, so \`toHaveLength(1)\` would always fail.

## Fix

Single character fix: \`resuslt_set_1\` → \`result_set_1\`.

All type tests in \`query.spec.ts\` were already passing.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/stable/src/__tests__/api/instances.int.spec.ts
+++ b/packages/stable/src/__tests__/api/instances.int.spec.ts
@@ -326,7 +326,7 @@ describe('Instances integration test', () => {
         result_set_1: {},
       },
     });
-    expect(response.items.resuslt_set_1).toHaveLength(1);
+    expect(response.items.result_set_1).toHaveLength(1);
   }, 10_000);
 
   test('queryTyped', async () => {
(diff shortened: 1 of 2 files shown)


packages/stable/src/__tests__/api/files.int.spec.ts
```diff
@@ -134,39 +134,53 @@ describe('Files integration test', () => {
...
+        expect(retrievedFile.metadata).toEqual(testMetadata);
+      });
+    },
+    3 * 60 * 1000
+  );
+
+  test(
+    'download by instance id',
+    async () => {
+      // Syncing from instance to files API is eventually consistant.
```
EliasBjorne:
```suggestion
      // Syncing from instances to files API is eventually consistant.
```

PR #1460, opened 2026-07-28
feat(workflows): add dataDomainExternalId to Workflow and WorkflowUpsert types

Ratna-Phaneendra-Matukumalli:
This PR extends the alpha Workflows API types so clients can associate workflows with a data domain when creating or updating them. The optional dataDomainExternalId field is added to both the Workflow response type and the WorkflowUpsert request type, aligning the SDK with the backend’s support for data domain linkage on workflow resources. The JSDoc example for upsert is updated to show how to pass dataDomainExternalId, and the unit tests are updated


vikasatcognite (approved):

silverchen-cognite (approved):
🦄

PR #1464, opened 2026-08-31
feat(dms): add support for record views

evertoncolling:
This PR adds support for **record views**: `client.views.upsert()` accepts record view definitions, and `list()`/`retrieve()`/`delete()` work with them.

Record views are modeled as a **single flat type** rather than a union, following the same approach as record containers (#1345): `ViewCreateDefinition`/`ViewDefinition` gain an optional `streamId?: string[]`, and `UsedFor` widens to `'node' | 'edge' | 'record' | 'all'`, matching the [UsedFor schema](https://api-docs.cognite.com/20230101/tag/Views/operation/listViews#!in=query&path=usedFor&t=request) the API already uses.

```js
await client.views.upsert([{
  externalId: 'my_record_view',
  space: 'my_space',
  version: '1',
  streamId: ['my_stream'],
  properties: { ... },
}]);
```

## Listing

`GET /models/views` gained a `usedFor` query parameter that defaults to `[node, edge, all]`, so **record views are excluded from `views.list()` unless you opt in:**

```js
const recordViews = await client.views.list({ usedFor: ['record'] });
const allViews = await client.views.list({ usedFor: ['node', 'edge', 'all', 'record'] });
```

## Trade-offs of the flat type

Documented deliberately, mirroring how record containers are typed:

- `streamId` is typed optional but the API requires it when creating a record view (400 at runtime if omitted).
- The full filter DSL type-checks on record views even though the backend accepts a subset (no `not`, `nested`, `overlaps`, `instanceReferences`) results in a 400 at runtime instead of a compile error.
- Widening `UsedFor` means exhaustive `switch` statements over it need a `'record'` arm. Unavoidable under any design, since the API now returns `'record'`.

## Drive-by fix

The integration test cleanup helpers listed views and containers with the default `usedFor`, so record containers and record views were never deleted, and the subsequent space delete failed.

## Test plan

- [x] Unit tests: repeated `usedFor` serialization, scalar params alongside it, omission by default, and pagination carrying both `cursor` and `usedFor`
- [x] Integration tests (run locally against a live project, all green): upsert record view, excluded-by-default, `usedFor` filtering, retrieve, filter round-trip, delete
- [x] `yarn codegen generate-types --package=stable` produces no drift


packages/stable/src/__tests__/api/views.int.spec.ts
```diff
@@ -145,6 +244,153 @@ describe('Views integration test', () => {
     expect(views.items[1].name).toEqual(viewCreationDefinition2.name);
   });
 
+  it('should successfully upsert a record view', async (ctx) => {
+    if (!streamsAvailable) {
+      ctx.skip();
+    }
+
```
polomani: strange to skip tests, shouldn't it be a failure instead?

packages/stable/src/api/views/README.md
```diff
@@ -5,7 +5,30 @@ detecting all shared types across the different data modeling APIs.
...
+the `isRecordViewDefinition` / `isViewDefinition` guards.
+
+The union is discriminated on `usedFor`, which is `'record'` for record views
+and `'node' | 'edge' | 'all'` for regular ones.
+
+Note that `client.views.list()` excludes record views unless `usedFor` includes
+`'record'`, matching the API default of `[node, edge, all]`.
+
+Record views inside data models are not typed yet: `DataModelProperty` and
+`DataModelCreateProperty` are still view-only in the public contract.
```
polomani:
```suggestion
```

strange to mention like this.

packages/stable/src/api/views/types.ts
```diff
@@ -0,0 +1,54 @@
...
+  SpaceQueryParameter,
+  ViewCreateDefinition,
+  ViewDefinition,
+} from './types.gen';
+
+/**
+ * A view returned by the views endpoints: either a regular view backed by
+ * nodes/edges, or a record view backed by Streams & Records storage.
+ */
+export type ViewOrRecordViewDefinition = ViewDefinition | RecordViewDefinition;
```
polomani: to be honest I am not sure if it's worth it, only one property difference. maybe it's a better devX if we add one optional streamId property and be done with it?

polomani: ah, yes, filters also. but this might break people who use current types, so optional property & with same filters is much less disrupting

packages/stable/src/__tests__/api/views.int.spec.ts
evertoncolling: Agreed and addressed!

packages/stable/src/api/views/README.md
evertoncolling: Removed. The README now carries only a short note that the record-view fields are hand-added ahead of the OpenAPI snapshot (so a regeneration would erase them).

packages/stable/src/api/views/types.ts
evertoncolling: Went with your suggestion, reworking the PR as a single flat type: optional streamId on ViewCreateDefinition/ViewDefinition, UsedFor widened to include 'record', and the same FilterDefinition (no separate record filter type). 

Matches how record containers shipped in #1345. Nothing breaks for existing consumers. The trade-off is that streamId being required for record views and the reduced filter subset are now enforced at runtime by the API rather than at compile time (documented in the PR description).

packages/stable/src/api/views/README.md
```diff
@@ -9,3 +9,7 @@ Until this is resolved, follow the following process to generate new types:
 4. Resolve any new duplicate types by deleting them from the newly added export lines.
 5. Delete the newly generated exports in `types.gen.ts`
 6. Rename `codegen.json` to `codegen.skip.json`
```
polomani: let's just say: we the codegen is disabled here and it is not worth enabling it since the output is sub-optimal and instead of spending time fixing we can rely on llms to write more ts native / dev friedly code :)

evertoncolling: Updated the README making it clear that it is intentional to keep views outside of codegen :)

evertoncolling:
Created https://github.com/cognitedata/cognite-sdk-js/pull/1466 to address the overly frequent flaky tests on each CI run

packages/stable/src/__tests__/api/views.int.spec.ts
```diff
@@ -70,12 +115,53 @@ describe('Views integration test', () => {
...
+        description: 'Record container used for views integration tests.',
+        usedFor: 'record',
+        properties: {
+          temperature: {
+            type: { type: 'float64' },
+          },
+        },
+      },
+    ]);
+    // Raised from 25s: this hook also provisions a stream and a record container.
```
polomani:
```suggestion
```
self explanatory

packages/stable/src/__tests__/api/views.int.spec.ts
```diff
@@ -70,12 +115,53 @@ describe('Views integration test', () => {
...
+  }, 60_000);
   afterAll(async () => {
+    // The record view must go before the container it maps.
```
polomani:
```suggestion
```

packages/stable/src/api/models/types.gen.ts
```diff
@@ -1345,9 +1345,11 @@ export interface UpsertConflict {
   };
 }
```
polomani, line 1345: can we add en exception that this is not generated somehow? `file/path/types.gen.ts linguist-generated=false` perhaps?

evertoncolling: Yes, made it clear which files are not using codegen in https://github.com/cognitedata/cognite-sdk-js/pull/1464/commits/fa7e836ed8f8f01bb6b56da6d502afd9c1be1128

polomani (approved):
🦄

PR #1467, opened 2026-09-01
test: fix flaky integration tests

evertoncolling:
## Why

CI has been failing intermittently on master and on unrelated PRs (e.g. #1464) for weeks. Two consecutive runs of the same commit fail on *different* integration tests:

- `dataProductVersions.int.spec.ts` (alpha): `Hook timed out in 10000ms`
- `files.int.spec.ts` (stable): 408s / `File instance ids not found` on the *by instance id* tests

## What

**Timeouts (root cause of the alpha hook failure)**
- Apply the intended vitest timeouts (25s test / 30s hook) to the **alpha, beta and template** packages — the follow-up explicitly deferred in #1465. The data product suites' `beforeAll` (fixture setup + orphan cleanup) routinely exceeds vitest's default 10s hook timeout.

**Retry helper**
- `runTestWithRetryWhenFailing` now uses a capped backoff with a sleep budget (default 20s) that actually fits inside the 25s test timeout. Previously uncapped exponential backoff blew past the timeout after ~5 retries, and its `vi.setConfig(3 min)` call was a documented no-op.

**Missing `await`s — which exposed two real test bugs**
- `files.int.spec.ts` *download by instance id* and four `sequences.int.spec.ts` tests called the retry wrapper (or `expect().rejects`) without `await`, so they never retried or asserted anything and always passed. Awaiting them revealed:
  - the sequences `beforeAll` **reassigned** `sequenceToCreate` instead of mutating it, so the created sequence never carried the `assetId` the three asset-filter tests depend on — they could never pass;
  - the search test's wildcard query (`n*m* des*tion`) is never matched by the search endpoint (verified against the API: plain words match within seconds, wildcards never do).
  Both fixed; the asset-filter/search tests get a 3 minute budget since those indexes can lag.

**Cross-run fixture race (likely source of the files failures)**
- The shared `test_data_space` fixture space (files + datapoints suites) is now excluded from `deleteOldSpaces`, which the views/containers/spaces/datamodels suites run in parallel workers and in the node 20/22 matrix — the same race #1466 fixed for data products.
- The data product versions `list` test now pages through all results, like #1466 did for the products list test.

## Verification

Ran locally against the integration test project:
- `dataProductVersions.int.spec.ts`: 5/5 passed (suite takes ~12s — would have flaked on the old 10s hook timeout)
- `files.int.spec.ts`: 17/17 non-skipped passed
- `sequences.int.spec.ts`: 14/14 passed, twice
- `yarn lint` clean

### DIFF
--- a/packages/stable/src/__tests__/api/sequences.int.spec.ts
+++ b/packages/stable/src/__tests__/api/sequences.int.spec.ts
@@ -15,12 +15,15 @@ import {
   setupLoggedInClient,
 } from '../testUtils';
 
+const SLOW_INDEX_TEST_TIMEOUT_MS = 3 * 60 * 1000;
+const RETRY_SLEEP_BUDGET_MS = SLOW_INDEX_TEST_TIMEOUT_MS - 30 * 1000;
+
 describe('Sequences integration test', () => {
   let client: CogniteClient;
   let sequences: Sequence[];
   const testValues = [1, 1.5, 'two'];
   const testExternalId = `sequence${randomInt()}`;
-  let sequenceToCreate: ExternalSequence = {
+  const sequenceToCreate: ExternalSequence = {
     name: 'sequence1',
     description: 'description',
     columns: [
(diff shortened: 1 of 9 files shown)


packages/stable/src/__tests__/api/sequences.int.spec.ts
```diff
@@ -159,16 +176,22 @@ describe('Sequences integration test', () => {
...
-      });
-      expect(result.length).toBeGreaterThan(0);
-    });
-  });
+  test(
+    'search',
+    async () => {
+      await runTestWithRetryWhenFailing(async () => {
+        // The search endpoint matches whole words: wildcard patterns like
+        // 'des*tion' never match anything.
```
polomani, line 184:
```suggestion
```
strange.. this worked before

polomani (approved):
🦄

PR #1468, opened 2026-09-01
chore(deps): upgrade lerna to v10

polomani:
lerna v9 (even though it's only a dev dependency) pulls a bunch of old package versions with CVEs
we hopefully resolve a lot of them by just bumping lerna instead.

looks like dependabot/renovate gave up on this repo


BugGambit (approved):
risk review ok

PR #1469, opened 2026-09-01
feat(records): support record views as sources in the records API

evertoncolling:
This PR extends the Streams & Records API (`client.records`) to support **record views** everywhere a container reference was accepted, matching the ILA API spec. This is the client-side follow-up to #1464, which added record views to the DMS views API.

### Type changes (`packages/stable/src/api/records/types.ts`)

- New `ViewReference` and `SourceReference = ViewReference | ContainerReference` types.
- Widened the three container-only sites to the union:
  - `RecordData.source` (ingest/upsert)
  - `SourceSelector.source` (filter/sync)
  - the `hasData` filter leaf

Property path references (`FilterProperty`, `AggregateProperty`, `TargetUnitProperty`) were already structurally compatible with the view form, `['space', 'viewExternalId/version', 'property']`, so those got documentation updates only.

### Documentation

- View property-path form for filters, aggregates, and targetUnits.
- `'*'` wildcard in `SourceSelector.properties`.
- Response `properties`/`typing` maps keyed by `viewExternalId/version` when selecting through views.
- Sort supports container property references only.
- Aggregate requests can reference at most one property source when a view is involved.
- Container and view sources cannot be mixed in a single request (API-enforced; found via integration testing — not documented in the spec).
- `filter`/`ingest` doc snippets now show view sources.

### Tests

- New end-to-end integration test: ingest through a view (view property names), filter with a view source selector + view property path + `includeTyping`, and aggregate on a view property path. Uses a persistent record view over the existing persistent record container. Behavioral semantics (embedded view filters, unit conversion, aggregate math) are left to the API's own integration tests.
- New nock-based `records.unit.spec.ts` covering view-source request serialization and deserialization of view-keyed responses through the date transform.

### Breaking changes

None, widening the source types is source-compatible, and the new type names are shadowed at the package boundary by the structurally identical DMS instances re-exports (same precedent as the existing records `ContainerReference`).

## Test plan

- `yarn build`, `yarn lint`, `yarn test-snippets` all pass; `yarn codegen generate-types --package=stable` produces no drift.
- All records unit + integration tests pass locally against CDF (16 tests), plus the views suites that share the stream fixture.

### DIFF
--- a/packages/stable/src/__tests__/api/records.int.spec.ts
+++ b/packages/stable/src/__tests__/api/records.int.spec.ts
@@ -1,6 +1,7 @@
 // Copyright 2025 Cognite AS
 
 import { afterAll, beforeAll, describe, expect, test, vi } from 'vitest';
+import type { ViewCreateDefinition } from '../../api/views/types.gen';
 import type CogniteClient from '../../cogniteClient';
 import type { RecordDelete, SyncRecordItem } from '../../types';
 import {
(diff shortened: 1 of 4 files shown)


andersfylling:
:unicorn:

andersfylling (approved):

PR #1485, opened 2026-09-15
fix(codegen): dedupe types that match @cognite/sdk-core exports

vikasatcognite:
## Summary

- Add an AST transformer that structurally compares generated type declarations against `@cognite/sdk-core` and removes duplicates, replacing them with imports from `@cognite/sdk-core`.
- Dynamically loads core type definitions from `packages/core/src/types.ts` (no hardcoded exclusion list), so types like `CogniteExternalId`, `CogniteInternalId`, `InstanceExternalId`, and `InstanceSpace` are deduped when shapes match; types with different shapes (e.g. `CogniteInstanceId`) are kept.
- Unblocks the duplication feedback on extractors API PR #1480 — stable regen is intentionally **not** included here.

## Test plan

- [x] `yarn lint` in `packages/codegen`
- [x] `yarn test src/__tests__/codegen.uint.spec.ts src/__tests__/sdk_core_declarations.uint.spec.ts`
- [ ] CI codegen unit tests pass
- [ ] After merge: run `yarn codegen generate-types --package=stable` and commit stable regen separately (or as part of extractors stack)


packages/codegen/src/__tests__/codegen.uint.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('code generation', () => {
     });
 
     basicTypesServiceBGenFile = (
-      await fs.readFile(`${testFolder}/testdata/8-serviceB-types.gen.ts`)
+      await fs.readFile(`${testFolder}/testdata/8-serviceB-types.gen.snapshot`)
```
andersfylling: why?

packages/stable/src/api/documents/types.gen.ts
```diff
@@ -2,26 +2,19 @@
...
+import type {
+  CogniteExternalId,
+  CogniteInternalId,
+  CursorAndAsyncIterator,
+  InstanceExternalId,
+} from '@cognite/sdk-core';
 /**
  * The ID of an [instance in Cognite Data Models](https://docs.cognite.com/cdf/dm/dm_concepts/dm_spaces_instances#instance).
  */
 export interface CogniteInstanceId {
```
andersfylling, line 14: this is also in the core package and should be imported

andersfylling:
I think this is the right direction, but note that some of the modules now no longer holds a type declaration. That means any dependents will break, so we need to have some backwards compatibility support.

packages/codegen/src/__tests__/codegen.uint.spec.ts
polomani: +1, also wondering

vikasatcognite: .snapshot rename was just to show that the file is test-only. earlier it was showing a TS error but https://github.com/cognitedata/cognite-sdk-js/pull/1485/changes#diff-d80e996688ff15c23bdfa873b228390682ed869fc74d79faf14cb7b984e9a126R3 alone fixes it so the rename has no contribution here. reverted it

Image

packages/stable/src/api/documents/types.gen.ts
vikasatcognite: Although same name, the printed shape is different. In sdk-core it is `{ externalId: InstanceExternalId; space: InstanceSpace }`, in documents it is `{ externalId: InstanceExternalId; space: Space }`. InstanceSpace and Space are both string aliases though, but dedup compares the printed member types literally and this interface is not removed.

andersfylling:
:unicorn:

andersfylling (dismissed):

packages/codegen/src/ast_transformer/sdk_core_types.ts
```diff
@@ -0,0 +1,123 @@
...
+  importDeclaration: ts.ImportDeclaration
+): Set<string> => {
+  const imported = new Set<string>();
+  const namedBindings = importDeclaration.importClause?.namedBindings;
+  if (namedBindings == null || !ts.isNamedImports(namedBindings)) {
+    return imported;
+  }
+
+  for (const element of namedBindings.elements) {
+    imported.add(element.name.escapedText as string);
```
EliasBjorne: do we need this cast?

packages/codegen/src/ast_transformer/sdk_core_types.ts
```diff
@@ -0,0 +1,123 @@
...
+    ts.factory.createStringLiteral(SDK_CORE_MODULE)
+  );
+};
+
+const collectReferencedTypeNames = (
+  node: ts.Node,
+  referenced: Set<string>
+): void => {
+  if (ts.isTypeReferenceNode(node) && ts.isIdentifier(node.typeName)) {
+    referenced.add(node.typeName.escapedText as string);
```
EliasBjorne: same here, can we avoid casting

EliasBjorne (dismissed):
LGTM, there is a lot of casting to string, but not blocking

vikasatcognite:
> LGTM, there is a lot of casting to string, but not blocking

my bad. removed all typecasts

EliasBjorne (approved):
lgtm

andersfylling:
:unicorn:

andersfylling (approved):

andersfylling:
:unicorn:

andersfylling (approved):

